PR #15, opened 2020-09-30
fix for small displays that  cannot import settings due to scroll and…

osodebailar:
… display too large texts problem like Jelley

fix the problem with following changes:

- add marquee element to alert dialog header . marquee scrolling will be activated automatic if text is too large
- only show import filename, not the path. To show the filepath is not necessary for a normal user
- add nested scroll behaviour for import summary.
- minor padding and marging changes

due too future gui enhancements nested scroll with coordinator layout is used insted of linear layout with normal scroll element

### DIFF
--- a/app/src/main/res/layout/dialog_alert_import_summary.xml
+++ b/app/src/main/res/layout/dialog_alert_import_summary.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="wrap_content">
 
--- a/app/src/main/res/layout/dialog_alert_import_summary.xml
+++ b/app/src/main/res/layout/dialog_alert_import_summary.xml
@@ -7,9 +7,7 @@
     <androidx.core.widget.NestedScrollView
         android:id="@+id/scroll_content_frame"
         android:layout_width="match_parent"
-        android:layout_height="210dp"
-        app:layout_behavior="@string/appbar_scrolling_view_behavior">
-d
+        android:layout_height="210dp">
         <LinearLayout
             android:layout_width="match_parent"
             android:layout_height="match_parent"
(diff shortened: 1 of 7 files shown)


dlvoy:
Hi Andreas, have some issues with details button inside layout:

Screenshot_1601481326

Screenshot_1601481919

It should be at bottom of table, while it strangely hovers over it.
On higher screens `150dp` is way to small,  we should use screen space estate to show all details without requiring scroll, if possible. 
That table is meant for user security and IMHO - if only possible - everything should be visible at first glance (users are lazy and do not scroll, especialy in such non-obvious places).

At https://github.com/dlvoy/AndroidAPS/commit/d763614ae64006eaea6cc58b604d43f4194a6ad1 I've tried to achieve that by having main xml file with small screen fix, but as an alternative `layout-sw360dp-land/dialog_alert_import_summary.xml` with old version that use full height and do not scroll.

Marquee is funny trick but do its job - reminding ol good HTML 2-ish era :D

osodebailar:
Ok , I will check it and test in Addition with Button on bottom - in my Test the Button was not shown , so i did not see the Problem.

osodebailar:
@dlvoy Dominik i fixed the button and change message textsize , so i can set the height to 210dp as in your solution.

app/src/main/res/layout/dialog_alert_import_summary.xml
```diff
@@ -1,30 +1,46 @@
...
-        android:layout_height="wrap_content"
-        android:layout_marginLeft="8dp"
-        android:layout_marginRight="8dp"
-        android:stretchColumns="2" />
+    <androidx.core.widget.NestedScrollView
+        android:id="@+id/scroll_content_frame"
+        android:layout_width="match_parent"
+        android:layout_height="210dp"
+        app:layout_behavior="@string/appbar_scrolling_view_behavior">
+d
```
dlvoy, line 12: stray character - please remove

app/src/main/res/layout/dialog_alert_import_summary.xml
```diff
@@ -1,30 +1,46 @@
...
 
-</LinearLayout>
\ No newline at end of file
+            <Button
+                android:id="@+id/summary_details_btn"
+                style="?android:attr/buttonStyle"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_gravity="bottom|center_horizontal"
+                android:gravity="bottom"
```
dlvoy, line 33: `android:gravity="bottom"` causing v-align of button issue (as seen below), removing that property will fix it

dlvoy (changes requested):

app/src/main/res/layout/dialog_alert_import_summary.xml
osodebailar: fixed the issues .

dlvoy:
Reviewed & retested - looks good!

PR #34, opened 2020-10-11
[WIP] : Autotune v3

Philoul:
3rd version of Autotune Plugin (Pugin NOT VALIDATED, this work is shared only for tests and help to finish it)
Main modifications compared to https://github.com/MilosKozak/AndroidAPS/pull/2706 :
- Use average of last 4 hours of "PumpProfile" basal rate for IOB calculation (like it's done in Oref0 Autotune)
- Add Neutral TBR (TBR with active profile basal rate) when no TBR is running in Tempbasal table used for IOB calculation (to be able to use "CurrentBasal" basal rate as reference instead of Profile Basal Rate)
- Add an Automation Action for Autotune (to be able to run automatically Autotune). Do not use with profile activation enabled in Autotune settings...

Todo and known bug:
=> Results are not identical with oref0-autotune, iob calculated by oref0-autotuneprep are not consistent with iob calculated by AAPS (I need help for this point).
=> Create or Update an "Autotune" local profile when we press "Activate Profile" button (without this Local Profile, we can't make a profile switch with duration or percentage from a Tuned profile activated directly from plugin).
=> Build datasets to be able to do real "validation" of results in different situations (I started to add some tests)
=> I list some "Todo" in the beginning of AutotunePlugin.java file (some of them are required before a first released version in AAPS, other could be done in future)

=> Limitation with this version are the same than for oref0-autotune (only one single IC and ISF value). May be future updates after validation of a first version of autotune for full circadien profiles...
Sorry for hundreds of commits

Feedback would be appreciated...


linus2100:
Hello @Philoul , 
I'm not sure whether this is the right place or dev state for me to comment. So please just ignore if not.
I tested V2 successfully with minor glitches, but then returned to an automatic nightly docker run of Autotune (AT).
Two days ago I started with V3 collecting data, which was necessary for me with V2, where I couldn't run AT on data beyond the AAPS-AT-V2 installation date. But now there is no success with AT and a one day horizon of data. The AT-log begins with: 

00:01:34.542 [RxCachedThreadScheduler-81] E/CORE: [TreatmentService.createOrUpdate():467]: Unhandled exception
java.sql.SQLException: Unable to run insert stmt on object Treatment{date= 1604720556000, date= 11/7/20 04:42, isValid= true, isSMB= true, _id= 5fa617b1c7e7f652d68ab121, pumpId= 1604720556000, insulin= 1.3, carbs= 0.0, mealBolus= false, source= 2}: INSERT INTO `Treatments` (`_id` ,`boluscalc` ,`carbs` ,`date` ,`dia` ,`insulin` ,`insulinInterfaceID` ,`isSMB` ,`isValid` ,`mealBolus` ,`pumpId` ,`source` ) VALUES (?,?,?,?,?,?,?,?,?,?,?,?)

Are there known incompatibilities or settings (i.e. language locale, free peak oref, ...) for me to consider before running?

Philoul:
Hi @linus2100 ,
First of all, this plugin is not validated today (if you increase number of days of calculation, results are not convergent), so I share it only for tests (I need some help from more skilled developers to correct this). You shouldn't use it for your loop...
To analyse your problem I need autotune log file (for autotune plugin, there are specific zip files that containts all information of each autotune run done in AAPS, these files are in the same folder than AAPS log files) and a screenshot of Autotune plugin after calculation could help too (I add an error message in the interface when I can't get data) (you can send me AT log in gitter with PM if you want)
I got this kind of error a few month ago (see https://github.com/MilosKozak/AndroidAPS/issues/2708) but it was not directly in Autotune code but in a standard query done in another plugin of AAPS to get data from NS. To verify if it's the same problem, you can open History Browser in AAPS and select a day before your AAPS-AT-V2 installation date. if in History browser you don't have carbs data (orange curves missing), it's the same issue...
=> In my case it was a local issue on my loop phone (I could get data on my dev phone and test autotune plugin on it at the same time...), on my loop phone I just waited. (maybe a complete reinstallation of AAPS could have fixed the problem, but I do not recommend it ...) .

Today, there are no known incompatibilities (but very few persons test it too so I can't be 100% sure...).

linus2100:
Hi @Philoul ,
Thanks a lot for your quick reply and estimation. I won't compromise my loop. 
Coming from openAPS 2 years ago, I'm still a big fan of AT and would like to see it work well with AAPS. Unfortunately, checking the history browser is not an option for me. I did a databases reset with AAPS-AT-V3 install (05.11.), as this helped me with V2 once. I'll pm you on Gitter with additional data, as proposed.

Philoul:
I'll have to update all this work with a new database, so major changes required before a new "mergeable" proposal.
I propose to close it.

PR #41, opened 2020-10-17
Level informations added in Action>Careportal

Philoul:
Here an update of Careportal  info in Action tab to add level (see https://github.com/MilosKozak/AndroidAPS/issues/2744)
=> Reservoir level and Pump Battery level added
=> I added Sensor battery level too, but unfortunatly there is a bug in xDrip (it sends Phone battery level instead of sensor battery level, see https://github.com/NightscoutFoundation/xDrip/issues/1469 and https://github.com/NightscoutFoundation/xDrip/pull/1471), so I made a patch to disable this info today in current PR.
=> 2 thresholds are added too for Sensor Battery Level status

See result in screenshot below (on the right with patch to disable Sensor Battery Level). I hope xDrip bug will be fixed soon...
Careportal_Info

I finaly replaced LinearLayout by a TableLayout in careportal_stats_fragment, and I removed "Level" label for NSClient versions of AAPS.

Note: Eversens plugin has also Sensor Battery Level information, can someone send me a log with this informations (key "batteryLevel"), I don't know if it's an integer like Xdrip or if it's a string...)


rICTx-T1D (commented):
Is it possible to get the transmitter battery level (or age) from patched dexcom app?
Can you make an indicator too for sensor age?
(I like your idea to add new functions to careportal.)

Philoul:
I didn't find this information in Dexcom plugin (only in xDrip plugin and eversens plugin has sensor bat level but if available it's already prepared)
=> unfortunatly there is a bug in sensor bat level sent by xdrip (I proposed a PR in xdrip project to fix it https://github.com/NightscoutFoundation/xDrip/pull/1471)
=> I need a log sample of eversens broadcast to be sure I can convert to %

swissalpine:
Thank you for your work on many fields! It's very appreciated!
But: Please, think about small Displays like Atom. We don't need words like "age", keep it short, probably text only:
Sensor    3j0h | 70%
Or a possibility to choose if the extended informations are shown.
Thanks.

core/src/main/java/info/nightscout/androidaps/interfaces/BgSourceInterface.java
```diff
@@ -9,4 +9,6 @@
     boolean advancedFilteringSupported();
 
     void handleNewData(Intent intent);
+
+    int getSensorBatteryLevel();
```
MilosKozak, line 13: I'd prefer default function here instead of override everywhere and return -1

Philoul: That was my first idea but abstract method cannot have body...
I have to add a dedicated module like PumpDescription ?

Philoul:
@swissalpine Thank's for your feedback.
I shortened as I could labels and layout for low res phone (my dev phone has 960×480px) but I understand that it's not enough...
Can you attach a screenshot in full res of your phone to help me see how I can improve layout for "very low res devices" ?
I'm not fan of another additional setting...
@MilosKozak do you think it's better for all config to have a very short display as proposed by swissalpine or to add a new setting to select "short text" for very low res?

swissalpine:
Screenshot_20201019-124920

Philoul:
@swissalpine I finally added a new setting for very low res (Preferences/overview/status lights) (default false)

See results below in my high res phone. Can you check if it's ok on your Atom please ? (spacing is reduced in lower res)
CarePortalWithSetting

MilosKozak:
Adding a new setting for displaying every single element is a way to hell. It will lead to xDrip syndrome: so much setting, nobody understand what is it for.
I'd do one of following:
1) use only short or long
2) autodetect
3) integrate for example with general short labels menu setting

Philoul:
Do you have an autodetect elsewhere in AAPS?
My idea is to just select mode according to screen width (i.e. <480px), but I don't know if it works well...

dlvoy:
@MilosKozak maybe we can introduce new theme - as we have currently (Orginal / Buttons always  at bottom / Big screen) and add new: **Small screen**
And Atom users can select it to adjust layouts in many parts of the app

MilosKozak:
> @MilosKozak maybe we can introduce new theme - as we have currently (Orginal / Buttons always at bottom / Big screen) and add new: **Small screen**
> And Atom users can select it to adjust layouts in many parts of the app

good idea

swissalpine:
@Philoul 
Here are the screenshots.
The dash | was only thought/needed if the information is in one table cell. This is not necessary in two table cells, as there is a gap between them.
The information "age" etc. is unnecessary in any layout, as the units explain what is meant.
Screenshot_20201020-091103
Screenshot_20201020-091226

@MilosKozak @dlvoy 
I would also like to have a separate theme for small displays. Also, I have already reworked the "original" (and it also works for large displays) - and could make it available ...
Screenshot_20200901-120039

Philoul:
With my latest commits, I replaced dedicated setting by a new "Low Res Skin", and I shrinked a little bit more Low Res version of Careportal informations
=> For Low Res Skin, I only made a copy of classic skin for initialization...
(It's difficult for me to work on it because my dev phone (Hola) has a too high resolution...)

Philoul:
@swissalpine Low Res Skin is not the purpose of this PR (I just initialize it for setting labels of Careportal layout with additionnal informations)
Can you share your layout informations to update this skin please (I need heights of primary and secondary graphs if you modified them for Atom compared to classic skin).
If you made other changes (status lights layout, avg Delta...), I can try to include some of them only if it's not too complicated (I prefer later improvement of this new skin done in dedicated PR by specialists...)

swissalpine:
@Philoul Thanks.
Unfortunately I don't know what and where I have changed everything in the last two years ... I had never planned to publish it, because that was just my preference. But everything is available in my repo:
https://github.com/swissalpine/AndroidAPS/tree/dev-mod5
The adjustments in overview_statuslights_layout, overview_buttons_layout.xml, overview_info_layout.xml and overview_loop_pumpstatus_layout.xml can probably be applied.
I'm just unsure because some of the changes are also based on adjustments in the OverviewFragment.kt (e.g. the deleted 45 minutes delta). But it could work ...???
I have also adjusted the Combo tab.
And: Be generous with my code. I'm not a professional and I taught myself some programming to adapt my diabetes stuff.

Philoul:
@swissalpine Thank's, I will compare to see differences.
I'm not a professional too, I'm just a passionate trying to do my best to help...

Philoul:
@swissalpine It's same setting than skinclassic concerning heights of graphs, and I saw you have disabled landscape mode in overview.
There are lots of little modifications (strings replacement, hard coded "strings" modifications, "wrap_content" replaced by fix width in some layouts...) and I think some of these modifications that work on your Atom could have side effects on other low res devices...
=> It could be a good starting point but it's necessary to reworks lots of things (add additional "veryshort strings", add some specific layout, link these strings/layout with LowResSkin) and I think it's better to work on all these modifications in a dedicated PR or Branch 
(I tried to test this proposal on Atom with emulator, I think it's ok but text size is smaller than in your screenshots, even if I select Largest font size in settings...)

app/src/main/java/info/nightscout/androidaps/plugins/general/actions/ActionsFragment.kt
```diff
@@ -233,15 +235,44 @@ class ActionsFragment : DaggerFragment() {
...
-
+        val activeBgSource = activePlugin.activeBgSource
         actions_historybrowser.visibility = (profile != null).toVisibility()
         actions_fill?.visibility = (pump.pumpDescription.isRefillingCapable && pump.isInitialized && !pump.isSuspended).toVisibility()
         actions_pumpbatterychange?.visibility = pump.pumpDescription.isBatteryReplaceable.toVisibility()
         actions_temptarget?.visibility = (profile != null && config.APS).toVisibility()
         actions_tddstats?.visibility = pump.pumpDescription.supportsTDDs.toVisibility()
 
-        statusLightHandler.updateStatusLights(careportal_canulaage, careportal_insulinage, null, careportal_sensorage, careportal_pbage, null)
+        val shortlabel = skinProvider.activeSkin().description == R.string.lowres_description
```
MilosKozak, line 245: this is dangerous and hard to maintain

Philoul: @MilosKozak LowRes Layout added for Actions plugin and a new function in Skininterface (like overview)
=> is it better now ?

See results with emulator: LowRes enabled according to orientation and Width value (compared to Constants.SMALL_WIDTH)
AtomEmulator

PR #64, opened 2020-11-10
Patient age string Rename

Philoul:
With the addition of pregnancy for the safety limits, I think the category should be renamed as not directly related to age.
I suggest "Patient type"


rICTx-T1D (approved):
I think it usefully to rename this property.  

Maybe its better to create a new @string, because translations in other languages will not be changed. When you plan this, refactoring the variable too. But this way is very time consuming,  so I understand the smaller solution.

Issue #67, opened 2020-11-12
UI Feedback


jamespaulley:
### Overview Screen
I thought it might be nice to have some more useful buttons on the overview screen:

- Temp target buttons:
  - I find setting a temp target is quite fiddly and hard. There are too many long-presses, clicks and small UI elements involved. Definitely hard for people with shaky hands, etc. It might be nice to have large buttons for these functions on the overview screen, similar to the grey buttons already present.
  - User can choose which targets appear as buttons.
  - The buttons either immediately apply the temp target, or prompt for a confirmation if you think that's safer.
  - Maybe there could be a text area that immediately updates with the temp target that is going to be applied, once the connection to the pump has been established.
- Profile switch buttons:
  - Would be cool if there were -/+ buttons that jump in 10% (or user-defined) intervals. You can click the buttons multiple times (for example to set a profile of 130%), then after a short delay (2 seconds?) the profile switch is applied.
  - There's also a cancel button to set the profile back to 100%.
  - There could be a text area that immediately updates with the profile percentage that will be applied, once the connection to the pump has been established.
  - If you think there are safety concerns with this behaviour, there could at least just be a button that opens the profile switch interface?
- Loop button:
  - Would be cool if we could have the option of the Loop status button (that allows disconnection of pump, etc) to be a normal button, i.e not long-press

### General UI feedback
- Long press is hard for people with shaky hands, especially on small buttons. Can we have the option to have all button functions without long press?
- There are too many very small UI elements. These are hard to click.
- Drop down menu in Temp target interface (next to "reason") is too small and fiddly and annoying. Can we get large buttons for each temp target instead of a drop-down list?
- Percentage -/+ buttons on Profile Switch interface currently go up/down by 1%. Can we have them jump in intervals of 10%, or at least an option to have that.

![image](https://user-images.githubusercontent.com/26227058/98962057-253d4e80-24fe-11eb-9475-55a9bd9ab8e2.png)

mistermintsgh:
I like these!

Can I also add to this list... I **HATE** the 3 dot menu in the top right hand corner. I have no mobility issues or anything like that described above, but I cannot tell you how often I go to press those 3 dots and end up somehow pulling down the notification/quick settings menu of my phone.

[...]

PR #76, opened 2020-11-22
UI UX improvement proposals

Philoul:
UI-UX improvement (see #67)
This PR is only to be able to test proposals, please give feed back to help selecting which proposal is good or not.
I tried to make dedicated and independent commits for each proposal, that way it will be quite easy to revert each one. This is a first PR, it will be completed to includes other proposals...
- https://github.com/nightscout/AndroidAPS/commit/efa6991792204bb5bed0e9ef3fdd8b2378df2fe4 : Increase the height of 3 dots menu in Compact tab
- https://github.com/nightscout/AndroidAPS/commit/1ea3e5ec530276415fbc466c2dbac475eb18751d : replace sp by dp (only for buttons, info and status light layouts of overview) to keep layout design whatever text size selection in Android settings
- https://github.com/nightscout/AndroidAPS/pull/76/commits/d710f4e1b1b9171bc72ff091a1f82d2a518a7224 : remove label on bottom's buttons (+ https://github.com/nightscout/AndroidAPS/pull/76/commits/d1414fc84d697569ed273891148dcf4a3a13a642 + https://github.com/nightscout/AndroidAPS/pull/76/commits/4b0bab2005de22a792cf072e4261706bf5be215c + https://github.com/nightscout/AndroidAPS/pull/76/commits/fda3b5912e437f75ffeb8876da3fcb57dd058982)
- https://github.com/nightscout/AndroidAPS/pull/76/commits/e2321d7b0f899748114cf2e54363611ad30a7e32 : Show context menus with single click (light modification so it works with long press too) and increase a litlle bit height of Profile and Temp Target buttons

Concerning the "merge" of Carbs and Insulin buttons in one Treatment button, I'm afraid it's too complicated for most users so I let it "as is" (too many fields and buttons) and without labels, it's easy to add one or 2 additional buttons...
I have finished to work on this PR now (I'm waiting for comments to have feedback to update it if necessary...)


Philoul:
for commit https://github.com/nightscout/AndroidAPS/pull/76/commits/d710f4e1b1b9171bc72ff091a1f82d2a518a7224 see screenshots with or without quickwizard. I reduced size of quickwizard icon to not increase to much vertical heigth of buttons.
If not enough place availlable for all buttons, side effect is icon is not centered in other buttons (I don't know how to center them)
Note: you can see in this screnshot increase of heigth of 3 dots menu...)
Remove labels

MilosKozak:
i'd probably agree with moving timeago under arrow too

Philoul:
See below the impact of BG Value size on an Atom emulator and on my loop phone (Samsung A40). Current size is 60dp and if I reduce to 50dp the is a little more space for graphs (screenshot just to see the impact, no commit done)
BgSizeFrom60to50

MilosKozak:
i'd not make number smaller but maybe chage margin to get the same effect

Philoul:
See results (Atom emulator) with the 2 additional buttons added in preferences (6 buttons is probably too many for a small phone but it's just to show the result)...
Screenshot_Atom_Emulator

@MilosKozak I tried to change margin, but with constraint layout, negative margin doesn't work (that's why I suggested the reduction of bg value text size). In Layout the top of 5 icons (IOB, COB...) are constraint to bottom of BG Value. The other possibility I imagine is to constraint everything to parent and use positive margin for these icons to adjust position....

Philoul:
Not discussed in #67 but see in screenshot a proposal for ProfileSwitch Dialog (I think it's better to have first Profile selection (and reuse button), and then Duration / Percentage and finaly Time offset settings)

PF_layout_Modif

Philoul:
Two latest commits were to fix buttons layout  with no labels (vertically centered)
=> I created a new SingleClickButtonImage button type in ui, but I had to rename buttons in no_label layout to avoid conflicts...
=> It works but I don't know if it's "clean"
I think removing labels allow adding one or 2 buttons in overview.
See results with 6 or 7 buttons (portrait and landscape orientation), and with 5 buttons including quickwizard (my current config) that improve readability
ImageButtonNoLabel

Philoul:
I tried to use existing code to show ContextMenu with a single click (Profile, TT and loop icon/button). It works but results is not exactly what I expected...
See below differences between long press (left) and short click (right) with this commit...
SingleClick

I don't know if It's easy with other settings or simple modification to keep same results... (Other solution I imagine but more complicated for me is to modify OverviewMenu to refactor menu creation (with the example of graph menu)
=> Maybe it's enough for testing

I increased a little bit the height of Profile and TempTarget buttons too...

Philoul:
Swipe to delete replaced by button
=> 2 possibilities to move rules, click and drag on left button or long press on rule and drag

Screenshot_20201129-174111_AndroidAPS

Philoul:
I propose these buttons for Temp Target Dialog (I prefer icons than text buttons)
=> In this proposal, when you click on a button, it only updates Reason and Target / Duration values (that way it's possible to adjust these values before validation)
=> It's a first draft for icons, it could be improved a little bit if you think it's necessary...

Screenshot_20201129-210226_AndroidAPS

Philoul:
Just updated 2 icons on latest commit

Screenshot_20201201-215109_AndroidAPS

Philoul:
@MilosKozak I'm testing all these proposal for several days and it seems good for me, the only point that should be improved I think is Context menu on short click of Profil, Target and loop buttons (see https://github.com/nightscout/AndroidAPS/pull/76#issuecomment-735255046)... We can test with current PR but I think the style of the "Long Press" menu is better for a low res phone.
I don't know if there is an easy way to keep the same result with a short click or if we should refactor these menus with a popup (like graph menu), but I'll not work on it (a bit complicated for me...) if you don't think this modification ok for you...

Philoul:
Last point, I don't know how to keep the text size fixed for the graphics scales (with the current code if we increase the text size in Android settings, the text is too big for the scales ...)

jamespaulley:
> I tried to use existing code to show ContextMenu with a single click (Profile, TT and loop icon/button). It works but results is not exactly what I expected...
> See below differences between long press (left) and short click (right) with this commit...
> SingleClick
> 
> I don't know if It's easy with other settings or simple modification to keep same results... (Other solution I imagine but more complicated for me is to modify OverviewMenu to refactor menu creation (replace contextmenu by a popup menu...)
> => Maybe it's enough for testing
> 
> I increased a little bit the height of Profile and TempTarget buttons too...

The new UI looks good to me. What is the problem with it? This is all great work by the way!

jamespaulley:
> I propose these buttons for Temp Target Dialog (I prefer icons than text buttons)
> => In this proposal, when you click on a button, it only updates Reason and Target / Duration values (that way it's possible to adjust these values before validation)
> => It's a first draft for icons, it could be improved a little bit if you think it's necessary...
> 
> Screenshot_20201129-210226_AndroidAPS

I think these are a great start and the UI is looking great in general. My thoughts on the buttons: The target icon in each button could either be reduced in size, desaturated, faded or removed; making way for the unique icon to be the focus. The hypo icon is the same as what's used for "xDrip, CGM and calibration" so I think it needs to be unique. Perhaps just the word "HYPO" would be clearest? I think we need to be careful with our use of colour theory too. Currently the choice of the colours in the buttons doesn't make much sense. The fact that CANCEL and CUSTOM are the same green colour suggests they are in some way connected, and I think that's confusing. We should either be using a single colour for all buttons, or the colour choices should imply meaning or difference between the functions.

jamespaulley:
> Two latest commits were to fix buttons layout with no labels (vertically centered)
> => I created a new SingleClickButtonImage button type in ui, but I had to rename buttons in no_label layout to avoid conflicts...
> => It works but I don't know if it's "clean"
> I think removing labels allow adding one or 2 buttons in overview.
> See results with 6 or 7 buttons (portrait and landscape orientation), and with 5 buttons including quickwizard (my current config) that improve readability
> ImageButtonNoLabel

All looking great. Could the Loop button come down a touch so its top is in line with the text to its left? Currently it's a bit higher than the text to its left, which feels a bit messy.

Philoul:
@jamespaulley thank's for your feedback
1) Menu : The problem is not in my phone (high res) but on low res phone. On emulator with Atom configuration, the single click menu is bigger and takes all the screen. If you open the menu by mistake, there is no place on the screen to click outside menu and remove it, the only solution is to select an item then click on cancel. That's why I think we should find a solution to keep current design smaller, and attach to button clicked (but I don't know if there is an "easy solution" without a complete refactor of current context menu...)

2) Icons : I'm agree that we can improve them...
- for cancel icon,  I added the same "cross" icon used in AAPS application (and elsewhere this cross has the same color than original icon, see TBR and Cancel TBR icons or Autosens Satus Icon enable/disable), that's why I kept the same TT icon with the green cross to be consistent...
- for Eatingsoon and Activity, I used the same icons that Meal (in previous Careportal tab) and Exercise. with the same color to be consistent, but these icons have a signification today, so I think we cannot use exactly the same, that's why I added the green TT icon too (I can try to change both size if it's better, Meal and activity icons are probably too small in my first proposals). Orange and Cyan colors are associated with carb/meal and exercise/activity so I think we should keep these colors
- for Hypo, I didn't find any "Hypo icon" in my database...  I don't think a text "HYPO" is a good idea (because of translations and length can vary according to selected language), I think icon is better! Currently it is not exactly the same than Calibration icon but I'm aggre that it's too close, you're right ! I'll propose something else , why not a single drop of blood with a down red arrow. I think red color is adapted because associated with low BG value, and when we are in hypo, a BG check is a good practice, that's why I imagine something with blood icon and red color....
- for Manual icon, the pencil icon seemed suitable to me, but I didn't want to keep the orange color associated with a note, that's why I put it in the same green as the TT icon.
-  desaturated, faded color will change the TT icon color, so if we change it, we should change all TT colors with the same green everywhere too (I wanted to be as close as possible with current design with my proposals...)

=> If you have other ideas, you can send me even a handmade drawing or photoshop design, I can easily redesign or convert a draw or bitmap in vector. I'll try to propose further improvements on my side. 

3) loop icon : I will take a look to see if I can improve it, but's it's not so easy because in some config we can have the label below loop icon, and there are lots of overlaps in design between loop icon, label and autosens icon to reduc the height of info area.  I'll try to improve current layout design but I'm not a specialist...

Philoul:
@jamespaulley see this new proposal for temp target icons and colors:
=> I found a tempttarget green in color pallet (less flashy than previous one...)
=> Hypo icon replaced with a down arrow added
=> TT icon size reduced and reason icon size increased

Screenshot_20201203-220706_AndroidAPS

AdrianLxM:
I'm sorry to say, but the icons for the TT are not my favourite. Mostly because I don't like the cross hairs.
Imho they are not self-explanatory like text. I'd add a label to them.

And this one just seems brutal to me: Cross hairs plus blood. 
image

@Philoul there seems to be quite some code duplication regarding the different buttons. These are just for presentation purposes? (The version with labels, that I prefer, is fully hooked out, e.g.). If both versions were to stay, it would be possible to use the same key as we actually only address those via their common superclass `View` (for registering click listeners, ...) in code. Also as they are custom views already we could simply use either _material button_ or style (actually: dynamically set the theme) them to have a label or not.

Philoul:
@AdrianLxM thank's for your feedback 
The remove of labels was proposed in #67 to be able to add more buttons. These redundancies are because I kept label for landscape orientation but remove them for portrait (except for quickwizard button)
=> I'm agree with your remark concerning these redundancies, I created duplication because buttons with or without label are not the same objects and not in the same layout... I don't know how to improve that (each time I tried to use same label in both layout, I had conflicts in code, or crash of aaps after launch) can you please help me (of course if finally we decide to always keep or always remove labels, that solves this problem of redundancies...)
=> After 2 or 3 weeks of testing, I personally prefer no label because the button heights are lower (especially if Quickwizard is visible and lots of buttons are enabled) (and it's probably worse in German or French than in English, and I tried to shorten the French labels as much as possible)

Concerning TempTarget icon : all my proposals were done with existing TT icon (and I admit that my first proposal was not the best ...)
=> If we can change this icon to replace the cross hairs with something better (I imagine a green icon close than basal temp icon), I can make further proposals (I prefer a good self understanding icon than texts that is very difficult to fit in small area in all languages)

AdrianLxM:
We are not really using the functionality of a Button but actually treat them as views.
The only real advantage that I see in a Button then is, that it is automatically screen readable if it has a label. If it hasn't we are losing that anyway.

So it wouldn't be much of a problem inflating our own custom view that centers the image if no label is there. It could also auto-scale the text size to automatically fit. Or simply hide the label in case the buttons become too small for it. Or use a theme attribute that can be dynamically set for the screen. Or... I've done a couple of those custom buttons for another project recently. Maybe we can have a quick chat on Slack to find the requirements and I can do a prototype?

Philoul:
See screenshot with latest TT icons
NewTTIcon

Philoul:
With this latest commit https://github.com/nightscout/AndroidAPS/pull/76/commits/9541e7d827f500d521f31492c2ccf57bb98566d1 I redesign single click menu with a Popup menu, so now with have the same "look" between a long press and a short click.
=> We can test and compare with this commit both solutions: 
- Short click on Profile, Target and Loop buttons (with Ok /Cancel dialog to confirm for direct actions)
- Long press on these buttons without the OK dialog (current 2.7.0 interface)

Note: with the "Short click menu", if you click on the title, it closes the menu (nothing happens with the "long press menu" if you try to click on the header)

@AdrianLxM I create redundancies in OverviewMenus file for single click menu (I duplicated and slightly modified the existing code for context menu to create the popup menu because I couldn't get the same results with single click and existing context menu, see https://github.com/nightscout/AndroidAPS/pull/76#issuecomment-737169355 and https://github.com/nightscout/AndroidAPS/pull/76#issuecomment-737456413, in my mind it's just for tests and to be able to compare both solution and decide which one is the best...

Philoul:
Latest version after fine tuning of design (thinner wave and color of wave aling with inRange green)
Screenshot_20201208-224924_AndroidAPS

Philoul:
Most of above proposal are included in dev now so I propose to close this PR

jamespaulley:
> @jamespaulley thank's for your feedback
> 
> 1. Menu : The problem is not in my phone (high res) but on low res phone. On emulator with Atom configuration, the single click menu is bigger and takes all the screen. If you open the menu by mistake, there is no place on the screen to click outside menu and remove it, the only solution is to select an item then click on cancel. That's why I think we should find a solution to keep current design smaller, and attach to button clicked (but I don't know if there is an "easy solution" without a complete refactor of current context menu...)
> 2. Icons : I'm agree that we can improve them...
> 
> * for cancel icon,  I added the same "cross" icon used in AAPS application (and elsewhere this cross has the same color than original icon, see TBR and Cancel TBR icons or Autosens Satus Icon enable/disable), that's why I kept the same TT icon with the green cross to be consistent...
> * for Eatingsoon and Activity, I used the same icons that Meal (in previous Careportal tab) and Exercise. with the same color to be consistent, but these icons have a signification today, so I think we cannot use exactly the same, that's why I added the green TT icon too (I can try to change both size if it's better, Meal and activity icons are probably too small in my first proposals). Orange and Cyan colors are associated with carb/meal and exercise/activity so I think we should keep these colors
> * for Hypo, I didn't find any "Hypo icon" in my database...  I don't think a text "HYPO" is a good idea (because of translations and length can vary according to selected language), I think icon is better! Currently it is not exactly the same than Calibration icon but I'm agree that it's too close, you're right ! I'll propose something else , why not a single drop of blood with a down red arrow. I think red color is adapted because associated with low BG value, and when we are in hypo, a BG check is a good practice, that's why I imagine something with blood icon and red color....
> * for Manual icon, the pencil icon seemed suitable to me, but I didn't want to keep the orange color associated with a note, that's why I put it in the same green as the TT icon.
> * desaturated, faded color will change the TT icon color, so if we change it, we should change all TT colors with the same green everywhere too (I wanted to be as close as possible with current design with my proposals...)
> 
> => If you have other ideas, you can send me even a handmade drawing or photoshop design, I can easily redesign or convert a draw or bitmap in vector. I'll try to propose further improvements on my side.
> 
> 1. loop icon : I will take a look to see if I can improve it, but's it's not so easy because in some config we can have the label below loop icon, and there are lots of overlaps in design between loop icon, label and autosens icon to reduc the height of info area.  I'll try to improve current layout design but I'm not a specialist...

Hi, sorry for being away for a while.

1. **Menu.** Whatever you think is best. Yes, on a small screen there should be the ability for the user to close the menu, whether that's a back button or a greyed-out area behind the menu.
2. **TT Icons.** Personally I feel like the iconography is getting too complex and confusing to look at. I think something like the icons below might be better. I think that 'cancel' should be first or last in the row, and should be a grey colour. I appreciate your take on the colours, but I think the colour theory throughout AAPS is a bit of a mess and as long as the colours of these new elements aren't immediately confusing I don't think we should worry about the consistency of the colours too much. I appreciate that the crosshairs icon is a bit aggressive, but I think it portrays 'target' more clearly than the waveform icon. I think the waveform icon is confusing as it looks like an audio waveform, and the straight lines look like basal rates.

PR #83, opened 2020-11-25
Change to material

osodebailar:
after brian add filter to preferences, i prepared the project for material components and change the preference filter field to material textfield.
Screenshot_1606328967
Screenshot_1606328958

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/PreferencesActivity.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/PreferencesActivity.kt
@@ -33,10 +33,6 @@ class PreferencesActivity : NoSplashAppCompatActivity(), PreferenceFragmentCompa
             override fun afterTextChanged(s: Editable) {}
         })
 
-        pref_filter_clear.setOnClickListener() {
-            pref_filter.setText("");
-        }
-
         title = resourceHelper.gs(R.string.nav_preferences)
         supportActionBar?.setDisplayHomeAsUpEnabled(true)
         supportActionBar?.setDisplayShowHomeEnabled(true)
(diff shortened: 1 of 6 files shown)


twain47:
I like the style but this results on black on grey text when you select the input box.
Also the word 'Filter' and the border end up overlapping which makes it hard to read.

I've not seen any other obvious problems while testing it out - looks like there is no impact on the rest of the system.

osodebailar:
> I like the style but the black on grey text when you select the input box is too hard to read IMO.
> Also the word 'Filter' and the border end up overlapping which also makes it hard to read.
> 
> I've not seen any other obvious problems while testing it out - looks like there is no impact on the rest of the system.

Colors can be set individual , but for now thats material default.
I will see if we can change the color.
Overlapping i see, but its material standard.  i will have a look if other people have solved this.

MilosKozak:
by changing AppTheme you change alot everywhere. it would need to review everything in app. by saying this i don't mean not to do it, but this is too simple

osodebailar:
> by changing AppTheme you change alot everywhere. it would need to review everything in app. by saying this i don't mean not to do it, but this is too simple

No problem @MilosKozak , we have time. I will fix the issue @twain47 told me and than you can do the review.

AdrianLxM:
Regarding the switch to Material: I like it but agree that it might affect some components.

The change mostly affect material components. We don't have many - especially as most enforce Material Design.
That would be the tab-switcher (TabLayout) on top in `activity_main.xml` or differently mapped buttons like the one in `overview_notification_item.xml`

What I've noticed is though: the main activity actually uses `AppTheme.Launcher` that has not `AppTheme` as parent so many parts wouldn't move to material.
Could you please try to change that - and then probably add the `colorOnPrimary`, `colorOnSecondary`, `colorOnSurface` correctly?

AdrianLxM:
Is this PR still being worked on by you @osodebailar?

PR #84, opened 2020-11-27
Objective 3: Rewrite of questions to make them clearer.

CapnGrumpy:
The initial remit of this project was to look at how the questions in Objective 3 of AndroidAPS were worded. It was suggested to 
Joel Milner by Milos that each change should be submitted as an individual pull request for approval/rejection.

As we, as a group, looked at the questions, it became clear that there were a number of areas where we thought improvements could be made to Objective 3 that would reduce some of the frustrations that users experienced whilst maintaining the requirement for users to read the documentation and understand what they were doing without making the process an overly easy 'box ticking' exercise.

The result of that has been a more comprehensive re-write than originally planned - including changing the order of the questions to put them into logical blocks. As such, individual PRs for each change no longer make much sense and we felt that a bulk change was more appropriate.

We all agree that the Objectives are important – they provide a structured learning environment for new users and are an important safety feature.  Since the introduction of the current Objectives, we see far fewer questions asking simple things that relate to basic use of the app - a contrast to other DIY looping systems.

However, there has been a certain amount of frustration regarding the Objective 3 questions which could be resolved by a revision. Our aim is not to dumb down the questions, nor to make the answers obvious, but to reword and reformulate them in such a way as to remove ambiguities so that users avoid over-thinking the answers or making errors because the questions were not phrased in a way that makes immediate sense to them - particularly those for whose first language is not English. 

The rewritten questions have been carefully formulated to cover all of the knowledge objectives implicit in the current version. We have attempted to retain the existing learning aims and objectives, covering the same topics but in a more logical order.  The new format (asking users to select answers from a choice of four for all questions) follows recommended educational practice for multiple choice examination questions and makes the logic underpinning the correct answer clearer. We believe new version will actively encourage users to read and to understand the documentation and prove their knowledge by answering them correctly.  

This work has been a collabotative effort between Joel Milner, Matthew Gray (CapnGrumpy), Gem Pickford-Waugh, and Craig Waugh.


MilosKozak:
OK. I understand huge PR. But is there a list what has been removed? Now it's not visible

MilosKozak:
I'm now sure if adding "Question XX" is a good idea. If we decide to add a single question, all of them must be renumbered and it means they will be marked for new translation

CapnGrumpy:
> 
> 
> I'm now sure if adding "Question XX" is a good idea. If we decide to add a single question, all of them must be renumbered and it means they will be marked for new translation

The original idea was to organise the questions into Topics, but as we ended up with several Profile questions we decided to go with question numbers so that users who were seeking help could say that they wanted help with a specific question number, rather than an entire topic.

Yes, if a new question was added, then the questions would need to be renumbered, but I was under the impression that crowdin was free for open source projects. Is this not the case?

CapnGrumpy:
> 
> 
> OK. I understand huge PR. But is there a list what has been removed? Now it's not visible

The Offline profile question was removed as there are a number of questions about profiles and we suspect that very few people use them when it's much easier to set them up in Nightscout.

The Noisy CGM question was removed. Anyone who is capable of setting AAPS up must surely understand that if the data fed into the system is unreliable, then the results will be too. As there are a total of 26 questions, this one didn't seem required.

guydavies:
I would consider both of those to be risky assumptions.

I never use NS for profiles because I found the sync and activation process
so fiddly. Doing it all locally from templates using auto-tune as input
worked better for me.

And given the number of questions I still see about unreliable readings on
CGM, I would suggest a question on that remains relevant.

AdrianLxM:
I really like the initiative.
It is targeting `master`.  Could you please target `dev`?

From looking at the files it seems a few things are missing though:

**offlineprofile_whatprofile**
As offline profile is the recommended one, we should really point people to it.
Do we have a security related question where we could add this?

**noisycgm_whattodo**
As the DIY-Looping community overlaps with the DIY-CGM community where solutions exist to prolong sensors. like @guydavies I see many questions on that topic. And usually a failing sensor happens in the worst moment when you already don't have much time. So having people actively think of what to do isn't a bad idea. 
 Maybe there is a better formulation? Can we reduce it to only the "emergency protocol"?

**basalhelp**
I like the official statement that you should talk to your diabetes team.
Maybe we can do this with a Yes/No - Question?

**The Exchange Units:**
Also very important is the knowledge regarding **BE (exchange unit)**. If I had to point at one issue that people have setting up their profile, it is this. We have over 1k (probably much more) users whose diabetes team does not use IC and often doesn't know how to derive IC.

I agree that the way we ask the question currently is suboptimal as people who don't know what BE is have to learn something that doesn't affect them. But If we cannot find a better formulation, I'd keep this in as is.

Ideally we do have an initial dialog that asks the user if she are used to using an Exchange unit (BE/KE/KHE?...) or g carbs when estimating food. If the answer is Exchange unit, we could dynamically add another question only covering this topic. @MilosKozak what do you think?

MilosKozak:
Crowdin is free but if you change any word whole sentence is invalidated and must be retranslated in all languages

MilosKozak:
exchange units are used here too. especially older poeple. i'd keep the question there too maybe with explanation what is it for people who have no idea

twain47:
How about numbering questions within topics 'Profile Question 1' or 'Profile (1)' ?

PR #86, opened 2020-11-28
BGSourceFragment UI tweaks and recognize history change

AdrianLxM:
1) Updates history if BG is invalidated.
2) Remove button and INVALID flag are mutual exclusive
3) Invalidating a BG reading will update the view
4) Some groundwork for RecyclerView to dispatch updates only to updated items

image
image

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
+++ b/app/src/main/java/info/nightscout/androidaps/db/DatabaseHelper.java
@@ -417,6 +417,8 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
         try {
             getDaoBgReadings().update(bgReading);
             openHumansUploader.enqueueBGReading(bgReading);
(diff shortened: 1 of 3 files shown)


MilosKozak:
Problem with BGReading tab was caching `Fragment` content by Tab manager.
If you display content of BG tab and then swipe out of it, content is cached and the fragment is unregistred from RX. when you swipe back cached content is displayed but outdated because fragment missed RX events
I believe this change wil not resolve the issue

AdrianLxM:
@MilosKozak this was not the issue I was trying to solve as I wasn't aware of it still existed after you added the `updateGui()` call to `onResume()`. Anyhow: it is solved now in this build.

What I an trying to solve:
AAPS didn't recognise the delete as history change and wouldn't re-calculate the Autosens/COB data.

Also instead of updating the data we swapped the whole adapter. Dispatching the change only to the updated data is not that important in that view, admittedly, but if we do that, RecyclerView will have some nice visual feedback for the user when updating the list. We could use it in other places as well.

PR #94, opened 2020-12-04
New careportal Icons for battery-, canula-, insulin- and sensor age

Rig22:
Here are some new icons for careportal. It was important to me that they fit in well with the previous design. The icons should appear elegant and clear

careportal

careportal_overview


ArthurusDent:
Why does the battery symbol use the symbol for on and off? That doesn't make a lot of sense to me. The battery isn't a switch. It's storage for electrical energy. If you look at Android's or iOS' symbols, they don't use that symbol either. I think they only use the symbol of a lightning flash, if the battery is being charged. I think the battery symbol with the on/off symbol would be confusing.

Rig22:
My thoughts on the battery symbol were as follows. When I take the battery out of the pump, the pump is off. When I insert it, the pump is on. On/off.

MilosKozak:
my opinion is:
these icons are using too different graphics to other icons used in AAPS (thin line vs thick of other icons)
if we decide to update it it must be whole set of icons and other must agree on it

ArthurusDent:
> 
> 
> My thoughts on the battery symbol were as follows. When I take the battery out of the pump, the pump is off. When I insert it, the pump is on. On/off.

Thank you for explaining it. This still doesn't make a lot of sense to me.

Rig22:
Maybe you are right. 🙂

PR #98, opened 2020-12-05
SVG Icons for pumps fragment tab

Rig22:
I built some svg pump icons and put them in the appropriate layouts.

danars
pod
veo
combo
insight


CaroGo:
Would love to see them integrated!

swissalpine:
Sorry for my question: What's the benefit of these icons?

Rig22:
Benefit, absolute nothing. Onlyo looking nice. 😉

osodebailar:
Very nice ,
WhatsApp about the optical Aspect to Place the svgs to the Top of the Page , so as i did in my gui Fork ?

Rig22:
Thank you. 🙂
I think both approaches have their advantages and disadvantages. I found that the graphic below gives the actual pump information more space without having to scroll if necessary.

swissalpine:
@Rig22 Thank you. Ok, something pretty.
Please don't move it to the top of the tab. It is as you've written: For me it would be annoying to scroll down just for looking at a pretty but useless graphic.

MilosKozak:
it would be nice to see then in ConfigBuilder in pump selection

core/src/main/res/drawable/ic_danars.xml
```diff
@@ -0,0 +1,64 @@
+<vector android:height="100dp" android:viewportHeight="300"
```
MilosKozak, line 1: should be placed under dana module

Rig22: Hello Milos. :) Thanks for your review.

I'm not sure I get it right. Should the graphic be moved to AndroidAPS / dana / src / main / res /? There is no drawable folder there. Should I create one? Sorry, I'm not a programmer.....

Rig22:
> it would be nice to see them in ConfigBuilder in pump selection. they would be usefull there

Thank you. :) 

Great idea with the Configbuilder. I'll have a look at it tomorrow. Let's see if I can do that?

Rig22:
@MilosKozak 

I looked at the Config Builder thing. But that's too difficult for my very limited programming knowledge. If others like it, someone else would have to build it in.

osodebailar:
graphic in bottom or top is not so important here. for me in top is the better idea ( in my gui fork i have the graphis in top because on the right side i integrate the bottom menu bottoms as the newer material fab buttons ).
So i am free where to place the icons in the old gui.

@MilosKozak 
the idea of placing icons in the configbuilder is a good idea.
i played with it a While and my suggestion would be :
extend the plugin datastructure with an icon resource for each plugin. The additional advantage is we can place the icons on the left side of the navigation menue in the navbar.
But i would to seperate this things.
A little PR for the icons on the pump side and than, because there are more things to do, a additional pr with the icon extension for the config builder.

@Rig22 if @MilosKozak  wants it, than i can help and do it.

osodebailar:
@MilosKozak   @Rig22 the solution can be look like this example.
Its created on my gui fork, but the look and feel here is the same as in the original gui.

@Rig22 the image size is fixed to 36x36dp , may be you have to create seperate icon for it for a better look and feel . the omnipod icon has to expand here

Screenshot_1607350931

Rig22:
> @Rig22 if @MilosKozak wants it, than i can help and do it.

Hi Andreas. :) Would be great. I added you as a collaborator...

PR #108, opened 2020-12-09
Delete Automation rule with button

Philoul:
I propose to replace delete on swipe by delete by a button (I tested both solutions and I think button is really better than swipe)...

See "Swipe / Don't section" in https://material.io/design/interaction/gestures.html#types-of-gestures

> Don't
> Avoid situations where a single gesture might produce two different results.
> https://kstatic.googleusercontent.com/files/d2d7b6dce2efa234cb0b51a0576fb7cfef2a80b0cacc7bb6a23f10537924575a5a03aaed0f0ec143d91ae1c15c3a15fc7e0704ca0c7826fa231ded1e87aa09c0

Screenshot to see buttons size and spacing : 
Screenshot_20201209-233532_AndroidAPS


StephenBrown2:
I've almost accidentally deleted some of my rules when swiping back and forth between panels, such that I removed the automation tab so I only have it in the hamburger menu list. This is definitely an improvement in my mind.

PR #112, opened 2020-12-11
Move Ago below Arrow

Philoul:
Move ago below arrow (and fix arrow position too low in latest dev)

Replace complex Linear layout by table layout for delta area (cleaner I think)
Split average delta (15min and 40min) in 2 textviews (I think 2 TextViews is cleaner than one TextView with CRLF)

Screenshot_20201211-101727_AndroidAPS


Philoul:
Note, I can improve the vertical position of Loop icon in same PR (currently overview_apsmode_text is most of time empty but visible...)
=> Commit done in my local branch but not pushed...

See proposal with or without text compared to above previous screenshot...
Screenshot_20201211-105242_AndroidAPS

guydavies:
I like both these proposals.  Thanks @Philoul.

MilosKozak:
does it work in nsclient tablet portrait and landscape mode?

Philoul:
I quickly setup a tablet in emulator to verify, see screenshot (I think it's good):
TestTablet

Philoul:
Hum, I think screenshot is with FullRelease version, (I don't know how to launch easily NSClient version on Emulator...)

Philoul:
Ok, not sure this is the easiest way to launch an NSClient on the emulator (I dragged and dropped an NSClient apk on the emulator...)
=> It looks to be good (but I don't know if this is what you expect...)
NSClient

PR #117, opened 2020-12-14
Short Click menu added in overview

Philoul:
Replace Context menu in overview (loop icon, profile and temp target buttons) by a popup menu (to avoid redundancy)
- Show menu on shortClick (with OK/Cancel dialog on direct actions like Disable loop, disconnect pump...)
- Show menu on longClick (without Ok/Cancel dialog as it is today with context menu)

In both case we can hide menu if we click on Header (new compare to current context menu)


Philoul:
Removed menu for Temp Target button and show TT Dialog (works on short or long click)
Cleanned code (in both overview menu and TT Dialog / Layout)

MilosKozak:
let's remove TT menu at all. show dialog directly

MilosKozak:
do you have an idea what to do with profile menu?

osodebailar:
@MilosKozak @Philoul 
you can do it in the same way . Open direct the change Profil dialog and in the dialog itself add an additional button to show the Profil.

MilosKozak:
@osodebailar it's different function

osodebailar:
@MilosKozak ok . My mistake . I thought you mean the Profil pill.

Philoul:
We could use short click to show Profile Switch dialog and Long Click to view profile

Philoul:
With latest commit I replaced Profile Menu by a short click (ProfileSwitch Dialog) and a Long click (View Current Profile)
=> Compared to previous solution, there is one click less to launch ProfileSwitch dialog and View Profile
I don't have a better idea at the moment ...

osodebailar:
And whats about my idea i Wrote before to Integrate a Button in the Dialog to view the Profile ? But the way you did it is also a good one. It follows @MilosKozak idea to integrate long click for Special functions

Philoul:
I think View Current Profile and ProfileSwitch is not the same function, so I'm afraid it's confusing to launch ProfileSwitch dialog to View Current profile (and it's one more click to view profile).
=> In this case Special function for Profile is "View Current Profile" (but I don't know if @MilosKozak has in mind other special functions for Profile Button...)

Philoul:
If I follow your idea, we can add a button in Profile Switch dialog too (on the righ of Profile selection spinner) to view selected profile...

MilosKozak:
i think  a functionality to display running profile directly without any other dialog and option is crucial. Just to make absolutely clear this is currently applied profile, even for unexperienced users

osodebailar:
Maybe with a text : "Current active profile"

Philoul:
@MilosKozak do you prefer a switch compare to my first proposal: short click to view current profile and long click to show profile switch dialog or do you have another idea?

MilosKozak:
> @MilosKozak do you prefer a switch compare to my first proposal: short click to view current profile and long click to show profile switch dialog or do you have another idea?

maybe this would be better. let's try it. no change for users

MilosKozak:
can you create icons for suspend/disconnect menu and replace menu with a dialog?

Philoul:
> can you create icons for suspend/disconnect menu and replace menu with a dialog?

It will take me a little while but I can of course propose something...

PR #121, opened 2020-12-17
Configbuilder icons for all plugins and further minor GUI changes

osodebailar:
consequent continuation of integration configbuilder icons.

Now each plugin has its own icon that can be used also in nav menu for example.
The pump icons were now fixed for a better look and feel in the configbuilder.

Therefore the nav menu bar now has a modern look and feel. 
The Header in the nav bar can be ued later to show the patient name for example.

As an further addition the name in the 3 dot menu 'plugin prefernce' was changed against  the plugin name itself ( Overview preference) for example.

The most icons were made by @Rig22 . 

@MilosKozak : this is not in progress. Ready to merge if you like it

Screenshot_1608218083
Screenshot_1608218091
Screenshot_1608218095
Screenshot_1608218103
Screenshot_1608218111
Screenshot_1608218117
Screenshot_1608218123
Screenshot_1608218128
Screenshot_1608218131


app/src/main/java/info/nightscout/androidaps/MainActivity.kt
```diff
@@ -266,9 +274,16 @@ class MainActivity : NoSplashAppCompatActivity() {
         return super.dispatchTouchEvent(event)
     }
 
+    private fun setPluginPreferenceMenuName() {
+        val plugin = (main_pager.adapter as TabPageAdapter).getPluginAt(main_pager.currentItem)
+        this.menu?.findItem(R.id.nav_plugin_preferences)?.title = plugin.name + ' ' + resourceHelper.gs(R.string.nav_preferences)
```
MilosKozak, line 279: this is not possible translate correctly
Use %s

danar/src/main/java/info/nightscout/androidaps/danar/AbstractDanaRPlugin.java
```diff
@@ -86,6 +86,7 @@ protected AbstractDanaRPlugin(
                         .mainType(PluginType.PUMP)
                         .fragmentClass(DanaFragment.class.getName())
                         .pluginIcon(R.drawable.ic_dana_rs)
+                        .pluginIcon(R.drawable.ic_danars_128)
```
MilosKozak, line 89: twice?

app/src/main/java/info/nightscout/androidaps/MainActivity.kt
osodebailar: Can you explain what you mean with %s ? The function here replaces the title with concetinating 3 Strings.
And ist worked well. 
I dont know what is Not possible ?

danar/src/main/java/info/nightscout/androidaps/danar/AbstractDanaRPlugin.java
osodebailar: Ohh sure , i Delete the one line.

osodebailar: line deleted

Philoul:
Cool, but just a question, why you don't use a black &white version of cgm app icon (xDrip, NSClient BG, Dexcom app, tomato...).
I only recognized Glimp icon...
I can proposed alternative (I already done vector version of several ones...)

osodebailar:
@Philoul The Glimp icon is the only one , which is a little bit different . Dexcom , tomato and so on should show the cgm itself and not the app.
This is done because myself uses the icons in my own GUI fork in the header .

The icon itself were made by @rig22
I am satisfied with the solution in the moment, but 
i am free and i think @MilosKozak should decide .

MilosKozak:
please remove the AAPS picture on top of menu. it's just wasting of space and force unnecessary scrolling

osodebailar:
> please remove the AAPS picture on top of menu. it's just wasting of space and force unnecessary scrolling

done

Philoul:
Hello @osodebailar See accucheck Insight pump icon (in menu)... A simple filled rectangle, do you think you can improcve it ?
Screenshot_20201220-195106_AndroidAPS

Do you have svg source files of all your new icons (I added svg files of all vector icons I created in \icons folder, that way it's easier to work if we want to make other icons from existing ones...)

osodebailar:
Hi @Philoul 
i would like to do it better ,but it's not possible.
The icons you See will automatic coloured ( tintcolor) by the navigation component ( material Navigation drawer) istself.
The icons you see are all the Same as defined in the configbuilder through the new Element in the plugin datastructure ( pluginIcon)

osodebailar:
And all Icons you See are svg files

Philoul:
Ok, so I understand that we have to "remove" some dark grey zones (for exemple reservoir, buttons or screen background) to keep a close design of "colored" icons if we want to improve the white activity icon... (I think it could work if we always have a dark grey background)

osodebailar:
A better way maybe is to try not to use special colors and 
try to color these areas transparent . 
For example the Insulin icon has a transparent part.
the assumption to say the background color is always a fixed color is not the right one.
Because many Android components change or can change background color a little bit ( if you set elevation to a component for example ).
The other point is that maybe the background can differ , if we extend to themes or night and light mode.
You can see , if you want , many examples in my GUI fork in the new_gui_to_dev branch . Here you can open the "set colors" menu entry in the 3dot menu.
For further discussion we can do this via email , FB messenger or any where you want

osodebailar:
Here you can see for example the attributes for the navigation drawer : https://material.io/components/navigation-drawer/android#anatomy

PR #128, opened 2020-12-19
Fix typo in russian translations

GermanBluefox:
Hello, yes I know, that translations must be changed via CroudIn, but no one cares about my suggestions there.

Can we fix this embarrassing error just in normal way as all programmers over the world it normally do? Via Gtihub PR.

Thank you.

P.S. just to understand what the words are fixed if they would be in english: Lad => Load, Cange the place => Change the place. I think every one who will see it english will fix it without doubts. In the same matter these words look for me in russian.

### DIFF
--- a/core/src/main/res/values-ru-rRU/strings.xml
+++ b/core/src/main/res/values-ru-rRU/strings.xml
@@ -136,8 +136,8 @@
     <string name="careportal_announcement">Оповещение</string>
     <string name="careportal_note">Примечание</string>
     <string name="careportal_question">Вопрос</string>
-    <string name="careportal_exercise">Нгрузка</string>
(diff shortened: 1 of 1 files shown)


rICTx-T1D:
When the Pull Request was approved,  every new CroudIn Update will be override it. So the way is very ineffective and complex.

I recommend to stop this PR and solve the problem of translation in source system.

GermanBluefox:
@AdrianLxM , @MilosKozak can you give me please access to accept translations at least in Russian and German? The editors on croudin are sleeping and I am tired to suggest 3 times the corrections and still see same errors in texts in every release. 
I am very responsible and active member of open source community and have own project with about 50k users. I know, what I am doing.

MilosKozak:
ask for join crowdin. it's descibed in wiki

PR #129, opened 2020-12-20
TT Dialog Add 5 buttons for quick selection of TempTarget's type

Philoul:
Drop down menu remains "as is", just corrected "Manual Selection" with default values when selected.
=> We can select reason either from the drop down menu or from one of the 5 buttons, then we can update the parameters or the valid dialog as we could do previously ...

See screenshot (in french)...
101545849-29974300-39a8-11eb-84df-0316dcc890f9


MilosKozak:
i like it.
i'd add comment to buttons to make it more clear and remove spinner at all

Philoul:
@MilosKozak Label added below icons, spinner removed and Cancel to Hypo buttons show confirm dialog directly.
I tried to add a SetOnLongClickListener on these buttons to fill values but it doesn't work and I don't know how to do that on this Dialog (code used in plugins fragments doesn't work here).
=> Can you just give me an example how I can make it work in this dialog please ?

Philoul:
Ok, after help from google, I could implement longclick to fill values.
Not sure that code is optimal but it works...

=> See below final result:
- Only short click available for Manual button (it reset to default values)
- On Short click on Cancel to Hypo buttons, => Set values and show confirm dialog
- On Long click on Cancel to Hypo buttons => Set values according to selected button
Screenshot_20201222-112059_AndroidAPS

guydavies:
This sounds awfully complicated compared to the previous/current approach.
Short clicks and long clicks on the same space having different actions is
not great IMHO.

I remember recently that someone (sorry, don't remember exactly who)
specifically complained that long clicks on the buttons in the main screen
were difficult for them.

It's always going to be difficult to please everyone but I think this
approach is a backwards step in terms of usability. Just my opinion,
clearly 😊

Philoul:
@guydavies 
The idea is short click is most often use to directly set selected TT
Long click is only if you want to cutomize values from one TT type...

guydavies:
Wouldn't another button for custom be easier than a long press of a specific button?  I know it creates smaller buttons but it follows better the current model but with consistently short presses.with confirmation. I know some don't like the current long press to access this function but once in, the behaviours of every button are consistent. Introducing special cases is not a good idea IMHO.

Philoul:
No need for another buttonsfor customization, you can select directly target and duration before validating dialog
Compare to previous interface, a short click on cancel, Eating Soon or Activity is enough to select values and validate dialog (one click less)
Long click is only if you want to slightly modify Activity, Eating soon or hypo default values before validating dialog
=> If you oftenly use default values for Eating soon, activity or hypo TT second solution is better
=> If you frequently modify default values before validating these predefined TT, first one is better
On my side I like both...

> I remember recently that someone (sorry, don't remember exactly who)
specifically complained that long clicks on the buttons in the main screen
were difficult for them.

These complaints was mainly on profile, TT buttons and loop button for showing context menu (I proposed this #117 PR to improve it) but this is a frequent action done on main screen (and I didn't see any complaints on QuickWizard long press to cutomize quickwizard...)

guydavies:
Fair enough 😁. I haven't seen other complaints here. I'll leave it to
others to decide if they want to raise any other issue.

MilosKozak:
It works well. It's definitely improvement.
I'd hide Cancel TT button if there is no TT running
I was too premature by removing Reason spinner. Now on long click, you don't see what you are going to enter.

So:
I'd put back spinner between Duration and Time but only for entering Reason (not changing preset values)
Move icons under or above Time (not sure what looks better)

I'm unsure regarding Manual button. Can be there real reason to use it? Is it used so often to assign special button? Would not be enough to allow Manual selection in spinner?

With this change entering TT is simple enough. I'd consider removing menu on overview at all and let this dialog run on both: click and longclick to not confuse people much.

Comments?

osodebailar:
@MilosKozak if you want to establish both, single and long click , than I would prefer that both ways are implemented in the same way with same dialogs. Different ways can confuse the operator

MilosKozak:
in the future I'd say use click for common controls and long click for extra/advanced actions
I'm not sure what would happen if we change this for TT only now thus I was thinking about both variants for now

Philoul:
Manual button is only to select reason and fill default values, if we add spinner, we can remove it (default value for manual is min TT value so not very important)
My other idea of ​​Reason was to make the background of the selected button lighter instead of adding the spinner (but we have to keep manual button selected by default with this solution)
I think we can disable "Long Press" for Cancel button too
For layout organisation, I'll try different solutions

osodebailar:
I would not suggest to change the button background in a special way. Buttons and other standard components should behave in the way Android does.

Philoul:
- Spinner put back between duration and time (and doesn't change any values)
- Buttons moved below Time (Best place because close to OK/Cancel buttons, and most of time when we use these buttons it closes dialog)
- Manual Button Hidden and Cancel is shown only if TT is running
- Additionnaly I removed Cancel from Spinner menu and I removed long press too for cancel

Tell me if you are ok with this proposal, (I have one remaining commit to clean unused code/button)
Screenshot_20201223-194553_AndroidAPS

osodebailar:
@Philoul very good work . well done. as an attachment you see the solution for my gui fork to fit different theming in light and dark
Screenshot_1608757623
Screenshot_1608757645
Screenshot_1608757676
Screenshot_1608757707
Screenshot_1608757500
Screenshot_1608757520
Screenshot_1608757546
Screenshot_1608757590

PR #137, opened 2020-12-24
Increase height of Compact Tab and Profile/TT row

Philoul:
Some users have difficulties in Compact mode to select the 3 dots or hamburger menus due to height too small.
=> I propose to use same Height for Compact tabs than Standard tabs
=> I increased a little bit height of Profile / Target buttons for the same reason

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/MainActivity.kt
+++ b/app/src/main/java/info/nightscout/androidaps/MainActivity.kt
@@ -225,7 +225,11 @@ class MainActivity : NoSplashAppCompatActivity() {
         if (sp.getBoolean(R.string.key_short_tabtitles, false)) {
             tabs_normal.visibility = View.GONE
             tabs_compact.visibility = View.VISIBLE
-            toolbar.layoutParams = LinearLayout.LayoutParams(Toolbar.LayoutParams.MATCH_PARENT, resources.getDimension(R.dimen.compact_height).toInt())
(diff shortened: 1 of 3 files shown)


osodebailar:
Do you have to pictures with before and after ?

Philoul:
See screenshots below (in french)
Screenshot_20201224-115432_AndroidAPS

Philoul:
If the height increase is too much for compact tabs, I can just increase the default compact height a bit (currently 30dp, something between 40dp and 48dp might be a good compromise)

osodebailar:
I think it's a good solution . 
It's nearly the same as I did in my bottom toolbar in my GUI fork. And as I know all users are very satisfied with the size .

I would prefer your solution .

MilosKozak:
how does it look like with short names used?

Philoul:
In my first proposal, Heigth of Compact menu was the same than Standard menu (but I think it's too important), so I made a second commit to increase a little bit Compact height (see in screenshot below on the left current heights, in the middle first version and on the left  with latest commit.
Screenshot_20201225-164639_AndroidAPS

AdrianLxM:
Up until Android 9 I'd have said: For the compact one I'd go with the first.

Now with swipe gestures to go back on newer Android versions people might actually want to press the burger button. As the gesture to open the drawer and "back" conflicts.

PR #145, opened 2020-12-27
Insight icon

Philoul:
@osodebailar What do you think of this little modification of the Insight icon for the plugin (works because the background is dark)

Screenshot_20201227-215758_AndroidAPS

svg files you sent me for all pumps included in this PR too


osodebailar:
Fine . Accepted from side.
What's about modification of other pump icons for the menu ?

Philoul:
> Fine . Accepted from side.
> What's about modification of other pump icons for the menu ?

I'll take a look on other pump icons.
Plugins icons are very small, I think we can simplify them compared to big size one to improve a bit more results...

osodebailar:
There is Not a Problem to let the other pump icons as they are .
I would Not simplify the icons that are also Used in configbuilder .

Philoul:
I agree, we can easily recognize other pumps with the shape.
Only first version of Insight icons was necessary to be improved to recognize a pump in menu.
So no more work on it...

AdrianLxM:
@MilosKozak can we merge this before the next release? Seems like a good idea to make the icon tintable.

PR #170, opened 2021-01-02
enablebolusadvisor_summary

xJoe:
Few more details to make it easier to understand without reading docs.


rICTx-T1D (commented):
@xJoe It is possible to change here, but I think many translators has translate this in their own language. No algorithm will reminder the translators to correct this.

xJoe:
I am aware of this as I manage AAPS docs in Crowdin. But change of the original string will inform all translators that they have to have a look at this string.

rICTx-T1D:
> But change of the original string will inform all translators that they have to have a look at this string.

That's new for me, ok I learned :-)

xJoe:
There is no direct info by mail but Crowdin shows that something is new or changes. This is especially obvious if language was at 100% before.

MilosKozak:
create new PR based on dev. this one is editing gradle too

xJoe:
@rICTx-T1D sent you a message on gitter concerning watchface for docs.

PR #181, opened 2021-01-03
Status lights battery

xJoe:
added "pump" as there is also sensor battery as of AAPS 2.8


xJoe:
If possible settings in preferences could be sorted the way they are displayed in actions tab.

app/src/main/res/values/strings.xml
```diff
@@ -520,17 +520,17 @@
     <string name="key_statuslights_sbat_critical" translatable="false">statuslights_sbat_critical</string>
     <string name="statuslights_sbat_critical">Threshold critical sensor battery level [%]</string>
     <string name="key_statuslights_bage_warning" translatable="false">statuslights_bage_warning</string>
-    <string name="statuslights_bage_warning">Threshold warning battery age [h]</string>
+    <string name="statuslights_bage_warning">Threshold warning pump  battery age [h]</string>
```
StephenBrown2, line 523: extra space:
```suggestion
    <string name="statuslights_bage_warning">Threshold warning pump battery age [h]</string>
```

Issue #156, opened 2020-12-30
Profile helper text button text dark on dark theme


guydavies:
Transferred from MilosKozak/AndroidAPS at request of @andyrozman

When using the profile helper in the current Dev build and every time I've previously tested, the button at the top of the screen shows black text on a dark background so it is barely visible.

I suspect this is just not adjusting at all to the dark mode.

https://photos.app.goo.gl/eNxD8nNpie5RxY1e9

osodebailar:
test it on 2.7.3-dev and button show text and icon correct
![Screenshot_1609347737](https://user-images.githubusercontent.com/25795894/103368954-34d22000-4ac9-11eb-81d7-9fd2458769a2.png)
![Uploading Screenshot_1609347794.png…]()

Philoul:
The issue on this layout (minor) is for me the + and - buttons (black square around them and not centered vertically for "+" char)

guydavies:
Agreed. My issue does now seem fixed but I also see the minor issue
described by Philoul. It's quite irritating once pointed out to me 😉

Philoul:
It's link to ProfileHelperAppTheme style, but I don't know how to fix it.
I tried to use default AppTheme style (but AAPS crash), and I don't find which properties and setting to add/modify in ProfileHelperAppTheme to fix it.

rICTx-T1D:
> It's link to ProfileHelperAppTheme style, but I don't know how to fix it.
I tried it too and I don't solved it :(

But the vertical-alignment of labels are improved.

AdrianLxM:
Issue behind this: If the theme is a Material theme, the general Button will become a MaterialButton.
On a `MaterialButton` the `background` attribute alone will not work. `app:backgroundTint` should do it.

AdrianLxM:
:) Thanks @guydavies for taking this over. Fixed, I hope.

PR #183, opened 2021-01-04
Fix for issue #156 (correct vertical alignment of label)

rICTx-T1D:
See also https://github.com/nightscout/AndroidAPS/issues/156
* correct vertical alignment of label
* remove comment of create-date and created by


Philoul:
For vertical alignment, I prefer keeping `android:layout_height="wrap_content"` in TextView and adding `android:gravity="center_vertical"` in TableRow

That way if you have long text (depending on translations), you keep full text (not vertically truncated), and all objects of the row are centered vertically)

rICTx-T1D:
> For vertical alignment, I prefer keeping `android:layout_height="wrap_content"` in TextView and adding `android:gravity="center_vertical"` in TableRow
> 
> That way if you have long text (depending on translations), you keep full text (not vertically truncated), and all objects of the row are centered vertically)

Thank you for your comment and code improvement. I have change it.

PR #184, opened 2021-01-04
Layout improvement

Philoul:
datetime vertical center for languages with "long translations" (more than one line)
Add marging for sort icon of automation (icon currently too close to border)


Philoul:
vertical alignment of time with label (see below in french label for time is too long for one line and not centered vertically) (on low res screen it could be on 3 lines so result is worse)
Screenshot_20201223-224109_AndroidAPS

with PR label vertically centered with time
Screenshot_20210105-011135_AndroidAPS

For sort icon marging, see result with PR
Screenshot_20210105-011158_AndroidAPS

PR #201, opened 2021-01-08
Correction xDrip -> xDrip+

xJoe:


MilosKozak:
hmmm this will lead to many unapproved strings.
I'll merge in next version

guydavies:
I know that Crowdin is the process, but is there not a way to do a global
search and replace along the lines of s/xDrip /xDrip+ /g and then set all
those changed translations to approved automatically?  It would definitely
be a useful feature request to Crowdin if it's not possible.

xJoe:
That's true. I can understand your point, @MilosKozak.

MilosKozak:
@AdrianLxM touched the same string so I merged this now

PR #243, opened 2021-01-19
Add friendly relative date display

jhaydraude:
This PR adds a method to DateUtil.java that takes a date and returns a string.

If the date is more than 1 week before or after today, then it returns the original DateUtil.dateString(date)
If the date is today, tomorrow or yesterday then it returns "Today", "Tomorrow" or "Yesterday"
If the date is less than a week out, but not today, tomorrow or yesterday, then it returns "Next (Day Name)" or "Last (Day Name)" e.g.: Next Wednesday 

PR also includes a usage of this method in the Omnipod Overview tab as shown in the screenshot.

Issue: 
https://github.com/MilosKozak/AndroidAPS/issues/3005 

Screenshot_20210119-173028_AndroidAPS

### DIFF
--- a/core/src/main/java/info/nightscout/androidaps/utils/DateUtil.java
+++ b/core/src/main/java/info/nightscout/androidaps/utils/DateUtil.java
@@ -5,6 +5,8 @@ import android.content.Context;
 import androidx.collection.LongSparseArray;
 
 import org.joda.time.DateTime;
+import org.joda.time.LocalDate;
+import org.joda.time.Period;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 import org.joda.time.format.ISODateTimeFormat;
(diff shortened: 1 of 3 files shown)


core/src/main/java/info/nightscout/androidaps/utils/DateUtil.java
```diff
@@ -169,6 +171,46 @@ public String dateStringShort(long mills) {
...
+                break;
+            case -1:
+                friendlyDate = resourceHelper.gs(R.string.yesterday);
+                break;
+            default: {
+                // When plus or minus a week, but not today, tomorrow or yesterday, display "Next <DayName> or Last <DayName>"
+                SimpleDateFormat format = new SimpleDateFormat("EE",  Locale.getDefault());
+                String dayName = format.format(date);
+               // String modifier = diffPeriod.getDays() > 0 ? resourceHelper.gs(R.string.next) : resourceHelper.gs(R.string.last);
+               // friendlyDate = modifier + " " + dayName;
```
AdrianLxM, line 201: Do you think this commented out lines will be needed in the future?
If not, I'd delete them

omnipod/src/main/java/info/nightscout/androidaps/plugins/pump/omnipod/ui/OmnipodOverviewFragment.kt
```diff
@@ -542,6 +542,19 @@ class OmnipodOverviewFragment : DaggerFragment() {
         }
     }
 
+    private fun readableZonedTimeRelative(time: DateTime): String {
```
AdrianLxM, line 545: Not necessarily for this PR, just a remark: There is quite a lot of general format utility functions in a Fragment.
We could move them to DateUtil as well.

AdrianLxM (changes requested):
Looks like a good addition. I only wonder how well this behaves around midnight?
There are two commented out lines of code. Can they be removed?

MilosKozak:
@jhaydraude can you finish this PR?

Issue #101, opened 2020-12-08
Make "INFO: Pump unreachable" notification actionable?


StephenBrown2:
As a user, I would like to be able to tap the "INFO: Pump Unreachable" notification that pops up and go directly to the "Medtronic" tab view.

I can see the `PumpUnreachable` `MedtronicNotificationType` here: https://github.com/nightscout/AndroidAPS/blob/bc43a926f9fce29eb3f3c262b4b6e955da088962/medtronic/src/main/java/info/nightscout/androidaps/plugins/pump/medtronic/defs/MedtronicNotificationType.java#L12 is used here: https://github.com/nightscout/AndroidAPS/blob/bc43a926f9fce29eb3f3c262b4b6e955da088962/medtronic/src/main/java/info/nightscout/androidaps/plugins/pump/medtronic/MedtronicPumpPlugin.java#L475 and here: https://github.com/nightscout/AndroidAPS/blob/bc43a926f9fce29eb3f3c262b4b6e955da088962/medtronic/src/main/java/info/nightscout/androidaps/plugins/pump/medtronic/MedtronicPumpPlugin.java#L570, but not being an app developer, I'm not sure how to add an intent to the dismiss/tap action. Possibly a check in https://github.com/nightscout/AndroidAPS/blob/bc43a926f9fce29eb3f3c262b4b6e955da088962/medtronic/src/main/java/info/nightscout/androidaps/plugins/pump/medtronic/util/MedtronicUtil.java#L235-L237?

andyrozman:
Notifications are just that... Notifications. They give you information, and then you can go and do something. 

Pump Unrechable notification is displayed when command can't be sent to pump.
 
Most of commands, will be retried automatically in 5 minutes (when next CGMS data comes in)...

ArthurusDent:
I've also wondered why I cannot tap on the notification and get at least straight into AAPS. That kind of breaks the typical Android behaviour. Getting straight into the Medtronic tab would be even better.

StephenBrown2:
Related, perhaps the notification can also be automatically dismissed when the pump is reachable again, similar to how https://github.com/nightscout/AndroidAPS/pull/118 added automatic dismissal of some notifications.

Alamo04:
@StephenBrown2 
I want to know that there was a notification,  if I missed one.

StephenBrown2:
[...]

PR #258, opened 2021-01-23
open app on notification press

AdrianLxM:
This will allow to open the app on tap of (system) overview notifications.
https://github.com/nightscout/AndroidAPS/issues/101

Also: small cleanups - removing code duplication.
Also: make system notifications the default option.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.java
@@ -449,14 +449,14 @@ public class LoopPlugin extends PluginBase implements LoopInterface {
                             && resultAfterConstraints.carbsReq >= sp.getInt(R.string.key_smb_enable_carbs_suggestions_threshold, 0)
                             && carbsSuggestionsSuspendedUntil < System.currentTimeMillis() && !treatmentTimethreshold(-15)) {
 
-                        if (sp.getBoolean(R.string.key_enable_carbs_required_alert_local, true) && !sp.getBoolean(R.string.key_raise_notifications_as_android_notifications, false)) {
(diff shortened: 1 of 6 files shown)


StephenBrown2:
Should this have been merged into dev?

StephenBrown2:
@AdrianLxM this is great! It opens the app when I tap the notification. However, two things are missing. One that I didn't realize I was expecting, and the other that I'd hoped for in the original request:
- The notification is not dismissed on tap
- The app opens to the main activity, rather than the activity that relates to the notification (Pump Unreachable -> Medtronic tab)

I'm sure the latter is much more complex than the former, especially if the notification is coming from NS and not the app.

AdrianLxM:
@StephenBrown2 it will open the app. So most likely you will be directly on the Overview where the in-app notification is shown that gets mirrored by the system notification. In the Overview you can mute the notification.

Navigating to a certain plugin is hard. It could be in different places in the app - or even another plugin might make much more sense. E.g. if no CGM values arrive the overview again.

So I went for the MVP here.

StephenBrown2:
> Navigating to a certain plugin is hard. It could be in different places in the app - or even another plugin might make much more sense. E.g. if no CGM values arrive the overview again.
> 
> So I went for the MVP here.

That's what I figured. 👍🏻 

> In the Overview you can mute the notification.

But this won't dismiss the system notification (or at least, it hasn't in my recent testing)

AdrianLxM:
> But this won't dismiss the system notification (or at least, it hasn't in my recent testing)

Would you mind trying to reproduce that and open another issue if it still happens. I fear in an already merged PR this will get lost.
I'll then take a look.

Thanks!

StephenBrown2:
Sure thing, I'll watch for it and create a new issue when it happens.

PR #279, opened 2021-02-04
Update strings.xml

aendes:
Changed mispelling


MilosKozak:
you must do it in crowdin. this would be overwritten

PR #331, opened 2021-02-13
irish langugage

andyrozman:
- added irish language back into selection box (was removed some time ago


AdrianLxM:
We currently have 5% of all strings translated.
Do you know of someone who would help getting Irish translations to a higher level?

image

Issue #299, opened 2021-02-09
Carb time alarm in BW not working


Ashley112885:
This feature does not appear to be working for several users according to Facebook comments. Does not appear to be a permissions issue. I've tried it on both a Jelly 2 (Android 10) and Pixel 3a (Android 11)

reijoahola-nightscout:
Works in Samsung Galaxy S9 / Android 10.

happygirlnyc70:
Does not work for Pixel 4a 5g or Pixel 3XL

susannaca:
It does not work in Mi A3, Android 11

AdrianLxM:
This PR adds a second way to get the alarm - via the system alarm app (countdown/kitchen timer): https://github.com/nightscout/AndroidAPS/pull/333

Issue #292, opened 2021-02-08
feature request 2.8.2: amend alarm of prebolus to carb menue


cheers2:
Carb Alarm for prebolus is only available in the calculator.
Please amend it also to the carb input screen.

Many thanks

AdrianLxM:
@cheers2 this PR includes it: https://github.com/nightscout/AndroidAPS/pull/333

cheers2:
@AdrianLxM thx for informing and fixing

PR #333, opened 2021-02-13
Adrian/system carb timer

AdrianLxM:
This PR does 3 things:
### **1) Sets the system countdown timer if the user selected system notifications.** 

If not, it will use the Automation reminder.
This has the advantage, that it integrates well with the system notifications and some smart watches.
`scheduleEatReminder` always stays as `AutomationEvent` as it has other trigger events like deltas.
This should be at least a proper workaround for https://github.com/nightscout/AndroidAPS/issues/299

### **2) Refactors the carb reminder generation to an own class**
Can be used in other dialogs as well. E.g. for the carb dialogs

### **3) Adds a carb reminder to the carbs dialog as well**
Corresponding feature request: https://github.com/nightscout/AndroidAPS/issues/292

Video of system timer:
https://user-images.githubusercontent.com/9692866/107861295-41c28d00-6e45-11eb-8e31-a62f3476bd3e.mp4

Screenshot of new dialog
image


MilosKozak:
definitely better than failing system dialog
I'd not use only for `key_raise_notifications_as_android_notifications` but all the time. can we do the same for bolus advisor to have it unified (ie start in 5 sec if condition met)?

AdrianLxM:
@MilosKozak I've done the simplification to not use `key_raise_notifications_as_android_notifications`.

The alarm of the advisor - I'll think of something and add it in a separate PR.

Philoul:
Just to be consistent with Wizard dialog, the checkbox is on the right of number picker in your proposal of carb dialog, and on the left (between "Time offset" label and number picker), in wizard dialog...
I think it's better if in both dialog the checkbox and associated icon are at the same place.

AdrianLxM:
@Philoul coming up soon :) Thanks for the suggestion.
image

Philoul:
just switch icon and check box and it will be perfect 😁

MilosKozak:
tests are failing

AdrianLxM:
@Philoul 
image

@MilosKozak nice catch, thanks. was only the test setup.

PR #346, opened 2021-02-18
Fix user Entry

Philoul:
Little bug fixed (i2 view always visible even if property=0)
I Improved layout (0 or "" views hidden to reduce and have regular spaces)
**I tried to translate labels, but I didn't find how to replace english labels by a translated string...**

See results below
Screenshot_20210219-001907_AndroidAPS


MilosKozak:
simple & easy, thanks!

Philoul:
@MilosKozak I think I understood for user entry translation.
It's currently hard coded in english in each uel.log command.
I propose 2 solutions.
1) replace hard coded string by a translated string (user entry is recorded translated in database)
2) create key strings of current english string (no change in database), and convert key to translated string in treatment's user entry fragment

I prefer second solution (user entry always available in selected language)...
Am I right in my analysis and is it ok for you ?

PR #353, opened 2021-02-20
WIP User Entry translation

Philoul:
With this first commit, I created keys for each `uel.log` entry
- 82 keys in app module
- 3 keys in core module
- 1 key in dana module

Note: the 3 keys of core module are not consistent with others (lowercase, long text string)
=> It's the only strings I modifyed a little bit to be consistent with others
=> I create 3 strings (a bit shorter than previous one) for "translation"

Some' other comments:
- RECONNECT key is missing (RESUME key is used for RECONNECT action) => should we add it ?
- similar user entry texts found : "TT REMOVE" and "REMOVED TT" => should we merge or not ?

With the list of created keys, if you have other comments tell me...

I propose to keep in database current uppercase english text as keys for each user entry and create only some new strings for treatment user interface (with a function to convert key => user interface string)
=> the advantage is that all user entry entries will be translated according to language selection
=> if we want later to build statistics on dedicated user entry, we can use key stored in database

Next step is in `TreatmentsUserEntryFragment` (creation of convertion function), and create all missing strings for translation


MilosKozak:
unify it
this db table is going to be record of user interventions. i'm still undecided if to translate it
what i was planning is export to text/csv file for analysis ... if you are looking for work :)

MilosKozak:
and if translations then probably store it in db as an enum and translate only for display

Philoul:
@MilosKozak Can you take a look on this proposal of enum class (name and content) ?
=> Cleaning made to remove duplicate entries
=> I tried to organize the list for readibility
=> I added some missing actions (I think) or potential futur actions (EXTENDED_CARBS, RECONNECT, CANCEL_TT, OBJECTIVE_STARTED, EXPORT_DATABASES, IMPORT_DATABASES)

Philoul:
My first proposal is not good, I'll analyse trendarrow enum class to understand how it works...

Philoul:
@MilosKozak With latest commit, enum class is better, but I can't build because of ressources.
=> database module don't have the resources folder and strings.xml files
=> I put uel strings in core module, but I can't access to them from database.

MilosKozak:
ok, if you go this way i'd do one more change.
not store stupid doubles and int but list of classes holding value and unit

MilosKozak:
for resource issues, create an extension in core module for enum -> string conversion
check trendarrow for example

Philoul:
> for resource issues, create an extension in core module for enum -> string conversion
> check trendarrow for example

@MilosKozak Sorry, I don't understand your answer...
- TrendArrow don't have external resources (all strings are hardcoded inside GlucoseValue.kt file)

I tried to add in build.gradle of database module an additional dependency:
```
    implementation project(':core')
```
When I do that, all strings included in strings.xml file of core module are found in UserEntry.kt file with `R.string.xxxx` , but I cannot build the app
```
Circular dependency between the following tasks:
:core:generateFullReleaseRFile
\--- :database:generateFullReleaseRFile
     \--- :core:generateFullReleaseRFile (*)
```
I don't have a string.xml files in database module to move my strings from core module to database module (and I don't know how to create res folder and strings files)

Philoul:
=> I can build, (it was quite hard...) but it's too late to test it on my dev phone to check if it works or not...
Now that I understand a bit more how everything works in database, I'll try to see if I can improve them with units
=> Where is for you the best place for this new "ValueWithUnits" class ? (embedments folder ?)
=> Do you have in mind an example in existing database code to help me ?

Philoul:
General comment on convertor file with line below, if a string in database doesn't exist in enum key, then AAPS crashed immediatly (I just copied existing lines in all typeconvertor
```
fun toAction(action: String?) = action?.let { UserEntry.Action.valueOf(it) }
```
Crash report on the first line of my UserEntry database
```
Caused by: java.lang.IllegalArgumentException: No enum constant info.nightscout.androidaps.database.entities.UserEntry.Action.ACCEPT TEMP BASAL
```
I had to patched it in Converters file by the line below
```
fun toAction(action: String?) = action?.let { UserEntry.Action.fromString(it) }
```
Now it works, and you can see results of database, on the left with entries done from this PR with enum (but shown with current dev branch, we can see hardcoded strings and key strings for latest entries), on the right, same screenshot with current PR (keys are correctly converted to translatable Strings, and previous entries that doesn't exist as enum key are converted to "UNKNOWN"
Screenshot_20210222-144050

Philoul:
Note, layout still need improvements (for profile switch with long strings, the view in constrained to the right, and if we swipe up and down, "s" disappears...

MilosKozak:
we can have a cardview with text under date nad label

Philoul:
@MilosKozak 
> ok, if you go this way i'd do one more change.
> not store stupid doubles and int but list of classes holding value and unit
> like:
> uel.log(UserEntry.Action.TT_EATING_SOON, ValueWithUnits(eatingSoonTT, MMOL), ValueWithUnits(eatingSoonTTDuration, MINUTES))

I agree, I propose to always store in database values in mgdl units, and add 4 text fields to store the different units used for each value recorded
I'll add somewhere an enum class to manage the different units used in app (the key name for database and units convertion and Strings references for UI), a ValueWithUnits class to make the interface between uel.log and database, and probably methods in UserEntry class to manage requested information: raw data (int or double) stored in database (mgdl unit if BG value or TT...), Strings data with units converted or not according to data type and default selected unit in app.
I'll try to propose you something...

MilosKozak:
store list of ValueWithUnits
https://stackoverflow.com/questions/44986626/android-room-database-how-to-handle-arraylist-in-an-entity
do not fix it to 4 values

Philoul:
See #379 with ValueWithUnits

PR #364, opened 2021-02-22
More Omnipod Dash preparation + some Eros fixes

bartsopers:
- Move several elements from module omnipod-eros to module omnipod-common
- Rename several Omnipod Eros classes and resources in order to better distinguish during development
- Add some Dash classes
- Fix battery age display for Omnipod Eros
- Fix wrong type casts in Omnipod Eros wizard (introduced with view bindings)


rICTx-T1D:
Is a approximate timeline present, when omnipod dash can supported by AAPS?

bartsopers:
> Is a approximate timeline present, when omnipod dash can supported by AAPS?

Not at the moment, sorry.

PR #379, opened 2021-02-27
UserEntry ValueWithUnits

Philoul:
@MilosKozak It's just a draft of ValueWithUnit to know if I'm in the right direction
=> it's a sub branche of https://github.com/nightscout/AndroidAPS/pull/353 (when all work will be done and ok, I'll merge this branche in it)

You  can see examples of ValueWithUnit only in CalibrationDialog.kt and CarbDialog.kt
Currently there are some redundancies in UserEntryLogger.kt (but I don't know if we can improve it...)
=> A convertion from mmol/mg/dl is done in TreatmentsUserEntryFragment to always show data in selected units
=> Formated Strings (with parameters) are recorded as "Strings data with units" (can be shown in selected language in UserEntry tab (see examples in SmsCommunicatorPlugin)
=> No other convertion done currently (see opened questions below for time units)
=> Any number of ValueWithUnit (double, int and long accepted) can now be included in database
- [x] All other call of uel.log have to be updated
- [x] Code for migration of UserEntry table has to be done (I propose to remove existing table created from v1 to v2 and create a new one in v3 to v4)
- [x] Increase version of Database (3->4)
- [x] TreatmentsUserEntryfragment has to be updated with ValueWithUnits to show values with translated units (see question below in we include a label or not for each value)
- [x] Tested yet and works (on my dev phone)
- [ ] Add Careportal additional information (for example Bg value if BG Check, Time and duration if Activity...)
- [ ] Create text/csv export

Minor problem:
- Units are not refreshed when changed in preferences (I tried to add a rxBus subscription but I don't now how to "updateGui" with database
- When I have long a long list of user entries, every information is shown on first page, but if I scroll to the bottom, then back to the top, some informations (usally notes, profile names...) may disappear


core/src/main/java/info/nightscout/androidaps/logging/UserEntryLogger.kt
```diff
@@ -18,20 +20,62 @@ class UserEntryLogger @Inject constructor(
...
+            )
+    }
+
+    fun log(action: Action, vararg listvalues: ValueWithUnit) {
+        val values = mutableListOf<ValueWithUnit>()
+        for (v in listvalues){
+            var vConverted = v
+            // Convertion to always store all values in the same units in database
+            when(v.unit) {
+                Units.Mmol_L -> { vConverted = ValueWithUnit(v.dValue * Constants.MMOLL_TO_MGDL, Units.Mg_Dl)}
```
MilosKozak, line 52: as this only for viewing i'd prefer store original units

app/src/main/java/info/nightscout/androidaps/dialogs/CarbsDialog.kt
```diff
@@ -224,7 +225,7 @@ class CarbsDialog : DialogFragmentWithDate() {
                         }
 
                         eatingSoonSelected -> {
-                            uel.log("TT EATING SOON", d1 = eatingSoonTT, i1 = eatingSoonTTDuration)
+                            uel.log(Action.TT_EATING_SOON, ValueWithUnit(eatingSoonTT, if (units == Constants.MMOL) Units.Mmol_L else Units.Mg_Dl) , ValueWithUnit(eatingSoonTTDuration, Units.M))
```
MilosKozak, line 228: yes, exactly
create another constructor to bypass `if (units == Constants.MMOL) Units.Mmol_L else Units.Mg_Dl` for MMOL/MGDL units. Pass only `units`

MilosKozak:
> Opened question :
> 
> * I don't record for time values which kind of information is recorded (for example minutes and hours can be time offset or duration), but it could be usefull for fragment if we want to labelize each value (time units could be splitted to in include type, for ex. M replaced by M_Duration and M_TimeOffset to do that)
no need to distinguish
> * Should I always convert hour to minutes ?
no, keep as user enter this information
> * ValueWithUnit and Units classes are included in UserEntry Class, but I saw data and embedments folders, should I move them in one of these folder ?
> * Fragment: should we labelize some values or not? (15 , 15m, Duration: 15m, TimeOffset: 15m) (ex in carb dialog units for TimeOffset is minutes and duration is hours, in Profile Switch units for TimeOffset is hours and duration is minutes...)
this is good if every info is in different units. then it's clear without label

MilosKozak:
👍

app/src/main/java/info/nightscout/androidaps/dialogs/CarbsDialog.kt
Philoul: Constants class is not available in database module
I propose two solutions using a new constructor...

1) test in constructor ("mmol" hardcoded in it)
```
    data class ValueWithUnit (val dValue: Double, val iValue: Int, val lValue: Long, val unit: Units) {
        constructor(dvalue:Double, unit:Units) : this(dvalue, 0, 0, unit)
        constructor(ivalue:Int, unit:Units) : this(0.0, ivalue, 0, unit)
        constructor(lvalue:Long, unit:Units) : this(0.0, 0, lvalue, unit)
        constructor(dvalue:Double, bgunit:String) : this(dvalue, 0, 0, if (bgunit=="mmol") Units.Mmol_L else Units.Mg_Dl)
    }
```

2) move MGDL and MMOL definition in Units class
```
        @SerializedName("mg/dl") Mg_Dl,
        @SerializedName("mmol") Mmol_L,
```
in Constants class
```
public class Constants {
    public static final String MGDL = Units.Mg_Dl.name();        // This is Nightscout's representation
    public static final String MMOL = Units.Mmol_L.name();
```

```
constructor(dvalue:Double, unit:String) : this(dvalue, 0, 0, Units.fromString(unit))
```
I prefer second one...

MilosKozak: agree

Philoul:
See result below (Careportal action translated, unit added, 
Screenshot_20210303-003640

And the little bug I don't know how to solve, after a scroll to the bottom then back to the top, some information are hidden (???), I must click again on upper UserEntry icon to show them again
Screenshot_20210303-004823

@MilosKozak What do you think of this proposal ?

Philoul:
My comment on this work:
- I tried to put an extension .toStringWithUnit for ValueWithUnit or UserEntry to have less code in TreatmentsUserEntryFragment, but I don't know how to @Inject in extension's file (I need dateUtil, resourceHelper... for string convertion).
- I think we should add rxBus (minimum for EventPreferenceChange), but I don't know how to convert `treatmentsPlugin.treatmentsFromHistory` for userEntryList in the line below to include an `updateGui()` function
```binding.recyclerview.swapAdapter(RecyclerViewAdapter(treatmentsPlugin.treatmentsFromHistory), false)```

- Some informations disappear when we scroll up and down (this little issue exists in current dev too)
- Translation and unit convertion works well

MilosKozak:
content is OK. just the view is somehow distracted now with more information.
maybe use different style (font, color etc) of date + type  and of detailed info

Philoul:
@MilosKozak I had a blue screen crash during a pull, my local and remote branch seems to be corrupted (impossible to checkout my local or remote branch, impossible to delete my local branch...)
=> I propose to close this branch (I can't work any more on it) and reopen a new one with a backup (I made a merge with latest dev branch)

PR #393, opened 2021-03-03
User entry valuewithunits

Philoul:
=> this branche replace #379

You can see examples of ValueWithUnit only in CalibrationDialog.kt and CarbDialog.kt
Currently there are some redundancies in UserEntryLogger.kt (but I don't know if we can improve it...)
=> A convertion from mmol/mg/dl is done in TreatmentsUserEntryFragment to always show data in selected units
=> Formated Strings (with parameters) are recorded as "Strings data with units" (can be shown in selected language in UserEntry tab (see examples in SmsCommunicatorPlugin)
=> No other convertion done currently (see opened questions below for time units)
=> Any number of ValueWithUnit (double, int and long accepted) can now be included in database

 All other call of uel.log have to be updated

- [x]  Code for migration of UserEntry table has to be done (I propose to remove existing table created from v1 to v2 and create a new one in v3 to v4)
- [x]  Increase version of Database (3->4)
- [x]  TreatmentsUserEntryfragment has to be updated with ValueWithUnits to show values with translated units (see question below in we include a label or not for each value)
- [x]  Improve layout (bold, color)
- [x]  Tested yet and works on previous dev (I must fix duration from NS long -> min)
- [x]  Add Careportal additional information (for example Bg value if BG Check, Time and duration if Activity...)
- [x]  Create text/csv export
- [x]  Add onResume and rxBus in Fragment


Philoul:
@MilosKozak See proposal for text format (I'm not a specialist of colors, and I found lots of different blue, green, red, orange..., I tried to propose something "understandable" but this could probably be improved, every action's color is defined in extension file `UserEntryExt.kt`)
- Each Action can have a dedicated color
- All `ValueWithUnit` are in defaulttext color and "Bold"
- Date of record and event notes (notes are stored in "s" database field) are in standard default color

Screenshot_20210304-145015

-> Not the purpose of this PR, but I think we will have to standardize and simplify the usage of different colors (ex Careportal orange is very close to Carbs orange, Extended Bolus blue is very close to basal color, temptarget green is very close to ClosedLoop green, icon for TempBasal is purple but tempbasal is light blue, blue of OpenLoop/iob... And all these definitions are splitted bettween App colors.xml file and Core colors.xml file...)

MilosKozak:
Much more readable now. 👍 
Too much colors is not necessary. I think it may be usefull to have different color for evey area (treatment, tt, profile, loop, careportal) but not to every single item

Philoul:
See screenshot with "ColorGroup" (the 8 proposed colors are in it)
Screenshot_20210306-174516

- I splitted Treatment in InsulinTreatment and CarbTreatment
- I changed the standard Orange color for CarePortal to a "High" Yellow (Orange was the same than Carbs color)
- Green of TT and Loop are different but close
- I selected the blue IOB for pump action (for ex disconnect, reconnect)

I kept for the moment the previous colorId attached to each action (but I can remove it if colorGroup Ok
=> I'm not a specialist of colors so everything can easily be change today (or later, no impacts on database) : 
- Color selection for each group in UserEntryExt.kt
- Group selected for each Action in UserEntry.kt

Philoul:
Careportal detailed information added (bg value, duration)
Screenshot_20210306-194337

core/src/main/java/info/nightscout/androidaps/utils/extensions/UserEntryExt.kt
```diff
@@ -1,224 +1,230 @@
...
-        UserEntry.Action.SMS_PROFILE                -> R.color.white
-        UserEntry.Action.SMS_PUMP_CONNECT           -> R.color.loopDisconnected
-        UserEntry.Action.SMS_PUMP_DISCONNECT        -> R.color.loopDisconnected
-        UserEntry.Action.SMS_SMS                    -> R.color.defaulttext
-        UserEntry.Action.SMS_TT                     -> R.color.tempTargetConfirmation
-        UserEntry.Action.TT_DELETED_FROM_NS         -> R.color.tempTargetConfirmation
-        UserEntry.Action.TT_FROM_NS                 -> R.color.tempTargetConfirmation
-        UserEntry.Action.TT_CANCELED_FROM_NS        -> R.color.tempTargetConfirmation
-        else                                        -> R.color.defaulttext
+        Action.EXTENDED_CARBS             -> R.color.carbs
```
MilosKozak, line 110: Shouldn't it be based on `ColorGroup`?

core/src/main/java/info/nightscout/androidaps/utils/extensions/UserEntryExt.kt
```diff
@@ -1,224 +1,230 @@
...
+        Action.TEMP_BASAL                 -> R.color.basal
```
MilosKozak, line 111: feel free to create new constant R.color for every group

core/src/main/java/info/nightscout/androidaps/utils/extensions/UserEntryExt.kt
Philoul: With ColorGroup, Action.colorId is not used anymore. I can remove it.

core/src/main/java/info/nightscout/androidaps/utils/extensions/UserEntryExt.kt
Philoul: I didn't want to create new colors, I think we should try to limit and standardize the existing colors to make it easier to understand (ex Orange Carb, Blue Insulin...) and use the same colors everywhere for the same theme (Text, icons, graphs...).
But for "multicolor items" (ex Careportal, Loop, BG values), or when there is no standard color (Pump) we could propose new ones or use the most frequently use (if not already taken by another theme) as "theme color".
When 2 themes have colors that are too close (eg TT, BG In Range and Closed Loop), I don't know which is the best solution ...

osodebailar: If you Use Colors you Can have a look into my dark/light Branch.
To will be hopefullly merged in the next times.
I began to refactor colors and the only access to the colors succed by Attributes.i began to rename Colors to gray, Light gray ... and so on

Philoul: Good idea to work on colors, and reduce / harmonize the diversity.
I think it's better if we could "name" colors according to standard use in AAPS (I prefer "Light Carb" and "Dark Carb" than neutrals "Light Orange" / "Dark Orange")
In this PR, I select colorId with R.color.xxxx (I don't know how it works with Materal, I suppose we can replace Id by an attribute in UserEntryExt.kt and replace resourceHelper.gc in TreatmentsUserEntryFragment...)
I'll take a look in your PR.

osodebailar: The Standard use will be clear by the Attributes and Not by the colorname in the Future .
In my branch resourceHelper.gc will Not be used anymore.
Instead we use the new function resourceHelper.getColorbyAttributes

Philoul: For UserEntries, Milos asked me to create "groups of Color" according to actions. I initialize 8 groups.
Do you think we could map these groups (we can of course update this list) with your attributes?

osodebailar: I think yes.
But the access function is only available in my Branch in the Moment .
Its nö Time to merge to avoid additional work.

Philoul: Don't bother, I'll take a look on my side too with your PR, to see the impact (and prepare modifications if your PR is merged before mine).

osodebailar: Perfect. And don't Hesitate for question of help . Still Write me a Mail and i help wherever i Can.

MilosKozak:
except export is this done?

Philoul:
I worked on a parallel branch at the same time of this one to be able to finish "tuning" here...
csv export works now (several trials before having a file that works with Excel...), but it's probably not the best architecture (csv is build in ClassicPrefsFormat.kt, and the confirmation function is dedicated to pref...).

Philoul:
Latest dev merged (but I had to reset database on my dev phone to avoid crash because database version increased to 4 in this PR too...)
Include csv exports (tested)

app/src/main/java/info/nightscout/androidaps/dialogs/CarbsDialog.kt
```diff
@@ -270,10 +271,13 @@ class CarbsDialog : DialogFragmentWithDate() {
                     }
                     if (carbsAfterConstraints > 0) {
                         if (duration == 0) {
-                            uel.log("CARBS", d1 = carbsAfterConstraints.toDouble(), i1 = timeOffset)
+                            if (eventTimeChanged)
```
MilosKozak, line 274: would not be helpfull to do
`ValueWithUnit(condition = eventTimeChanged, eventTime, Units.Timestamp)` and add only if first param is true?
it's on so many places .....

MilosKozak:
is it ready?

Philoul:
Just one last question concerning durations.
=> I saw that in database we have some durations in "ms" with long values (in uel.log, I convert these durations to min inside `ValueWithUnit(xx/60000, Units.M`)
=> We can keep it as is, but if you prefer, I can manage another unit (ms) with long values to keep original information in userEntry table (and only make convertion ms-> min in fragment, to be decided in csv export)
- If we keep as is it's ready
- If you prefer store in UE Table duration in ms with long values, I have to add one more commit...

Philoul:
I think we can probably reduce a bit more the number of "Actions" before merge (for ex remove TT_ACTIVITY, TT_HYPO and TT_EATING_SOON, and give information with ValueWithUnit(tempTarget.reason.text, Units.TT_Reason), we can simplify the 9 SUSPEND_xxH/Disconnect_xxH with one SUSPEND or DISCONNECT Action and duration as ValueWithUnits)

MilosKozak:
> I think we can probably reduce a bit more the number of "Actions" before merge (for ex remove TT_ACTIVITY, TT_HYPO and TT_EATING_SOON, and give information with ValueWithUnit(tempTarget.reason.text, Units.TT_Reason), we can simplify the 9 SUSPEND_xxH/Disconnect_xxH with one SUSPEND or DISCONNECT Action and add a duration with ValueWithUnits)

👍

Philoul:
Number of actions is reduced a little bit (I removed Actions when the equivalent TherapyEvents existed)
=> This PR is now Ok for me

Note : I kept for the moment in Convertor `fun toAction(action: String?) = action?.let { Action.fromString(it) }` which is more "tolerant" than `Action.valueOf(it)` (if an action is removed during dev, it's not necessary to increase database version and it will just be replaced by "Unkown" (I propose to keep it as long as we are working on migration to room)

MilosKozak:
Test is failing

Philoul:
I think test failed because of addition in the constructor of ClassicPrefsFormat.kt of DateUtil, Translator and ProfileFunction needed for csv export
=> I tried to add them in ClassicPrefsFormatTest (the same as resourceHelper) but it still doesn't work

```
java.lang.RuntimeException: Invoking the beforeTestMethod method on PowerMock test listener org.powermock.api.extension.listener.AnnotationEnabler@17e7df9e failed.
	at org.powermock.tests.utils.impl.PowerMockTestNotifierImpl.notifyBeforeTestMethod(PowerMockTestNotifierImpl.java:84)
	at org.powermock.modules.junit4.internal.impl.PowerMockJUnit44RunnerDelegateImpl$PowerMockJUnit44MethodRunner.executeTest(PowerMockJUnit44RunnerDelegateImpl.java:308)
	at org.powermock.modules.junit4.internal.impl.PowerMockJUnit47RunnerDelegateImpl$PowerMockJUnit47MethodRunner.executeTestInSuper(PowerMockJUnit47RunnerDelegateImpl.java:131)
	at org.powermock.modules.junit4.internal.impl.PowerMockJUnit47RunnerDelegateImpl$PowerMockJUnit47MethodRunner.access$100(PowerMockJUnit47RunnerDelegateImpl.java:59)
	...
```
Unfortunatly I don't know how to fix it (except create a dedicated file for csv export and keep ClassicPrefsFormat as it is in dev)

PR #403, opened 2021-03-07
Bug354 mdi problem

andyrozman:


MilosKozak:
resolved in dev2 by refactoring to kotlin and processed nullability correctly

Issue #417, opened 2021-03-14
closed issue #382 : wear bolusing still not possible


thomasmenges:
Please see issue #382

Philoul:
Fixed by #445

PR #445, opened 2021-03-27
load Action String Handler

AdrianLxM:


Philoul:
@Adrian 👍 
#417 fixed!

PR #458, opened 2021-04-03
callback if carbs are added to db but not pump

AdrianLxM:


Philoul:
Hi Adrian
In `TreatmentDialog.kt` (line 176), if I enter in Dialog Bolus + Carbs, I now have a double entry in uel.log (I think it's probably a consequence of your modification...):
```
      commandQueue.bolus(detailedBolusInfo, object : Callback() {
          override fun run() {
              if (!result.success) {
                  ErrorHelperActivity.runAlarm(ctx, result.comment, resourceHelper.gs(R.string.treatmentdeliveryerror), info.nightscout.androidaps.dana.R.raw.boluserror)
              } else
                  uel.log(action, Sources.TreatmentDialog,
                      ValueWithUnit.Insulin(insulinAfterConstraints),
                      ValueWithUnit.Gram(carbsAfterConstraints).takeIf { carbsAfterConstraints != 0 })
          }
      })
```
To avoid that I can move uel.log before or after `commandQueue.bolus` (it is currently outside in `BolusWizard.kt`)

PR #459, opened 2021-04-05
Secondary Graph Sensitivity Scale (100) and Menu

Philoul:
I propose this modification to improve menu and secondary graphs:
- Move BGI curve in menu just below Deviation
- Use the same color in menu that Deviation (proposed colors can of course be changed, espacially when #326 will be merged)
=> Now curves with the same color in menu have a common scale and unit so they can be compared when on the same graph
- Change Graph scale for sensitivity curve (from -1/1 to -100/100) we can directly compare values in secondary graph with sensitivity percentage in Overview

I think Color in menu could suggest the same scale if both curves are on the same graph...
Screenshot_20210405-103915


MilosKozak:
i belive that having 100% at 0 is more clear. you see at first look if you are more resistent or sensitive.
In your proposal you'd be able read only direction as there is no 100% level marked

Philoul:
Scale of Sens modified to have 100% at 0
Note mini scale proposed is 95 - 105 (even if Sens equals 0 for all period), and 80% of maxY if with other curves.
See result in screenshot below (Sens alone in a graph or included with other curves)
Screenshot_20210414-233415_AndroidAPS

PR #468, opened 2021-04-10
Fix double Entries in Treatment Dialog (+ add new uel.log for plugin enabled/disabled

Philoul:
With VirtualPump (not tested with real pump), if I enter Carbs + Insulin in TreatmentDialog, I get 2 userEntries (uel.log executed twice)
@MilosKozak I saw you put some uel.log inside `commandQueue` (TreatmentDialog and InsulinDialog)
=> I suggest to always move uel.log before `commandQueue` (sometimes if a problem occurs during callback, we can miss an uel.log or get redundant entries...)

I added an uel.log for new plugin selection

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/dialogs/TreatmentDialog.kt
+++ b/app/src/main/java/info/nightscout/androidaps/dialogs/TreatmentDialog.kt
@@ -167,18 +167,18 @@ class TreatmentDialog : DialogFragmentWithDate() {
                                     { aapsLogger.error(LTag.DATABASE, "Error while saving carbs", it) }
                                 )
                     } else {
-                        if (detailedBolusInfo.insulin > 0)
+                        if (detailedBolusInfo.insulin > 0) {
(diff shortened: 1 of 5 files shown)


Philoul:
I think most of uel.log are now created in App
=> Just missing (I think) some uel.log inside the pump drivers to identify the manual actions that are done in the pump and are imported in aaps (Bolus, multiwave bolus, TBR, Site/Reservoir Change, Prime)
I started to take a look in Insight plugin to see if I can try to add some of them, but it's tricky...

Philoul:
I found an important bug found this evening: I have to replace or change "ValueWithUnit.StringResource" (because id can change between 2 builds or 2 releases)... Not a good idea to save id in db!
I'll probably propose an update with `ValueWithUnit.StringResource(R.string.xxxx)` replaced by `ValueWithUnit.SimpleString(resourceHelper.gs(R.string.xxxx)` (already modified on a parallel branch), and I'll probably remove most of StringResource with param... (to keep only param) because most of them was only friendly sentences for fragment

PR #470, opened 2021-04-11
Combo PumpSync update

jotomo:
- Mostly removing unneeded code (after ensuring in `deliverTreatment` the request is sound). 
- Updated docs, added explanation on lack of TBR IDs
- Untangled date vs ID, now that they're different fields, yay!

Test protocol:

- [x] User-initiated bolus
- [x] User-initiated bolus with carbs
- [x] User-initiated carbs
- [x] Cancelling bolus during pump delivery  
- [x] User requested TBR
- [x] User requested TBR cancellation
- [x] TBR requested by OpenAPS
- [x] SMB requested by OpenAPS
- [x] Ending TBR when user cancelled TBR on pump
- [x] Starting TBR when user started a TBR on the pump
- [x] TBR end and start record when user changed TBR on pump
- [x] Adding bolus issued on pump
- [x] (User) suspended pump creates 15m 0% TBR
- [x] Nightscout uploads still work (bolus, carbs, TBR, announcements)

### DIFF
--- a/combo/src/main/java/info/nightscout/androidaps/plugins/pump/combo/ComboPlugin.java
+++ b/combo/src/main/java/info/nightscout/androidaps/plugins/pump/combo/ComboPlugin.java
@@ -96,7 +96,7 @@ public class ComboPlugin extends PumpPluginBase implements PumpInterface, Constr
      */
     private volatile boolean scripterIsBolusing;
     /**
-     * This is set to true to request a bolus cancellation. {@link #deliverBolus(DetailedBolusInfo)}
+     * This is set to true to request a bolus cancellation. {@link #deliverTreatment(DetailedBolusInfo)} (DetailedBolusInfo)}
      * will reset this flag.
      */
     private volatile boolean cancelBolus;
@@ -105,7 +105,7 @@ public class ComboPlugin extends PumpPluginBase implements PumpInterface, Constr
      * This is set (in {@link #checkHistory()} whenever a connection to the pump is made and
      * indicates if new history records on the pump have been found. This effectively blocks
      * high temps ({@link #setTempBasalPercent(Integer, Integer, PumpSync.TemporaryBasalType)} and boluses
-     * ({@link #deliverBolus(DetailedBolusInfo)} till the queue is empty and the connection
+     * ({@link #deliverTreatment(DetailedBolusInfo)} till the queue is empty and the connection
      * is shut down.
      * {@link #initializePump()} resets this since on startup the history is allowed to have
      * changed (and the user can't possible have already calculated anything with out of date IOB).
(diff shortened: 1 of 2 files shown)


combo/src/main/java/info/nightscout/androidaps/plugins/pump/combo/ComboPlugin.java
```diff
@@ -1113,6 +1067,8 @@ private void checkForUnsafeUsage(CommandResult commandResult) {
     private void checkAndResolveTbrMismatch(PumpState state) {
```
Philoul, line 1067: Hi @jotomo 
How do you manage TBR done manually on the pump in recent past (for exemple started one hour ago) ?
I don't see in this function any `pumpSync.syncTemporaryBasalWithPumpId` with `state.timestamp` to update database (all starts `now`... )
I misunderstood something ?

jotomo: You're reading that right.
Unlike bolus handling, TBR-handling doesn't read the pump history. On the pump, TBR records are only created after a TBR has finished. So when a TBR is started on the pump, it can't be known when it started until the TBR ends or is cancelled. Cancelling would assume a user works against the loop, and creating a temporary TBR (AAPS-side) and updating it once a record exists on the pump has other problems, since there's no ID for TBRs, only timestamps.
For the regular uses where the user doesn't set TBR (or rather infrequently for some special cases), TBRs are only changed on the pump for error conditions where the pump is stopped, those need to be synced.
The approach taken is to create a TBR record in AAPS when AAPS requests a TBR and forego the TBR history on the pump entirely. The pump state is refreshed often enough to tolerate not seeing the full length of a TBR on the pump if it was changed. Thus, during a pump refresh a new TBR starting _now_ is created in AAPS if a mismatch between expected state and actual pump state is detected (Method `checkAndResolveTbrMismatch`).
This approach skipped implementing edge-cases that pose no real risk, in part due to limited resources to implement every edge-case scenario. Insulin amount given via boluses are significantly higher, so the priority was there to make that as safe as possible.

Philoul: I had in mind "finished TBR" (use case for example is Pump remains out of range for a long time, phone forgot at home, and user make several manual entries in pump, TBR, simple Bolus, and why not square/multiwave boluses even if he shouldn't do that with AAPS...)
I'm currently working on UserEntry informations, my idea was to record informations concerning all actions done manually into the pump by the user, and that could impact the loop (Boluses, TBR, Site/reservoir change if available, unsafe alerts...)
I initialized 2 uel.log (in blind coding because I don't have Combo), one for boluses from history and the second for an "unsafe alert" as an example. I'll propose a "WIP PR" this evening to open discussion if we go further in this direction in all pump drivers...
(You can see results of uel.log in Treatment Tab latest icon on the right, dev2_dana_combo_only branch).

PR #472, opened 2021-04-13
Meallink deliver treatment change

jotomo:
Starting points for other pump drivers ...

### DIFF
--- a/insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
+++ b/insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
@@ -578,6 +578,9 @@ public class LocalInsightPlugin extends PumpPluginBase implements PumpInterface,
 
     @NonNull @Override
     public PumpEnactResult deliverTreatment(DetailedBolusInfo detailedBolusInfo) {
+        if (detailedBolusInfo.insulin == 0 || detailedBolusInfo.carbs > 0) {
+            throw new IllegalArgumentException(detailedBolusInfo.toString());
+        }
         PumpEnactResult result = new PumpEnactResult(getInjector());
         double insulin = Math.round(detailedBolusInfo.insulin / 0.01) * 0.01;
         if (insulin > 0) {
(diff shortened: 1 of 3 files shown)


insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -578,6 +578,9 @@ public int getBatteryLevel() {
 
     @NonNull @Override
     public PumpEnactResult deliverTreatment(DetailedBolusInfo detailedBolusInfo) {
+        if (detailedBolusInfo.insulin == 0 || detailedBolusInfo.carbs > 0) {
+            throw new IllegalArgumentException(detailedBolusInfo.toString());
```
Philoul, line 582: Hi,
During my tests on Insight Driver, I saw that `throw new IllegalArgumentException(detailedBolusInfo.toString());` occurs about 20 times per day (and each time AAPS restart...)
I added a log and I get
`Illegal argument : Insulin: 0.0 Carbs: 0.0` (Insight doesn't manage carbs and when Insulin equals 0 nothing is done in driver...)
=> I commented this line to avoid too many restart....

jotomo: The contract for `Pumpinterface.deliverTreatment` now specifies the the caller must ensure no requests are made to the pump which the pump cannot satisfy. The statement verifies that contract to pin down where the issue is. In this case the caller is not adhering to the contract. If you change the line to `throw new IllegalArgumentException(detailedBolusInfo.toString(), new Exception())` you'll get a stacktrace that shows who is making the call.
tl;dr: this is an error that needs to be fixed. Removing the line will likely just cause issues further down where they're harder to trace back.

Philoul: Ok, thanks for the explain

Philoul: See Stacktrace below    
```
java.lang.IllegalArgumentException: {"bolusType":"SMB","carbTime":0,"carbs":0.0,"carbsDuration":0,"deliverAtTheLatest":1619475827673,"eventType":"CORRECTION_BOLUS","insulin":0.0,"lastKnownBolusTime":1619286422354,"timestamp":1619475832231}
        at info.nightscout.androidaps.plugins.pump.insight.LocalInsightPlugin.deliverTreatment(LocalInsightPlugin.java:567)
        at info.nightscout.androidaps.queue.commands.CommandSMBBolus.execute(CommandSMBBolus.kt:30)
        at info.nightscout.androidaps.queue.QueueThread.run(QueueThread.kt:118)
```

Philoul:
I noticed that with branch dev2_dana_combo_only (with virtual pump) or with my proposal for Insight Driver #488, when I enter carbs with Calculator or Carbs dialog, it can take several minutes (sometimes more than 4 min) before having the COB updated in overview... 
I don't know what's the reason and if we can improve it, but I think it's really too long (in master, overview is immediatly updated after dialog is closed...)
Note: The first time I tested this branch on my dev phone, I entered 3 times carbs because I thought I didn't entered correctly the value or it was not taken into account!

jotomo:
> I noticed that with branch dev2_dana_combo_only (with virtual pump) or with my proposal for Insight Driver #488, when I enter carbs with Calculator or Carbs dialog, it can take several minutes (sometimes more than 4 min) before having the COB updated in overview...

@Philoul Not seeing the issue here (with virtual pump or Combo). If this persists, please make Milos aware of it, so it doesn't get lost in this issue here.

PR #482, opened 2021-04-16
[WIP] uel.log for Combo Pump

Philoul:
I propose to add uel.log for all actions that the user has performed manually in the pump and which may have an impact on the loop (mainly Boluses/TBRs from the pump history, and if available Site/Reservoir changes, unsafe usage notifications, ...)
- Blind coding for this PR (I don't have a Combo to check if it's OK or not) so review required (@jotomo ?) for Bolus and Basal Temp entries from pump history
- Open question: I have included an uel.log for one unsafe usage as an example, should I remove or generalized?

### DIFF
--- a/danar/src/main/java/info/nightscout/androidaps/danaRv2/comm/MsgHistoryEvents_v2.kt
+++ b/danar/src/main/java/info/nightscout/androidaps/danaRv2/comm/MsgHistoryEvents_v2.kt
@@ -7,9 +7,6 @@ import info.nightscout.androidaps.events.EventPumpStatusChanged
 import info.nightscout.androidaps.logging.LTag
 import info.nightscout.androidaps.plugins.pump.common.defs.PumpType
 import info.nightscout.androidaps.utils.T
-import info.nightscout.androidaps.utils.userEntry.UserEntryMapper.Action
-import info.nightscout.androidaps.utils.userEntry.UserEntryMapper.Sources
-import info.nightscout.androidaps.utils.userEntry.ValueWithUnitMapper
 import java.util.*
 
 class MsgHistoryEvents_v2 constructor(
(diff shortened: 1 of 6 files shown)


Philoul:
@MilosKozak see latest commit for DanaRV2, if you are ok with this proposal I'll continue with other pump
... and if ok can you review code, I added uel.log for DanaRv2 pump in blind coding!

danar/src/main/java/info/nightscout/androidaps/danaRv2/comm/MsgHistoryEvents_v2.kt
```diff
@@ -81,17 +83,26 @@ class MsgHistoryEvents_v2 constructor(
                     pumpId = datetime,
                     pumpType = PumpType.DANA_RV2,
                     pumpSerial = danaPump.serialNumber)
+                aapsLogger.debug(LTag.PUMPBTCOMM, (if (newRecord) "**NEW** " else "") + "EVENT TEMPSTART (" + recordCode + ") " + dateUtil.dateAndTimeString(datetime) + " (" + datetime + ")" + " Ratio: " + param1 + "% Duration: " + param2 + "min")
                 status = "TEMPSTART " + dateUtil.timeString(datetime)
+                if (newRecord)
```
MilosKozak, line 88: These events are not user events. these are result of AAPS work and should not be looged as "manual" actions
I didn't check other places but we should only log actions when user click something

Philoul: Here I thought it was manual entries that the user entered directly on pump (and imported in AAPS from pump history), that's why I proposed to add them.

For all other uel.log, most of them are "click" done by user in AAPS.

The exceptions are (and I can remove all or some of them if you want):
- in LoopPlugin (uel.log for SMB and TBR) (these uel.log are recorded but hidden by default in UserEntryFragment)
- in Automation Plugin (uel.log for TBR, PS, Loop Enabled/Disabled done by an automation rule, indirect actions done by the user but not with a click)
- in NSClient Plugin (uel.log for new entries done from NsClient Plugin, equivalent of uel.log of SMS Plugin but for NSClient entries)
- Dexcom Plugin (Sensor change or Calibration done by user in Dexcom Patched app)

I thought that in case of investigation, all these entries could be usefull in the csv file...

MilosKozak: the rest seems to be ok except NsClientAddUpdateWorker which will generate a lot of events until we finish migration to API
so, for now remove only these Dana* history
we can review the rest when we finish all changes

jotomo: Gotta say I don't really like the idea of having `UserEntry` code at the driver level.
For boluses and TBRs, it could be reported back whether a bolus was "found" on the pump, or read after AAPS requested it, so that `PumpSync` (or the service calling it) could do such logging, or record the origin of it.
Entirely related to that: the Combo produces notifications and dispatches them via `rxBus`. Which makes me wonder whether including methods in `PumpSync` to report events back to AAPS might be a good idea.
That way the driver could focus on driving the pump and wouldn't need to access `uel`, `rxBus` or `Notification` ...

Philoul:
Ok, so I propose to close this proposal, we will see later if we record these manual entries done by the user directly on pump and the best way to do that...

PR #488, opened 2021-04-20
[WIP] Driver Insight with Database

Philoul:
Test done with Insight v2
Action from LoopPlugin:
- [x] SMB
- [x] Set Ext Bolus
- [x] Cancel Ext Bolus
- [x] Set TBR
- [x] Cancel TBR

Manual Action from AAPS
- [x] Bolus
- [x] Bolus + Cancel during injection
- [x] Extended Bolus (TBR emulation disabled)
- [x] Cancel Extended Bolus
- [x] Set TBR (when no TBR running)
- [x] Cancel TBR

Manual Action in Pump (result in AAPS after pump synchro)
- [x] Bolus
- [x] Bolus + cancel on Pump
- [x] Bolus Extended (running)
- [x] Bolus Extended canceled in pump
- [x] Bolus Multiwave (running) (Ok in AAPS for immediate bolus and extended bolus)
- [x] Bolus Multiwave canceled in pump
- [x] TBR set in pump
- [x] TBR Canceled in pump

TherapyEvent
- [x] Pump Battery change
- [x] Insulin Change
- [x] Site Change

Pump history Stop and start event
- [x] set zero temp during pump is stopped
- [x] sync history if fake TBR (on start or Extended bolus done on pump)

Add Insight database
- [ ] Boluses
- [ ] TBR (start and stop)
- [ ] Pump (Start and stop)
- [ ] Simplify code (closer to original Insight Driver)


MilosKozak:
NSUpload and uploadQueue will be removed. no need to take care about

Philoul:
To illustrate see screenshots with my loop phone on the left (without room, dev-e) and my dev phone on the right with new database... (only BG value are identical, I made several tests on dev phone so not exactly same COB, IOB, sensitivity...)
Insight v3 on loop, Insight v2 on dev
Screenshot_20210420-220213_AndroidAPS

Philoul:
After lots of work to understand how driver works and, I think I now have something that is ready to test wider (everything seems to work now with my dev phone)...
Combo plugin helps me too...

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1281,130 +1283,194 @@ private void processTotalDailyDoseEvent(TotalDailyDoseEvent event) {
...
+                    timestamp - 1500L,
+                    timestamp - 500L);
+            temporaryBasals.add(temporaryBasal);
+    }
+
+    /**
+     * Checks the main screen to determine if TBR on pump matches app state.
+     */
+    private void checkAndResolveTbrMismatch(String serial) throws Exception {
+        // compare with: info.nightscout.androidaps.plugins.PumpDanaR.comm.MsgStatusTempBasal.updateTempBasalInDB()
```
DigitalDan1, line 1404: This comment says PumpDanaR

Philoul: Yes, I didn't change comment, this function checks and update AAPS if there are inconsistances between TBR in database and TBR running in pump after history synchro.
I copied this function from Combo Driver (himself copied from danaR...).
I kept this function for security but I'm not sure it is still necessary now (with my latest commit, I didn't see any new call of it).

jotomo: Initially the combo copied the logic the DanaR pump used to handle that case. The method and logic has since been removed. I just removed the comment from the combo code, the same can be done here. @DigitalDan1 thanks for pointing this out.

Philoul:
Since https://github.com/nightscout/AndroidAPS/commit/37e3c4532a5156927552015ba5cf2e7217563ef9 I cannot run anymore my branch to test Insight Driver because I can't set a profile... (I work on another branch without this commit)
Even after an "Activate Profile" I still have NO PROFILE SET in main screen 
@MilosKozak is it a bug, the work with profile integration in room that is still in progress or should I modify something in Insight driver to make it work again ?
Screenshot_20210502-184315
Screenshot_20210502-184630

=> I included a lot of `aapsLogger.debug` to test and check if everything is consistent between aaps and pump, I continue my tests before remove them...

MilosKozak:
set different profile like 110%

Philoul:
Thanks
Before I clean the code and remove all my aapsLogger.debug, it would be nice if other developpers could test this PR (I think it works fine now, but I might have missed something...).

MilosKozak:
Insight specific database tables should be moved to Insight module too
See: https://github.com/nightscout/AndroidAPS/tree/dev2_dana_combo_only/dana/src/main/java/info/nightscout/androidaps/dana/database

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1261,150 +1278,240 @@ private void processDateTimeChangedEvent(DateTimeChangedEvent event) {
...
+                            duration,
+                            false,
+                            PumpSync.TemporaryBasalType.NORMAL,
+                            lastStopEvent,
+                            PumpType.ACCU_CHEK_INSIGHT,
+                            serial
+                    );
+                lastStartEvent = 0;
+                aapsLogger.debug(LTag.PUMP, "XXXX event STOP: " + lastStopEvent + " HMS: " + dateUtil.dateAndTimeAndSecondsString(lastStopEvent) + " ZeroTemp Duration (min): " + (duration)/60000);
+                if (sp.getBoolean(R.string.key_insight_log_operating_mode_changes, false))
```
TebbeUbben, line 1399: Why do you save a TBR here? You can only determine the duration once the pump is started again by searching for the previous stop event. Also, you should not save a fake TBR if the previous operating mode change was a pause event. If the pump is started again after having been in pause mode (for a maximum of 15 minutes), all the missing insulin from that time frame is delivered.

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1261,150 +1278,240 @@ private void processDateTimeChangedEvent(DateTimeChangedEvent event) {
...
-        temporaryBasal.isAbsolute = false;
-        temporaryBasal.date = timestamp;
+        aapsLogger.debug(LTag.PUMP, "XXXX event StartTbr history rec Timestamp: " + timestamp + " date: " + event.getEventYear() + "/" + event.getEventMonth() + "/" + event.getEventDay() + " " + event.getEventHour() + ":" + event.getEventMinute() + ":" + event.getEventSecond() + " Rate: " + event.getAmount() + " Duration(min): " + event.getDuration());
+        TemporaryBasal temporaryBasal = new TemporaryBasal(
+                timestamp,
+                T.mins(event.getDuration()).msecs(),
+                event.getAmount(),
+                false,
+                PumpSync.TemporaryBasalType.NORMAL,
+                timestamp,
```
TebbeUbben, line 1420: You can use `event.getEventPosition()` as an ID here.

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1261,150 +1278,240 @@ private void processDateTimeChangedEvent(DateTimeChangedEvent event) {
...
-        temporaryBasal.date = timestamp - 1500L;
-        temporaryBasals.add(temporaryBasal);
+        aapsLogger.debug(LTag.PUMP, "XXXX event StopTbr history rec Timestamp: " + timestamp + " date: " + event.getEventYear() + "/" + event.getEventMonth() + "/" + event.getEventDay() + " " + event.getEventHour() + ":" + event.getEventMinute() + ":" + event.getEventSecond());
+            PumpSync.PumpState.TemporaryBasal temporaryBasal = new PumpSync.PumpState.TemporaryBasal(
+                    timestamp - 1500L,
+                    0L,
+                    100.0,
+                    false,
+                    PumpSync.TemporaryBasalType.NORMAL,
+                    timestamp - 1500L,
```
TebbeUbben, line 1435: You can use `event.getEventPosition()` as an ID here.

TebbeUbben (changes requested):
One major issue that I see with the current PumpSync interface is that it undermines how the Insight work.

Boluses essentially have three IDs:

- A 16 bit Bolus ID (that can repeat in rare cases if the pump is used long enough and enough boluses are delivered)
- A 32 bit ID for the history start event (BolusProgrammed)
- A 32 bit ID for the history end event (BolusDelivered)

It is advisable to store the event IDs and only use the Bolus ID (that is included in each history event) to identify them. If the Insight offers that amount of precision, we should use it.

Also, TBRs have two IDs:

- A 32 bit ID for the history start event (StartOfTBR)
- A 32 bit ID for the history end event (EndOfTBR)

Currently, you are already saving a TBR to the database when the command is issued. I purposefully did not do that, so that I get stable IDs when reading the history. The same applies to fake TBRs for time frames in which the pump was stopped.

You should also be careful about the order in which you process history events, e.g. you should not override a BolusDeliveredEvent with a BolusProgrammedEvent as the former contains the final amount (in case of a canceled bolus).

Philoul:
@TebbeUbben thank you very much for all your comments and explanations ...
I recently started working on another version which is very close to your original driver (I was not happy with the code and the difference now is that I understand why ...).
Would you rather I share my other version as a starting point for you (this is your pilot ... As I told you, it was just a way for me to better understand how the pilot works and the database), or that I continue to work on it with your help?

MilosKozak:
For now i'd do only necessary changes to make it working

Philoul:
Replaced by #502

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -811,19 +812,8 @@ public PumpEnactResult setExtendedBolusOnly(Double insulin, Integer durationInMi
...
-            insightBolusID.pumpSerial = connectionService.getPumpSystemIdentification().getSerialNumber();
-            databaseHelper.createOrUpdate(insightBolusID);
-            ExtendedBolus extendedBolus = new ExtendedBolus(getInjector());
-            extendedBolus.date = insightBolusID.timestamp;
-            extendedBolus.source = Source.PUMP;
-            extendedBolus.durationInMinutes = durationInMinutes;
-            extendedBolus.insulin = insulin;
-            extendedBolus.pumpId = insightBolusID.id;
-            treatmentsPlugin.addToHistoryExtendedBolus(extendedBolus);
+            Integer bolusId = connectionService.requestMessage(bolusMessage).await().getBolusId();
```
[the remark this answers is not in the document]
Philoul: Yes it was

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
Philoul: My idea here was to immediatly start zero TBR only on Stop event (not for pause event), even if Start event is not in History (duration now + 15 min so duration updated on each readHistory).
=> removed on #502

PR #489, opened 2021-04-22
Add search device before Bluetooth connection

bubbledevteam:
Add search device before Bluetooth connection


rileylink/src/main/java/info/nightscout/androidaps/plugins/pump/common/hw/rileylink/ble/RileyLinkBLE.java
```diff
@@ -593,4 +610,132 @@ private String getGattStatusMessage(final int status) {
...
+        }
+    };
+
+    public void stopScan() {
+        handler.removeMessages(TIME_OUT_WHAT);
+        if (bluetoothAdapter == null) {
+            return;
+        }
+        try {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
AdrianLxM, line 714: minSDK is 26 -> is there a reason for this distiction?

rileylink/src/main/java/info/nightscout/androidaps/plugins/pump/common/hw/rileylink/ble/RileyLinkBLE.java
```diff
@@ -593,4 +610,132 @@ private String getGattStatusMessage(final int status) {
...
+
+    private List<ScanFilter> buildScanFilters() {
+        ArrayList scanFilterList = new ArrayList<>();
+        ScanFilter.Builder scanFilterBuilder = new ScanFilter.Builder();
+        scanFilterBuilder.setDeviceAddress(macAddress);
+        scanFilterList.add(scanFilterBuilder.build());
+        return scanFilterList;
+    }
+
+    @TargetApi(Build.VERSION_CODES.M)
```
AdrianLxM, line 623: minSDK is 26. Is this really needed?

rileylink/src/main/java/info/nightscout/androidaps/plugins/pump/common/hw/rileylink/ble/RileyLinkBLE.java
```diff
@@ -593,4 +610,132 @@ private String getGattStatusMessage(final int status) {
...
+        @Override
+        public void onScanFailed(int errorCode) {
+            super.onScanFailed(errorCode);
+            stopScan();
+        }
+    };
+    private BluetoothAdapter.LeScanCallback mLeScanCallback = new BluetoothAdapter.LeScanCallback() {
+        public void onLeScan(final BluetoothDevice device, final int rssi,
+                             final byte[] scanRecord) {
+            Platform.get().postDelayedMain(new Runnable() {
```
AdrianLxM, line 682: Is there a reason to run this on the UI thread?

rileylink/src/main/java/info/nightscout/androidaps/plugins/pump/common/hw/rileylink/Platform.java
```diff
@@ -0,0 +1,66 @@
...
+        return Executors.newCachedThreadPool();
+    }
+
+    public void postDelayedMain(Runnable r, long t) {
+        mainhandler.postDelayed(r, t);
+    }
+
+    public void removeCallbacks(Runnable r) {
+        mainhandler.removeCallbacks(r);
+    }
```
AdrianLxM, line 40: This seems to be unused code?

rileylink/src/main/java/info/nightscout/androidaps/plugins/pump/common/hw/rileylink/Platform.java
```diff
@@ -0,0 +1,66 @@
...
+    private static Platform findPlatform() {
+        try {
+            Class.forName("android.os.Build");
+            if (Build.VERSION.SDK_INT != 0) {
+                return new Android();
+            }
+        } catch (ClassNotFoundException ignored) {
+        }
+        return new Platform();
+    }
```
AdrianLxM, line 28: This seems to be code to detect the platform while we know we are running on Android?

rileylink/src/main/java/info/nightscout/androidaps/plugins/pump/common/hw/rileylink/ble/RileyLinkBLE.java
```diff
@@ -593,4 +610,132 @@ private String getGattStatusMessage(final int status) {
...
+                if (bluetoothLeScanner == null) {
+                    bluetoothAdapter.startLeScan(mLeScanCallback);
+                    return;
+                }
+                bluetoothLeScanner.startScan(buildScanFilters(), buildScanSettings(), scanCallback);
+            } else {
+                bluetoothAdapter.startLeScan(mLeScanCallback);
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
AdrianLxM, line 648: Please use the logger. Especially for such general exceptions.

AdrianLxM (changes requested):
Thanks for the PR.
I'm not sure if `master` is the correct target or if `dev` would be better.

I do have a few questions I added as inline comments.
Also I wonder if this isn't code duplication with the BLE code in `info.nightscout.androidaps.plugins.pump.common.dialog.RileyLinkBLEConfigActivity` and could be unified?

If it uses scanning all the time - even if not needed - wouldn't the connection then stop as soon as people switch off location on the phone?

andyrozman:
This functionality is totally unnecessary... Once you have RL configured there should be NO MORE search done... We do search in configuration and that is only place this should be done...

bubbledevteam:
> This functionality is totally unnecessary... Once you have RL configured there should be NO MORE search done... We do search in configuration and that is only place this should be done...

We have tested and found that some Bluetooth devices must be searched before they are connected or they cannot be connected

StephenBrown2:
Would this fix the issue where my OrangeLink can't be connected to sometimes if I force close the app or restart my phone?
The only reliable solution I've found in that case is to switch AAPS to Virtual Pump, then open nRF Connect and connect then disconnect from the OrangeLink, then switch AAPS back to Medtronic, and it initializes and connects fine.

bubbledevteam:
> Would this fix the issue where my OrangeLink can't be connected to sometimes if I force close the app or restart my phone?
> The only reliable solution I've found in that case is to switch AAPS to Virtual Pump, then open nRF Connect and connect then disconnect from the OrangeLink, then switch AAPS back to Medtronic, and it initializes and connects fine.

yes

andyrozman:
> > This functionality is totally unnecessary... Once you have RL configured there should be NO MORE search done... We do search in configuration and that is only place this should be done...
> 
> We have tested and found that some Bluetooth devices must be searched before they are connected or they cannot be connected

Then please make this optional. So add setting into configuration and users can turn this on or off. Something like "Search for device before connecting" and in description write that some RileyLink replacement require this in brackets OrangeLink. And this should be off by default.

StephenBrown2:
Limiting it to "Orange" devices as has been done in [`8a9a5da` (#489)](https://github.com/nightscout/AndroidAPS/pull/489/commits/8a9a5da6f59a52afbffabc50fa64029f16fabc0b) seems like it would be sufficient to me.

andyrozman:
While this will cover Orange devices, maybe some other users have same problem with EmmaLink or RileyLink, which is why I would like to have this configurable. Using this option might work better on some BT implementations. 

So please make it configurable and then we will need to note this in documentation too, both for MDT and Omnipod.

bubbledevteam:
> While this will cover Orange devices, maybe some other users have same problem with EmmaLink or RileyLink, which is why I would like to have this configurable. Using this option might work better on some BT implementations.
> 
> So please make it configurable and then we will need to note this in documentation too, both for MDT and Omnipod.

Ok ,I have two designs for the configuration , we can discuss which one is more user friendly.

Design 1: 

image


Design 2:

image

StephenBrown2:
Personally I like the first option.

A note: This may not need to scan all the time. I've only noticed it when the connection or app was disrupted, so it would only need to scan again on app start, or if it cannot reach the OrangeLink. While scanning on every connect works, perhaps it could be made less aggressive and only try on app start and failure to connect?

andyrozman:
1st one... It needs to be visible when you open configuration. We don't use dialogs in configuration. Please extend text to "Scan before connecting to OrangeLink, it should improve connections (can also be used with other RileyLink clones, if needed)" and put it into rileylink module (so that can it can be used by both MDT and Omnipod configuration).

bubbledevteam:
> 1st one... It needs to be visible when you open configuration. We don't use dialogs in configuration. Please extend text to "Scan before connecting to OrangeLink, it should improve connections (can also be used with other RileyLink clones, if needed)" and put it into rileylink module (so that can it can be used by both MDT and Omnipod configuration).

I am a little confused about "put it into rileylink module" you mentioned, is it the place as design 1？If that, I just need to change the text only

andyrozman:
RileyLink module is used by both Medtronic and Omnipod-Eros, so if you put string into RileyLink module, both pumps can use it. Afterwards you need to change pref_medtronic.xml and omnipod-eros-preferences.xml... This setting needs to be in the main driver settings page, not on the one you have showed...

bubbledevteam:
> RileyLink module is used by both Medtronic and Omnipod-Eros, so if you put string into RileyLink module, both pumps can use it. Afterwards you need to change pref_medtronic.xml and omnipod-eros-preferences.xml... This setting needs to be in the main driver settings page, not on the one you have showed...

could you pls show me which page？

StephenBrown2:
Take a look at the changes done for showing the battery level for Orange/EmaLinks: https://github.com/nightscout/AndroidAPS/pull/430

StephenBrown2:
> could you pls show me which page？

You'll need to rebase on `dev`, and then edit the following:
```
omnipod-eros/src/main/res/xml/omnipod_eros_preferences.xml
```

The preferences are exposed in the General preferences page, but (for me) under the medtronic heading. They would show up under omnipod eros settings if that plugin was enabled.

andyrozman:
Could you please add configuration in main preferences for medtronic and omnipod, instead into this activity. And please close the other PR.
Where you need to put configuration in, is here:
omnipod-eros/src/main/res/xml/omnipod_eros_preferences.xml
/medtronic/src/main/res/xml/pref_medtronic.xml

And please follow @AdrianLxM instructions what needs to be changed.

bubbledevteam:
> Could you please add configuration in main preferences for medtronic and omnipod, instead into this activity. And please close the other PR.
> Where you need to put configuration in, is here:
> omnipod-eros/src/main/res/xml/omnipod_eros_preferences.xml
> /medtronic/src/main/res/xml/pref_medtronic.xml
> 
> And please follow @AdrianLxM instructions what needs to be changed.

As @StephenBrown2 replied, I need to rebase it on dev, that's why I opened a new one

PR #498, opened 2021-04-30
Fix: turn APSResult.bolusRequested into a function.

jotomo:
Previously this var was set while parsing determine-basal output, setting it to true if determine-basal provided a value which was assigned to var smb. However, the variable smb is changed when constraints are applied, making the val bolusRequested invalid afterwards. Hence, dynamically evaluate the smb var by turning bolusRequested into a function.


jotomo:
@Philoul Could you test this please? Thanks!

Philoul:
@jotomo with your PR I couldn't reproduce the problem after about 12 hours, (previously I had about one restart per hour)
Thanks

jotomo:
@MilosKozak ready for merge. Fixes the issue were the Insight received requests to issue 0 carbs and 0 insulin.

PR #502, opened 2021-05-05
Dev2 insight v2

Philoul:
This new version is very close to original Insight driver (so I think most of @TebbeUbben comments are taken into account here)

- [x] Check all comments done by Tebbe in #488 
- [x] migration of insight db to Insight room database (cf https://github.com/nightscout/AndroidAPS/pull/488#issuecomment-831082702)
- [x] remove additional `aapsLogger.debug`
- [x] add abstraction layer between Plugin and dao (suggestion of @AdrianLxM )
- [x] remove old db tables for insight

Tests done with Insight v2 and v3
Action from LoopPlugin:
- [x]  SMB
- [x]  Set Ext Bolus
- [x]  Cancel Ext Bolus
- [x]  Set TBR
- [x]  Cancel TBR

Manual Action from AAPS
- [x]  Bolus
- [x]  Bolus + Cancel during injection
- [x]  Extended Bolus (TBR emulation disabled)
- [x]  Cancel Extended Bolus
- [x]  Set TBR (when no TBR running)
- [x]  Cancel TBR

Manual Action in Pump (result in AAPS after pump synchro)
- [x]  Bolus
- [x]  Bolus + cancel on Pump
- [x]  Bolus Extended (running)
- [x]  Bolus Extended canceled in pump
- [x]  Bolus Multiwave (running) (Ok in AAPS for immediate bolus and extended bolus)
- [x]  Bolus Multiwave canceled in pump
- [x]  TBR set in pump
- [x]  TBR Canceled in pump

TherapyEvent (ok with Insight v2 and v3)
- [x]  Pump Battery change
- [x]  Insulin Change
- [x]  Site Change

Pump history Stop and start event
- [x]  set zero temp during pump is stopped


Philoul:
I think now most of code should be Ok for tests by other dev (new version very close to original Driver...)
=> I added some Todo for additional questions and verifications
=> I'll now start to test more deeply all functions

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -836,7 +829,7 @@ public PumpEnactResult cancelTempBasal(boolean enforceNew) {
         PumpEnactResult cancelEBResult = null;
         if (isFakingTempsByExtendedBoluses()) cancelEBResult = cancelExtendedBolusOnly();
         PumpEnactResult cancelTBRResult = cancelTempBasalOnly();
-        result.success((cancelEBResult != null && cancelEBResult.getSuccess()) && cancelTBRResult.getSuccess());
+        result.success((cancelEBResult == null || (cancelEBResult != null && cancelEBResult.getSuccess())) && cancelTBRResult.getSuccess()); //Fix a bug when Fake TBR is disabled and click on Cancel TBR button
```
Philoul, line 832: @TebbeUbben Here I propose this modification to fix a bug in current driver:
=> If Fake TBR is disable in setting, each time you click on "Cancel TBR" you get a communication Error Popup

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -906,21 +900,20 @@ private PumpEnactResult cancelExtendedBolusOnly() {
                     connectionService.requestMessage(cancelBolusMessage).await();
                     confirmAlert(AlertType.WARNING_38);
                     alertService.ignore(null);
-                    InsightBolusID insightBolusID = databaseHelper.getInsightBolusID(connectionService.getPumpSystemIdentification().getSerialNumber(),
-                            activeBolus.getBolusID(), System.currentTimeMillis());
+                    InsightBolusID insightBolusID = insightDatabaseDao.getInsightBolusID(serialNumber(), activeBolus.getBolusID(), dateUtil.now());
                     if (insightBolusID != null) {
-                        ExtendedBolus extendedBolus = databaseHelper.getExtendedBolusByPumpId(insightBolusID.id);
+                        /*  TODO() I don't know if we can remove bolck below (there is a readHistory after that will update AAPS database)
```
Philoul, line 905: Can we remove this block ? (I don't update database here because there is a readHistory later, NSUpload is no more necessary, and there is no delete or equivalent in pumpSync interface...)

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1202,36 +1195,63 @@ private void processHistoryEvents(String serial, List<HistoryEvent> historyEvent
...
-            temporaryBasal.durationInMinutes = (int) ((pumpID.timestamp - tbrStart) / 60000);
-            temporaryBasal.date = tbrStart;
-            temporaryBasal.source = Source.PUMP;
-            temporaryBasal.pumpId = pumpID.id;
-            temporaryBasal.percentRate = 0;
-            temporaryBasal.isAbsolute = false;
+            InsightPumpID stoppedEvent = insightDatabaseDao.getPumpStoppedEvent(pumpID.getPumpSerial(), pumpID.getTimestamp());
+            if (stoppedEvent != null && stoppedEvent.getEventType().equals("PumpStopped")) {             // Search if Stop event is after 15min of Pause
+                InsightPumpID pauseEvent = insightDatabaseDao.getPumpStoppedEvent(pumpID.getPumpSerial(), stoppedEvent.getTimestamp() - T.mins(1).msecs());
+                if (pauseEvent != null && pauseEvent.getEventType().equals("PumpPaused") && (stoppedEvent.getTimestamp() - pauseEvent.getTimestamp() < T.mins(16).msecs())) {
```
Philoul, line 1203: @TebbeUbben I added this to include the "Pause" in the zero temp if the Stop event is after 15 min of pause...

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1253,150 +1273,197 @@ private void processDateTimeChangedEvent(DateTimeChangedEvent event) {
...
+            bolusID.setStartID(event.getEventPosition());
+            insightDatabaseDao.createOrUpdate(bolusID);
             return;
         }
-        if (bolusID == null || bolusID.startID != null) {
-            bolusID = new InsightBolusID();
-            bolusID.timestamp = timestamp;
-            bolusID.bolusID = event.getBolusID();
-            bolusID.pumpSerial = serial;
+        if (bolusID == null || bolusID.getStartID() != null) {                        // TODO() Check StartID test is necessary
```
Philoul, line 1435: @TebbeUbben I don't understand here why you create a new bolusID if one is found with a StartID?
my first idea was to create a new bolusID in insight database only if `bolusID == null`

Philoul:
@MilosKozak should we keep old Insight db tables (to include a migration of datas to new Insight room tables..) or can I remove them ?

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
TebbeUbben: Yes. As far as I remember, this was necessary to properly sync canceled extended boluses if the initial entry was still in the queue.

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
TebbeUbben: Hmm, the relevant pause entry might still not be in the database if it's processed afterwards. Remember, we are processing history entries backwards for aligning the timestamps.

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
TebbeUbben: In rare edge cases two boluses can share the same ID. We don't want to overwrite that.

TebbeUbben:
My suggestion is to totally rework the history sync. I suggest to first read all the history events from the pump. Pass them into a database transaction and process them while the transaction has exclusive write access to the database.

Philoul:
@TebbeUbben The 2 points I have in mind on my side (for later improvement only) are
- convert driver to kotlin (I can try, if there are no tricky things)
- improve consistency between AAPS overview and Pump State (Pause / Stop status, see below)
=> For the rework of history sync, we could probably centralize all events in one table... 
=> I think it's better if you rework it yourself.

My suggestion would be to always have overview informations consistent with pump state:
- Pause event => Loop Suspended, and current TBR still running in overview (that way no new TBR can be set from loop) (currently Closed Loop still running)
- Stop event => Loop Suspended, and 0 TBR is immediatly set to see it in overview (and if Stop occurs after 15 min Pause, then 0 TBR is started at Pause timestamp). (Currently Loop Suspended but I see Profile basal rate in overview)
- Start event => Loop enabled again (and 0 TBR ended if previous state was Stop) (Currently 0 TBR is only set afterward on Start event if previous event was Stop)

Philoul:
Note for @TebbeUbben , I fix with https://github.com/nightscout/AndroidAPS/pull/502/commits/62901923550b5ff8a8012b70388c12ad32b977dd a little bug with Insight v3 and site change on current driver: site change event is always added in the history when you change reservoir or fill tube (even if you select No in pump)
=> I read event amount to identify if the site was changed or not

@MilosKozak I think there is an issue with `pumpSync.insertTherapyEventIfNewWithTimestamp`, I always get 2 additional empty notes in Treatment tab (see screenshot after a reset database and only a site change)
Screenshot_20210509-012752

insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
Philoul: We are processing backwards, but all history entries are processed (so in database) when we start to loop on `pumpStartedEvents` to set 0 temp if necessary...
I made 2 tests (one with only Stop and Start events, and another with 15min of pause before Stop, for me it works...

MilosKozak:
tests need to be fixed. otherwise ready?

Philoul:
Tests fixed and Yes I think it's ok now

MilosKozak:
test is still crashing

Philoul:
Sorry, I'm still having trouble with tests (seemed to work the first time though)
Hope this time it's ok

PR #533, opened 2021-05-23
Fix Sens curve scale if alone on a graph

Philoul:
Shift of 100 is necessary for Sens Curve
I also fix (for dev) deviation slope scale if on the same graph than Sens curve (middle axe is not at 0 but at 100.0...)

In third commit I hide in menu the curves visible on others secondary graphs to avoid having the same curve selected on several graphs (fix scale issues when you had for ex. a curve selected on third graph and you select it a second time on first or secondary graph)
(sorry wrong label for first commit, I forgot to change it)


app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewMenus.kt
```diff
@@ -113,6 +113,9 @@ class OverviewMenus @Inject constructor(
                     if (m == CharType.PRE) insert = predictionsAvailable
                     if (m == CharType.DEVSLOPE) insert = buildHelper.isDev()
                     if (used.contains(m.ordinal)) insert = false
+                    for (g2 in g + 1 until numOfGraphs) {
```
MilosKozak, line 116: this is not necessary as graph menu may appear only on previously saved configurations
once you untick it, new UI will now allow selecting graph twice (line 115)

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewData.kt
```diff
@@ -268,13 +268,13 @@ class OverviewData @Inject constructor(
...
     var minRatioValueFound = -maxRatioValueFound
-    val ratioScale = Scale()
+    var ratioScale = Scale()
     var ratioSeries: LineGraphSeries<ScaledDataPoint> = LineGraphSeries()
 
     var maxFromMaxValueFound = Double.MIN_VALUE
     var maxFromMinValueFound = Double.MIN_VALUE
-    val dsMaxScale = Scale()
-    val dsMinScale = Scale()
+    var dsMaxScale = Scale()
```
MilosKozak, line 276: are you sure about this?
this `val` instance is added to all elements during creation (everywhere)
if you want to change scale of prepared data you should call `setMultiplier` on this instance

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewMenus.kt
Philoul: If you select on graph1 a curve already selected on another graph, the item is hidden in menu but the curve remains selected in graphs preferences... So curve is shown in 2 subgraph but with wrong scale on first one
This is to avoid selection of all selected curves in all graphs in menu

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewData.kt
Philoul: `Scale(shift)` (default 0.0)
We have to change "shift" value for sensitivity curve (100.0 if alone or with deviation slope on a graph to read percentage in scale or 0.0 if it is with other curves)

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewMenus.kt
Philoul: New UI doesn't allow on higher graphs, but if a curve is selected on for example the Graph3, you can select it again in Graph1, the item is hidden in menu for Graph 3 but the curve remains selected and shown in Overview
See screenshot on the left and middle Dev Slope on Graphs 1-3, with PR on the right, only curves not already selected are available for selection... (shorter menu but two clicks are necessary to switch a curve from one graph to another)
GraphMenu

We have another possibility, it's to show all curves in all graphs, and when you enable one curve on a graph, you disable it in all other graphs in graph preferences... (only one click to switch a curve from one graph to another, but longer menu)

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.kt
```diff
@@ -153,20 +154,32 @@ class GraphData(
...
     // scale in % of vertical size (like 0.3)
-    fun addDeviationSlope(useForScale: Boolean, scale: Double) {
+    fun addDeviationSlope(useForScale: Boolean, scale: Double, isRatioScale: Boolean = false) {
         if (useForScale) {
             maxY = max(overviewData.maxFromMaxValueFound, overviewData.maxFromMinValueFound)
             minY = -maxY
         }
-        overviewData.dsMaxScale.setMultiplier(maxY * scale / overviewData.maxFromMaxValueFound)
-        overviewData.dsMinScale.setMultiplier(maxY * scale / overviewData.maxFromMinValueFound)
+        var graphMaxY = maxY
```
MilosKozak, line 172: What I mean is that `overviewData.dsMaxScale` instance is added to prepared data series in `OverviewData` before using `GraphData` for rendering. During data preparation is scale needed not known.
Then if you here assign to `overviewData.dsMinScale` diffent object, calling `setMultiplier` has no effect to scale used during  data preparation (or better say it has but only for next round, not for current data)
To apply scale immediately you must apply changes to current `Scale` object not create new

MilosKozak:
I updated `Scale` class to allow set `shift`

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/graphData/GraphData.kt
Philoul: Thanks, I'll take a look this evening to improve it.

Philoul:
👍 It works best now that I can set shift

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewData.kt
```diff
@@ -286,13 +286,13 @@ class OverviewData @Inject constructor(
...
+    var dsMinScale = Scale()
```
MilosKozak, line 295: revert to `val` here

Philoul: Sorry I forgot to revert them...
Done

MilosKozak:
👍

PR #543, opened 2021-05-28
Fix cancel extended bolus (Overview)

Philoul:
All Pump : Fix Overview (no information visible below icon and allow click on icon to show popup)
Insight Driver : Fix "Cancel Extended Bolus" if `isFakingTempsByExtendedBoluses` is `false`

### DIFF
--- a/insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
+++ b/insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
@@ -1450,6 +1450,12 @@ public class LocalInsightPlugin extends PumpPluginBase implements Pump, Constrai
                             bolusID.getId(),
                             PumpType.ACCU_CHEK_INSIGHT,
                             serial);
+            if (!isFakingTempsByExtendedBoluses())
+                pumpSync.syncStopExtendedBolusWithPumpId(
+                    timestamp,
+                    event.getEventPosition(),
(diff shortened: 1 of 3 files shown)


insight/src/main/java/info/nightscout/androidaps/plugins/pump/insight/LocalInsightPlugin.java
```diff
@@ -1450,6 +1450,12 @@ private void processBolusDeliveredEvent(String serial, BolusDeliveredEvent event
                             bolusID.getId(),
                             PumpType.ACCU_CHEK_INSIGHT,
                             serial);
+            if (!isFakingTempsByExtendedBoluses())
```
MilosKozak, line 1453: the above `syncExtendedBolusWithPumpId` doesn't need this?

Philoul: I thought, but I found a bug if I try to Cancel an extended bolus from button on Action Tab when fake tbr is false:
Extended Bolus is correctly cancelled on pump, but it remains active in overview, Action Tab and treatment tab...
I thought I miss this pumpSync call in driver (it works with this call it's ok), but it's possible that the bug is not in the driver but in pumpSync interface. (I didn't check with virtual pump...)

Philoul: Finally it works with Virtual Pump, So I have to understand what's wrong in insight driver (because it works with fake TBR...)

PR #544, opened 2021-05-28
Remove redundant keys in allowedKeys and one extra "

Philoul:
Simplification proposal (to check/review before merge or close if KO, I'm not a specialist of this plugin!)
I don't know if the " removal on `danar_visualizeextendedaspercentage` key may have side effect on Open Human synthsis (I think it's a typo)


MilosKozak:
did you try to search for active keys?

Philoul:
No, I just found this file because I searched for "hard coded" strings concerning Insight preferences to be sure all are replaced by key strings in resources inside the driver...
I'm agree it's not a very good solution to have hard coded strings here.

PR #554, opened 2021-06-06
Add feature to suppress food intake warning until a configurable threshold is reached.

juehv:
Code modifications are fairly simple. Default behavior does not change. Threshold can be defined in preference menu in subentry "loop". Tested on my dev rig (Pixel 3a + Insight).


Philoul:
1) I don't understand the difference with "Minimum carbs required for suggestion" setting available in OpenAPS SMB Plugin...
2) never include Strings in other languages, only english (translations are made with crowdin...)

juehv:
@Philoul As far as I understood is the "Minimum carbs required for suggestion" feature only available when OpenAPS SMB is used. Am I wrong ?

Philoul:
@juehv  I'm not sure, but I think "Carb Required feature" is available only in OpenAPS SMB...
Do you have Carbs required alerts with OpenAPS AMA ?
For translations, you're welcome here: https://crowdin.com/project/androidaps (new strings are available in crowdin for translation if the PR is merged in dev)

juehv:
@Philoul OK I'll take a look at the code if the carbs required feature is available with other algorithms. And adapt the pull request accordingly.

Thanks for the link, I'll take a look :)

PR #561, opened 2021-06-11
Fix delete logs

Philoul:
I propose to add a setting in maintenance plugin for the number of logs to keep (tell me if proposed default, min and max values are ok or not)
I changed order to delete older files


MilosKozak:
i'm not much fun of adding more options.
let's keep it fixed ..... maybe larger if in engineering mode?

Philoul:
Ok (I will revert the commit for preferences)
I thought it was a forgotten preference because `key_logshipper_amount` is not used anywhere else in the code...
=> Do you planned to use dedicated settings for normal or engineering mode in another file ?
I think 30 is to small for engineering mode (just a little more than one day of history on my phone...), an number above 240 or 300 would be better...

PR #605, opened 2021-07-11
[WIP] Graph Resize on Overview

Philoul:
I propose to modify OveriewData to always prepare data on Max range (from `fromTimeData` = `toTime - MaxRange` to `endTimeData` = `toTime + predictionDuration`)
=> When we resize graphes, they are immediatly updated to selected range even if the new range is longer than previous one (all datas are already prepared)
=> I replaced Max values for scale by tables of calulated max values according to selected scale

I propose a Draft PR because this modification doesn't currently work with HistoryBrowser (management of fromTime and toTime is a bit different compared to overview, and I don't know how  "<", ">" and ">|" buttons will work in Historybrowser if we are accross 2 days...)

If you have comments on this proposal


Philoul:
I close this PR even if it works for overview (not with historybrowser)...
I'm not fully satisfied by the results (it's seems aaps restart a bit more often with it, about once per day or per 2 days)

The number of calculation done to prepare overview still remains very high and probably very redundant (most of the time we add an additional entry "now", that doesn't require a new preparation of 24h of graph data, only last point and previsions...)

Other idea is to keep prepared lists as cached data and only calculate what should be updated (from timestamp to max(timestamp+dia, now) for iob, timestamp + max meal duration for cob...)?

I discussed with new "testers" of dev branch and ithey considered current solution as a bug...

PR #634, opened 2021-08-17
Fix Overview LoopDialog (Disable Loop)

### LINKED ISSUE #630: Only to be sure that the Dev's be aware of some issues. All are in dev versions with the new database until today.
- [**_Fixed by @Philoul _**]
For Philoul and Milos it works now, in my case it still doesn't work.
**Profile switch is  buged.:**
You can't change your personal profile say "Workday" to "Free Day.
If you do that, it was shown in NightScout as done, but AAPS is doing nothing and has still the old profile active.
To do the wanted profile switch, you have to change to the "Local Profile" first, then you can switch to the wanted one.
Because this Bug, the Automation  will also not work with profile switches.
**- This happened only, if the Basal rate was in both profiles the same and only the Target, ISF, or IC was different.** 

- [**_Fixed by @Philoul_**] 
**Disable Loop:**
If you disable Loop, then you can't activate it again.
You have only the option to disconnect the pump. If you do that, you can activate the Loop again and reconnect the pump.

- **_Fixed by Philoul, but TDD needs time_**
**Statistic:**
The last fixes will not work for TDD.
All will shown now immediately,  but not TDD. 
TDD needs as long as before.

- [**_MilosKozak says, "AMA doesn't provide this value"._**]
**Autosense Target:**
Autosense Target was not shown on main page, if you are using AMA Algorithm. 
It was shown if you use SMB Algorithm. 

- [**_Fixed by Philoul_**]
**Wrong values for Daily Basal in local profiles:**
All values of the daily Basal, in all local profiles, are showing the same number.
This is the daily basal number of the last saved, or resettet profile.
It's only a visual issue.

@MilosKozak 
If you have read that, you can close this issue. Wanted only that you are aware of this issues.

Philoul:
I tried to simplify User Interface of Loop Button:
When Loop is Disabled => Only Enable Loop Button
When Loop is Suspended => Only Resume Loop Button
When Pump is Disconnected => Only Reconnect Pump Button

I fixed #630 concerning "Disable Loop" in Loop Dialog, but code still remains complicated (I hope I didn't generate side effects...)
I'll try to see if I can simplify more...


Philoul:
In my latest commits I replaced the several `if ... else` by a `when` statement
=> Now code is cleaner and I found some other situations that was not correctly treated previously...

1) When pump is "stopped" (icon "Loop_Paused" without duration)
=> I propose to hide all buttons (I think first step is run pump), should we disable the dialog, show a text or keep some buttons ?

2) when profile is not valid
=> I kept one bloc of code (all buttons hidden) and comment another bloc (I think we can remove everything, probably old code remaining from previous menu but I'm not sure...)

Tested and ok with Insight pump, not tested with other pumps...

swissalpine:
First of all and again: Thank you for your work!
When simplifying, I do not agree with everything.
For example, I find it helpful to be able to extend the disconnect time without having to press the reconnect button first.I rebuilt the dialog here that way (not pretty, but functional):
Screenshot_20210819-084932


Overall, I think it would be good to show all buttons that can be pressed next without problems (e.g. change from suspended to disconnect), without unnecessary intermediate steps.

Philoul:
On my side I always shortened and never extend duration, but it's a good idea (I'll test for suspend and disconnect buttons if it works and with latest version it's easy to add it).

If you allow direct jump from one state to another I'm afraid of a too complex UI with difficulties to manage and understand what will happen when current state is finished (for ex if I disconnect for 15min pump during a 2 hours of suspend loop, what is expected result for the user after the 15min and pump reconnection... Closed Loop or Suspend Loop? Is it the same expected result for all users? Which result the user will have according to current implementation?

That's why I think it's better to avoid cross actions...

swissalpine:
There are a few use cases where I regularly switch from one mode to another, especially in preparation for sports activities:
With Suspend Loop or LGS I control or reduce the IOB before and during sports, and while one of the modes is active, I additionally select disconnect pump for a certain time so that the IOB drops even further. When disconnect pump expires, the old mode MUST be active again, otherwise the entire sport adjustment does not work, because a few minutes of normal loop can mean a lot of insulin, especially at high blood glucose levels.
Also, I use the disconnect pump buttons to extend or shorten the duration of the 0% TBR.
In addition, it happens again and again that I activate Suspend Loop a few hours before sport and then switch directly to LGS at the start of sport.
All the changes described here work completely unproblematically: once they are only mode changes, the other time only TBRs of 0% are set.
And this is not an individual use case: I know numerous looping athletes who use this.  Without these combinations, AAPS would be much worse to use for me as an athlete, sometimes even more dangerous: e.g. if a disconnect ends during the sport without falling back to the previously active LGS!
Thanks for our understanding.

Philoul:
@swissalpine, I'm not sure to understand exactly your request...
Currently we have one active loop mode, (open loop, LGS and closed loop), and from this active mode we can decide to disable loop or select temporary actions (suspend loop or disconnect pump) that will go back to previous active mode of the loop when temporary mode is finished.
I prepared a commit (not currently pushed because my computer is off for 3 days), to add the possibility, when pump is disconnected (or loop suspended), to directly select another duration to extend (or shortened) current state directly.
I'm just a bit afraid that for a very specific usage, the request of direct cross selection of temporary functions, with different durations,  leads to something too complex to understand for a majority of users...
So if you explain me exactly what is missing and associated logic, and if Milos agree, of course I'll implement if it can improve the usage of aaps for a majority of users...

swissalpine:
Nothing is missing, it is good as it is. But you write:

When Loop is Disabled => Only Enable Loop Button
When Loop is Suspended => Only Resume Loop Button
When Pump is Disconnected => Only Reconnect Pump Button

And I am concerned that I can no longer switch directly from suspend loop to lgs, that while suspend loop or lgs is active, I can no longer select disconnect pump. Especially the disconnect pump options must always (!) be possible, no matter if closed loop, suspend loop, lgs or disable loop is activated.

Another thing is the possibility to always choose a new disconnect pump time, because this always cancels the current TBR and sets a new one. This is completely unproblematic! Only this would be a change to the current behavior, a "new feature" (which was allowed some time ago).

I hope you understand what I mean: One point is that I'm afraid of the intended simplification, the second point is a change that shows options that are currently (i. m. O. unnecessarily) hidden.

---

Another question:
Another question for me is why the TempTarget dialog removed the ability to define a start time: Previously it was possible to specify at 9am that a higher TempTarget should start at 12pm ...
Since I some times observe that usefull possibilities disappear with layout changes, I am perhaps overcautious here.

Philoul:
- With latest commits, you can now always disconnect pump (when loop is enabled (Closed, LGS or Open Loop), disabled or suspended)
- You can only suspend loop when loop is enabled (Closed, LGS or Open Loop)
- When loop is suspended, you can resume Loop, directly select another duration, or disconnect pump (but in this case pump disconnection will replace the "suspend state", so when pump is reconnected, manually or automatically, the loop will be enabled immediatly, even if the duration of the previous suspend is longer than the duration of disconnection)
- When pump is disconnected, you can only reconnect pump or directly select another duration
- When OpenLoop is selected you can select Closed Loop only if objective 6 is completed (@Milos, I cannot use constraintChecker.isClosedLoopAllowed because it's always false in safetyPlugin if open loop is selected, even if objective 6 is completed, I don't understand the reason of this constraint...)
- When pump is stopped or suspended, and when profile is not valid, no buttons are available (fix another issue in current master and dev)

The logic is now simplified and can be updated easily within the `when` block...

swissalpine:
Thank you very much! That reads very plausible to me.

swissalpine:
I have merged your PR.
Everything works so far.
However, there is not always a return to the former state:
Suspend Loop (2 hours) -> Disconnect pump (15 min) -> Reconnect pump -> Closed Loop
I think here the previous state (i. e. Suspend Loop) should be restored. This is at least what I would expect: I disconnect my pump (for a shower) and I would see the state which was choosen before.

Philoul:
@swissalpine 
> However, there is not always a return to the former state:
Suspend Loop (2 hours) -> Disconnect pump (15 min) -> Reconnect pump -> Closed Loop

Exactly, that's what I explained in my third bullet point, and I think it might be confusing for users (this is the standard way AAPS works, LoopDialog is only a user interface but doesn't manage the memorising of loop and pump states)
=> That's why I think it's probably better to not allow direct switch from Suspend loop to Disconnect pump: if we hide disconnection menu when loop is suspended, there is 2 actions to switch to disconnection (you first resume loop and then you disconnect pump), but it's easier to understand for users...

Philoul:
In latest commit, I only disabled the possibility to directly select "Disconnect Pump" when loop is suspended (2 actions to directly switch from Suspend to Disconnect, but no risk of misunderstanding by the users).
All other comments are not in dialogLoop (this dialog is only an interface and doesn't manage the way AAPS works).

@swissalpine 
> The only thing that surprised me is the following: When I have supend loop enabled and set a TBR of 0% via actions tab, after expiration or when I cancel the TBR suspend loop is active again. Disconnect pump does the same thing, setting a TBR of 0%.
> This difference in behavior is faulty, in my opinionm and should be fixed.

I agree, I think we shouldn't be able to Cancel the 0 TBR set by a "Disconnect Pump" or automatically "Reconnect Pump" if the 0 TBR is canceled.
But it's not managed in LoopDialog, so not in this PR (can you please open another issue)

swissalpine:
Sorry, maybe my English is too bad to be understood:
I want "Suspend Loop" not to be terminated if "Disconnect pump" was active in the meantime.
And maybe I understand the code wrong, but if the lines 354 ff. are changed, this is also possible:
``` 
            R.id.overview_resume {
                uel.log(Action.RESUME, Sources.LoopDialog)
                disposable += repository.runTransactionForResult(CancelCurrentOfflineEventIfAnyTransaction(dateUtil.now()))
                    .subscribe({ result ->
                        result.updated.forEach { aapsLogger.debug(LTag.DATABASE, "Updated OfflineEvent $it") }
                    }, {
                        aapsLogger.error(LTag.DATABASE, "Error while saving OfflineEvent", it)
                    })
                rxBus.send(EventRefreshOverview("suspend_menu"))
                commandQueue.cancelTempBasal(true, object : Callback() {
                    override fun run() {
                        if (!result.success) {
                            ErrorHelperActivity.runAlarm(ctx, result.comment, resourceHelper.gs(R.string.tempbasaldeliveryerror), R.raw.boluserror)
                        }
                    }
                })
                sp.putBoolean(R.string.key_objectiveusereconnect, true)
                return true
            }
			
			R.id.overview_reconnect -> {
                uel.log(if (Action.RECONNECT, Sources.LoopDialog)
                commandQueue.cancelTempBasal(true, object : Callback() {
                    override fun run() {
                        if (!result.success) {
                            ErrorHelperActivity.runAlarm(ctx, result.comment, resourceHelper.gs(R.string.tempbasaldeliveryerror), R.raw.boluserror)
                        }
                    }
                })
                sp.putBoolean(R.string.key_objectiveusereconnect, true)
                return true
            }
            ```
Because here Resume and Reconnect are treated analogously, although they are different things. If we split it, we get the desired behaviour I thing ...

Philoul:
@swissalpine I think you can currently have have what you want without any modification of AAPS if you suspend loop and select a zero temp, then loop will remains suspended after the end of the zero temp if suspend is not finished...

PR #694, opened 2021-09-26
Add files via upload

fossy:
This never workt for me. I had to correct it in every new AAPS version. So I try it this way...

Upload of records does not work correct. It allways try to upload after connection is closed (?)
Dowload never stops and upload nearly never appear (?)

- changed to extra Thread for only reactivate up/downloads
- upload as last step of downloads
- stop connection if all is done


MilosKozak:
nsclient is completely different in dev branch. This is unmergeable
Test DEV and if you think it still needs this improvemt create new PR targeting dev

Issue #645, opened 2021-08-24
[2.8.2.2 Dev] Sometimes manual bolus are not correctly recorded in database


Philoul:
After several month of tests, I found about one week ago a very strange bug with manual bolus (done with "Insulin Button" on main screen). I saw this bug 3 times, 2 times last week, and last time this morning at 9h59.
(Hopefully the great majority of boluses are correctly recorded)

This morning at 9:59, I made a manual bolus with Insulin button, this bolus was correctly delivered (I saw the Pop Up with the bar graph), but later when I wanted to make another bolus, I saw that previous bolus made at 9h59 was missing in IOB graph and in Treatment tab, I could only see it in User Entry and in the history of my Pump ???
- I tried to analyse Log, but and don't see anything that could explain why this bolus is missing in treatments and IOB calculation
- In NS/Synchronization, I have only "Upload data to NS" enabled, all other settings are disabled
- Pump Accu-Check Insight

![BugManualBolus](https://user-images.githubusercontent.com/52934600/130680605-6697a7a0-5b18-42f0-a906-303efec61c3a.jpg)

I attached an Excel file with concatenation of log files and extract of each on of the 3 boluses made this morning:
- First tab with complete log (I only removed D/GLUCOSE and D/AUTOMATION to limit the number of lines)
- Second Tab Bolus 10.51U made at 08:04 with QuickWizard, (everything is OK)
- Third Tab Bolus 4U made at 09:59 with Insulin Button, correctly delivered in pump but not shown in Treatments, IOB calculation and IOB graph
- Fourth Tab Bolus 4U made at 11:15 with Insulin Button (everything is OK)

[LogBolusKO.xlsx](https://github.com/nightscout/AndroidAPS/files/7041758/LogBolusKO.xlsx)

overfrenk:
It happened to me 2 times too! I had lunch and I used the insulin and carbohydrate button, the bolus was correctly delivered and the carbohydrates were correctly registered but the insulin was not! In fact aaps had started doing 1U boluses via SMB and I noticed it. To solve the problem I then did a no delivery bolus to block SMB.

Philoul:
@overfrenk which pump?

overfrenk:
@Philoul accu-check insight

Philoul:
Thank's
I don't know if it's in the driver or not but I also have an Insight pump...
I propose to add logs and database verification in driver to help me...

MilosKozak:
[...]

PR #704, opened 2021-10-01
Fix missing bolus Insight Driver

Philoul:
Fix bug in calculated parseRelativeDate (existing in current master, but concerns only boluses manualy done in pump)
Replace in database Calculated timestamp from pump by Android timestamp when bolus is launched from AAPS


Philoul:
fix #645

PR #756, opened 2021-10-18
 Minor UI adjustment

erobinson:
Super tiny adjustment so that the battery % is not bumped up against the right side of the screen
(First PR, sorry if I missed something 😄  )


erobinson:
image

PR #804, opened 2021-10-29
DASH History colors

vanelsberg:
Add some color to the DASH history view.

### DIFF
--- a/omnipod-dash/src/main/java/info/nightscout/androidaps/plugins/pump/omnipod/dash/ui/DashPodHistoryActivity.kt
+++ b/omnipod-dash/src/main/java/info/nightscout/androidaps/plugins/pump/omnipod/dash/ui/DashPodHistoryActivity.kt
@@ -211,8 +211,8 @@ class DashPodHistoryActivity : NoSplashAppCompatActivity() {
             }
         }
 
-        private fun setTextViewColor(includeresult: Boolean, textview: TextView, record: HistoryRecord) {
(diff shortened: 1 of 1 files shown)


Philoul:
@vanelsberg We already have a set of colors in User Entries for all kind of events, the colors chosen for Dash history are all different...
It's probably better if we standardize colors (Bolus/SMB dark blue, TBR Cyan, Site Change/Prime dark grey, profile switch white, CarePortal event (alerts) Yellow...)
image

See set of colors in core module (`utils/userEntry/UserEntryPresentationHelper`)
`    fun colorId(colorGroup: ColorGroup): Int = when (colorGroup) {
        ColorGroup.InsulinTreatment -> R.color.iob
        ColorGroup.BasalTreatment   -> R.color.basal
        ColorGroup.CarbTreatment    -> R.color.carbs
        ColorGroup.TT               -> R.color.tempTargetConfirmation
        ColorGroup.Profile          -> R.color.white
        ColorGroup.Loop             -> R.color.loopClosed
        ColorGroup.Careportal       -> R.color.high
        ColorGroup.Pump             -> R.color.loopDisconnected
        ColorGroup.Aaps             -> R.color.defaulttext
        else                        -> R.color.defaulttext
    }`

vanelsberg:
Good point!. Did not think about this. But DASH history is not about treatment?
I did not want to make history like a Christmas tree ;-) But will see if I can match some of this on group colorId's!

Issue #816, opened 2021-10-30
[3.0-beta2] Pod failure during 1st pod deactivation


kristinsmotz:
Installed 3.0beta2 today, going from beta1 with an active Dash pod. Had to change the pod a few hours later. As the pod was deactivating (after it connected but before the process finished) I got an "unknown pod failure" announcement along with the scream of death on both phone and pod for a few seconds. When the deactivation finished both stopped screaming but the phone kept playing an alert sound every second. Then a system notification reading just "none" popped up, which I swiped away. At this time there was no alert inside AAPS to dismiss/snooze visible and no apparent way to get rid of the alert. I tried exiting the app, then filling and activating the new pod as fast as I could, but the "be bop be bop" didn't stop until I shut down the phone and restarted it. 

This happened at 22.23

AndroidAPS 3.0-beta2
Build: f49c93f22-2021.10.30-15:01
Remote: https://github.com/nightscout/AndroidAPS
Flavor: fullrelease
Nightscout version: 14.2.2
Engineering mode enabled 

[AndroidAPS._2021-10-30_22-23-54_.0.zip](https://github.com/nightscout/AndroidAPS/files/7447655/AndroidAPS._2021-10-30_22-23-54_.0.zip)

![received_2603697803094351](https://user-images.githubusercontent.com/71758682/139559468-02e19f0d-649a-43cf-986a-ad51fb344671.jpeg)

PR #817, opened 2021-10-30
fix deactivation

avereha:
Fixes https://github.com/nightscout/AndroidAPS/issues/816
Create one alert notification only if the pod is in alarm state, not if the pod is deactivated.


omnipod-dash/src/main/java/info/nightscout/androidaps/plugins/pump/omnipod/dash/OmnipodDashPumpPlugin.kt
```diff
@@ -340,20 +340,23 @@ class OmnipodDashPumpPlugin @Inject constructor(
...
-                    pumpId = Random.Default.nextLong(),
-                    pumpType = PumpType.OMNIPOD_DASH,
-                    pumpSerial = serialNumber()
+
+            podStateManager.alarmType?.let {
+                showNotification(
+                    Notification.OMNIPOD_POD_FAULT,
+                    it.toString(),
+                    Notification.URGENT,
+                    R.raw.boluserror
```
avereha, line 349: There is still the issue that this sound does not stop playing when the notification is dismissed.

Issue #876, opened 2021-11-08
[2.8-3.0][Usability] Change pump reconnect icon to be green


dlvoy:
Currently, loop reconnect icon is exact same gray colour like icons used to disconnect pump.
It is confusing while on reconnect dialog it is not aparent that reconect icon do positive action and resumes pump and loop.

As a proposed fix - reconnect icon `ic_loop_reconnect `colour should be changed from gray (`@color/loopDisconnected`) to green (`@color/loopClosed`)

dlvoy:
I will work on PR to fix that

PR #877, opened 2021-11-08
Make reconnect pump icon easier to understand by making it green

dlvoy:
Using same gray for reconnect icon as for disconnect icons is confusing.
By changing it to same green, as used for other active loop icons, improves eligibility and makes icon intent clearer.

**Before**
Before

**After**
After


dlvoy:
This is fix for issue #876

PR #887, opened 2021-11-09
Speed up TDD statistics calculation

### LINKED ISSUE #882: [3.0 beta4] Extremally slow TDD calculation
On dialogs that use TDD stats - Statistics & Profile helper - calculating stats for TDD can take more than 30 seconds.
Propably it is caused by executing Room DB SQL query by `getTempBasalIncludingConvertedExtended` in the loop few tousand times at `TddCalculator::calculate`

dlvoy:
Fixes #882 by aggregating DAO query for temporary basals and moving it outside loop


Philoul:
Tested this morning with Samsung A40 (A11), Insight pump with Extended Bolus enabled
=> between 1min40s and 2min5s to get results with latest dev
=> 15s in first trial, less than 1s in second trial to get result

Acceleration is impressive 👍 

Exactly same values (Total, Bolus, Basal) for each of the 7 days and the average 👍

PR #891, opened 2021-11-09
Glunovo support

MilosKozak:


DanielEColi:
In the [release notes](https://androidaps.readthedocs.io/en/latest/Installing-AndroidAPS/Releasenotes.html#release-notes) for version 3.2, it mentions "Glunovo CGM support," but I can't find any manual on how to set up a closed-loop system with the Glunovo CGM. Additionally, there are no settings related to data transmission between apps within the Glunovo application.
Is it possible to use Glunovo CGM with AAPS? And how to do if it can?

PR #907, opened 2021-11-11
dash: do not stop the fake tbr on pod activation

avereha:
If the previous TBR was set by a different pump it would not work anyway.

And now `connectNewPump` cancels any active TBRs anyway, after https://github.com/nightscout/AndroidAPS/commit/85178ee19ff628dfe4422409973c6f8ae7b511b5


MilosKozak:
isn't this resoveld by my recent commit?
i read `PumpType` and serial from running TBR

avereha:
Because of your recent commit, this is not needed anymore.

Before: this worked only for Dash. When deactivating a pod, we set a fake TBR of 0 with a duration of 3 days. When activating a new pod, we would cancel it. We always use the pump serial 4242 on Dash.

Now: we don't need to cancel the fake TBR because `connectNewPump`(on the next line) does this for us. It also works between pumps, if Eros creates this fake TBR, it would still be cancelled by `connectNewPump`

avereha:
What is by "this is not needed anymore" is that we don't need to cancel the temp basal after https://github.com/nightscout/AndroidAPS/commit/85178ee19ff628dfe4422409973c6f8ae7b511b5

Issue #801, opened 2021-10-29
[last dev] AAPS temporary frozen after a time rescale of graph


Philoul:
No crash, but just annoying, very often, when I change time range of the graph I get a popup AndroidAPS is not responding (always press Wait button and it's Ok).
I also get some temporary artefact in the graph during the following 5 min (vertical bar in the left, wrong scale for TBR...)
![Screenshot_20211025-080253_AndroidAPS](https://user-images.githubusercontent.com/52934600/139390550-f0eccdd6-c1ef-4365-8c19-3a62d2193628.jpg)

![Screenshot_20211025-080315_AndroidAPS](https://user-images.githubusercontent.com/52934600/139390660-a9ef1420-d532-42c4-9edf-2cdcfdbc8651.jpg)

Philoul:
Copy paste my comment from #822 to #801 (most of it is linked to this issue).

> I tried to fix it a few months ago with always preparing data for max range of data for all curves (24h + prediction) and return selected range.
> I made a PR (not merged) that fixed the temporary vertical bar on the left and displayed immediatly all curves on rescale, but I closed it because it was not compatible with History Browser 😞...
> My other idea was to only calculate required data on new entry or each 5 min (most of time 99% of graph values are already calculated, and it's only when we enter insulin or carbs in the past that we have to calculate again a subset of values), keep in memory 24 or 48h of graph values and save older one in database to be able to show immediatly all curves in historybrowser (without calculation) and in overview...

MilosKozak:
caching simliar to what you describe is in place

Issue #899, opened 2021-11-10
Crash on changing Overview graph granularity


rustymonkey:
When changing the granularity of the graph on the Overview tab, AAPS crashes.

0 AAPS is looping / graph is set to display 'basals' and 'activity' with no other graphs (graph1, graph2,...) enabled
1 Tap-hold on the graph to change granularity
--> after performing step 1 two or three times, AAPS crashes 

Note that the issue can be consistently replicated.
--------------
Tested on AAPS 3.0 B5, running on an S7/Android 11 (Pixel Experience) connected to MDT 554 via RileyLink
Logs shared directly with Milos. Logs were generated after restarting AAPS immediately after the crash occurred.

MilosKozak:
i improved UI much after beta5 and now i cannot replicate. but it can have another reason.
update this issue after you test it with beta6 (when released) or current dev

rustymonkey:
Tested on 2.8.2.10-dev build 2021-11-11 11:40 and the issues are partially resolved (it no longer crashes). Tap-hold once always works as expected. But tap-hold 2 or 3 times results in:
1) the UI doesn't respond immediately to the 2nd/3rd tap-hold (always)
2) the Android dialog 'AndroidAPS isn't responding' appears (often)

From an end user perspective, it seems as if AAPS is doing something CPU-intensive after switching granularity once, and attempting to switch again while that background activity is going on doesn't put a stop to that activity, resulting in the the UI being non-responsive.

I've shared new logs illustrating this with the above build.

MilosKozak:
well if you change scale data in the past must be calcalculated
on fast phone it's under 10s of 100% utilization
on slow it's tens of seconds

if you press it more times recalculation is restarted which takes even more time
i optimized the calculations many times and i don't see any space for futher speedup
so. if if you are not happy with current state just buy faster phone

Philoul:
This issue is similar to #801 (I never had crash but very often App not responding)
It's a bit better on latest dev, but I still have App Not responding popup and the same behaviour explained by @rustymonkey in his last comment

Philoul:
> well if you change scale data in the past must be calcalculated
on fast phone it's under 10s of 100% utilization
on slow it's tens of seconds

if you press it more times recalculation is restarted which takes even more time
i optimized the calculations many times and i don't see any space for futher speedup
so. if if you are not happy with current state just buy faster phone

[...]

PR #911, opened 2021-11-12
Fix Graph rescale very slow

Philoul:
I modified OverviewData to always prepare MaxRange for 24h + Prediction duration and store MaxRanges in a table according to selected scale (6h, 12h, 18h or 24h) and prediction duration/selection
=> When we change scale I just return correct Array and associated MaxRange so everything is shown very quickly in overview

The Prepare duration is a bit longer (always 24h + prediction duration, and not only selected range), but rescale is much quicker
Only remaining problem during my tests, if I try to rescale graph during the countdown of received new BG value (there are to many thing to do for the App and I can still get App not responding popup)


Philoul:
the objective of this PR is to Fix #801 and #899

MilosKozak:
I don't like these copy/pasted classes. it makes hard maintaing the code

Philoul:
I'm agree, I don't like too, I just did it quickly based on my previous work (and having something that works for historybrowser)
I propose to revert last commit, but it will be necessary to find a solution for historyBrowser that doesn't work with modification made...

rustymonkey:
Tested this PR on 2.8.2.10dev 2021-11-14-12:49, and it resolves the issue. i.e.
- the UI now responds consistently to Tap-hold actions
- tap-hold actions no longer trigger the 'AndroidAPS isn't responding' dialog

vanelsberg:
@Philoul Thanks for taking this improvement. Much appreciated.

vanelsberg:
See also https://github.com/nightscout/AndroidAPS/pull/911

Philoul:
I started to work to find a solution to also improve history browser that share same code in a different way, but still remains bugs so work is still in progress...

Philoul:
Now it works with Historybrowser
- No optimisation for HistoryBrowser (too complicated to do without Room database)
- For Overview, I always prepare 24hours of graph data (with Arrays of all Max values calculated for 6, 12, 18 and 24, with and without prediction)
- On rescale, I just build Series on the fly according to range selected and I return associated max Values

@vanelsberg @rustymonkey Can you tell me if it works fine on your phones ? (In overview, with and without prediction curves, and in History Browser)

vanelsberg:
@Philoul Just installed 1196/911 on my test phone (realized I did not  need an active DASH Pod on it)
Test phone: Nokia 5.4. Decent phone, not high-end.

**Rescaling 6/12/18/24hrs:
**Yes!** This is exactly what I needed: not only is it fast. I can now select preference range without rotating to al others! Love it!

**History viewer:**
Imho this really is a huge improvement. User interface changed from "...never mind" to "interesting!" ;-)
I see still some lagging when switching dates, but also delays are "acceptable" compared to what it was. I think for the user it is clear the system is working on a lot of data, so the need to wait feels ok?

- *Bug?*:  I can still long-press the Graph selection button. It presents the range 6/12/18/24 selection dropdown as on overview. However it does not work. Would suggest to disable log-press when in the history viewer?

Just tested for about 20 minutes selecting different ranges on overview an history browser.
How not seen any problem. Also no 


At a 6 hour window

vanelsberg:
For PR#1196/911: Suggest to freeze code as is, test for any bugs and then make this into a PR?
(When this gets merged with dev I think you will have a lot of testers getting out any problems that are in)

Would love to have this on 3.0 release or else as a fix shortly after.

Philoul:
> Bug?: I can still long-press the Graph selection button. It presents the range 6/12/18/24

I'll test again, it should work in history browser too 🤔

Philoul:
> History viewer:
Imho this really is a huge improvement. User interface changed from "...never mind" to "interesting!" ;-)

#911 doesn't improve history browser, (@MilosKozak fixed it in v3.0, we have some ideas to improve it but not in v3...)., the only improvement will come from #1196 and #1142 (if  they work...)

vanelsberg:
@Philoul Ok? But why do I see both overview and history browser rendering so much better? 
Then it must be #1196 - can't imagine it's the Phone (Samsung Not 9 vs Nokia 5.4)? 

Will retest today with the Note 9 to see....)

Philoul:
Phone is probably the best explain...
I have huge differences between my loop phone (Samsung A40, very good 2 years ago but a nightmare now), and my test phone (power U20 bought 2 months ago for my dev config).
I never got Aaps not responding on my dev phone, and updates after each bg received or each graph rescale is quicker on my new dev phone...

vanelsberg:
@Philoul Phone (and possible database?) has impact, only I do not get it:
For history viewer Testphone (Nokia 5.4) performs a lot better for history viewer (even though it is buggy)

Decided to test on main phone (Note 9, which has much higher specs then the Nokia);
- Overview ranging also works great and performs better then current dev!
- History browser: seems to have a lot of problems and after changing range and dates started to lock up AAPS completely. So much that I had to kill the app (close/wait pop-up). On restart it quickly became unresponsive again. Solved by updating to current stable version.

**My Conclusion on PR#1196+911 on for now:**
- Overview ranging with pop-up menue does great. Love it!
- History browser is still buggy and locks up AAPS when using (result may depend on setup).
Will do some retesting on the separate PR's

Philoul:
@vanelsberg Thanks
I think #1196 is ok but still remains improvements in history browser in this PR to be able to merge...

vanelsberg:
@Philoul Agree.
Yes really like current functionality for #1196 works. Even with current performance.
Even asking myself if we need the current way for range switching: replace current long-press on overview to present the range pop-up?

rustymonkey:
Tested 2022-01-07 build with this PR and all is working well: 
- Changing the Overview graph scale works well.
- Notwithstanding your comments, history browser does seem a bit faster with this PR (though as with the regular build the graph flashes more as additional data is displayed (e.g. Activity + graph 1 with IOB & COB).

Philoul:
Much quicker on latest version

Issue #821, opened 2021-10-31
Dash timeZone and DST detection is not correct


avereha:
* DST detection for Dash is not working correctly.
* the timezone displayed in Dash overview is not correct(it always assumes DST)

Workarounds:
 - suspend delivery/resume delivery should set the correct basals on the pod
 - ignore the displayed timezone for "Time on Pod"

PR #921, opened 2021-11-13
dash: fix DST and timeOnPod

avereha:
Fixes https://github.com/nightscout/AndroidAPS/issues/821


Alamo04:
@avereha 
Didn't saw a problem before.
Germany: TimeZone was set to MEST.

With this fix:
Now AAPS will give an alarm with "Wrong TimeZone on Pod"
On Dash Tab, the field of the time on Pod is empty.

avereha:
Click "Set time" or switch profile and this should fix it!

Alamo04:
Worked 👍

Issue #798, opened 2021-10-28
[3.0-beta1] Medtronic SMBs show as Meal Boluses as well as Incorrect IOB calculation from temp basals


siggy226:
1. Using Medtronic, SMBs initially show correctly, but after 10-15 minutes get changed to appear as meal boluses.  I read issue  #739, was unclear what that setup was (I've never heard of a Dexcom app uploading to NS).  In my setup, only AAPS is uploading to NS, and switching to Dash driver with all other settings identical shows SMBs correctly.
![image](https://user-images.githubusercontent.com/65743481/139304184-ade157aa-dfbf-4c64-be8a-7cf879abef80.png)

2. When AAPS has set a 0 temp basal for 30 or 90 minutes it is applying the full negative IOB, even though a new rate is set after the next loop.  In this case, a 90 minute 0.00 basal was set at 2:14, the full 90 minutes of IOB was deducted.  At 2:19, 0.00 was set again for 90 minutes, and the full 90 minutes of IOB was deducted, even though those rates only actually ran for 5 minutes each.  Same occurred at 2:39AM with 30 minutes of IOB deducted even though it only ran for 5 minutes.
![image](https://user-images.githubusercontent.com/65743481/139303108-36559621-d0cf-4c2a-beb0-18f2125b5f3b.png)

jroth-haj:
@andyrozman do you need more logs?

andyrozman:
I think I know where the problem lies, so I am trying to fix it (TBR one), problem with boluses is little more global, and will require more investigation.
@siggy226 will you be able to test this, as soon as I am done with fix? Should be done by end of day on sunday.

@jroth-haj If you can help test this, this would be great... Are you experiencing the same problem?

andyrozman:
So both fixes are ready in my branch, if you guys could test it. I don't have SMB on my test device, so I couldn't test it myself... 
@siggy226  @jroth-haj 

https://github.com/andyrozman/AndroidAPS/tree/bug798_meal_tbr

siggy226:
@andyrozman I can build later today and test

siggy226:
@andyrozman SMBs look good.  Thank you!

Temp basals/IOB are still an issue (see 3:05 and 3:40 entries).  I tested by enacting some high temp targets to get a 0.00 u/hr rate, then cancelling the temp target to trigger loop to restart an actual rate.  Profile basal at this time is 1.3 u/hr.  Manually refreshing pump history a few times didn't make any changes.
[...]

PR #929, opened 2021-11-14
Bug798 - smb to meal bolus and zero tbr termination

andyrozman:
Fixed:
- Problem with SMB boluses, becoming Meal boluses after history is read
- TBR: When Zero TBR is running it was not terminated correctly (problem with history)


MilosKozak:
#798

PR #965, opened 2021-11-21
Add strings for Command Queue translation

Philoul:


MilosKozak:
tests needs to be fixed (mock new strings)

Philoul:
After several hours 🤯 I finally find fixes for tests.
Not sure it's the best solution even if tests pass.
Please tell me if it's ok or if there is a better solution.

app/src/test/java/info/nightscout/androidaps/queue/QueueThreadTest.kt
```diff
@@ -41,6 +43,7 @@ class QueueThreadTest : TestBaseWithProfile() {
             }
             if (it is CommandTempBasalAbsolute) {
                 it.activePlugin = activePlugin
+                it.rh = rh
```
MilosKozak, line 46: this should be covered by `if (it is Command)` or?

Philoul:
Oups, Crash detected in CommandTempBasalPercent (strange, it seems to be only during Disconnect Pump)

Issue #977, opened 2021-11-22
[Enhancement] Add option of correction by % in Bolus Wizard


yossibivas11:
I suggest adding a correction by % in addition to the correction by U in the Bolus Wizard/ Calculator.

Philoul:
Included in #300 
See also this issue and associated comments in previous repo MilosKozak/AndroidAPS#2462

yossibivas11:
Looks like this won't happen anytime soon then

Issue #300, opened 2021-02-10
Bolus percentage improvements in bolus wizard, and quickwizard


szantos:
![Screenshot_20210210-135629_5](https://user-images.githubusercontent.com/64865503/107519020-00fe1600-6bb0-11eb-8eb3-199d79a1f5ba.jpg)

1) The units from "Deliver this part of bolus wizard result" should be considered before the Corr
   (it would be easier to manage the correction)

2) "Deliver this part of bolus wizard result" should have an own row within the details with the corresponding units on the right (maybe with a fine line above - meaning, that all above has been considered)

3) optionally with a checkbox

4) but more important with an input field for overriding the standard value
   example use cases:
   - if there is more active insulin on meal start, then you have to prebolus less
   - if you know, you'll eat faster acting carbs, you have to prebolus more, maybe 100%

5) in the row with the "Result:" 60% could be on the left, and Result: 1,30U on the right
   (stating, that 60% has already been considered)

6) Addition of Carb duration field under Carb

7) Moving Corr input field to the very bottom, or under Carb time
   (that the 3 Carb input fields are beside each other)

8) Ability to consider custom bolus percentage for each quickwizard button (either before creating the button for the button itself, or after pressing the quickwizard button on the approval window)

Thanks & feel free to discuss :)

-------------------------------

This is how we do it right now:
- input all carbs to bolus wizard to get 100% bolus (with every checkbox except TT checked)
- cancel
- calculate (in the head) the exact pre-bolus considering a custom %
- delivering pre-bolus with the "Insulin" button
- waiting for delivery
- entering extended Carbs with the "Carbs" button
All in all: you have to use 3 buttons (Bolus Wiz, Insulin, Carbs); calculate in head; wait for insulin delivery

MilosKozak:
wizard is the most used dialog and keeping it small and simple saves time to everyone. adding more controls for rare cases goes directly against it.
So sorry but no for now

szantos:
@MilosKozak 
Thanks for considering.
1,2,5,8,9,10 are not about adding anything new. They would help the understanding, or would put some things in the right place (considering corr after %; no superbolus if %).

Philoul:
@szantos FYI You can take a look on this proposal close to some of yours and associated comments... https://github.com/MilosKozak/AndroidAPS/issues/2462

Philoul:
[...]

Issue #976, opened 2021-11-22
[Dash] Bolus registred under treatments but not delivered


amdaini:
We had a serious issue this morning; my daughter bolused 2.80u for breakfast, one hour later I got a call from the school nurse as her bg was unexpectedly raising. I checked on the  nightscout and there was the 2.8u registered, also checked on the treatment tab of his AAPS and the 2.8u was there but on the pod history strangely I was unable to find the breakfast bolus. 
AAPS version: 2.8.2.10-dev
Pump: Omnipod Dash 
CGM: Dexcom G6/xdrip+
PFA to this issue the AAPS logs and some screenshots of the treatments tab and pod history:

[AndroidAPS._2021-11-22_00-00-02_.2.zip](https://github.com/nightscout/AndroidAPS/files/7584613/AndroidAPS._2021-11-22_00-00-02_.2.zip)

![Uploading 20211122_130636.jpg…]()
![Uploading 20211122_130738.jpg…]()
![Uploading 20211122_130834.jpg…]()

avereha:
This is **not** dash specific.
It is also not a bug, but maybe we can improve the 'Treatments' button.

After reading the logs, it looks like the entry was added via the 'Treatments' button and the 'Do not bolus, record only' option was checked:
```
08:07:44.204 [RxCachedThreadScheduler-68] D/CORE: [UserEntryLogger$log$2.invoke():44]: USER ENTRY: TREATMENT TreatmentDialog Enregistrement [SimpleString(value=Record), Insu
lin(value=2.8), Gram(value=28)]
08:07:44.238 [RxCachedThreadScheduler-60] D/DATABASE: [TreatmentDialog.submit$lambda-13$lambda-12$lambda-5():160]: Inserted bolus Bolus(id=557, version=0, dateCreated=1637564864207, isValid=true, referenceId=null, interfaceIDs_backing=null, timestamp=1637564864192, utcOffset=3600000, amount=2.8, type=NORMAL, isBasalInsulin=false, insulinConfiguration=null)
08:07:44.241 [RxCachedThreadScheduler-35] D/DATABASE: [TreatmentDialog.submit$lambda-13$lambda-12$lambda-8():166]: Inserted carbs Carbs(id=260, version=0, dateCreated=1637564864232, isValid=true, referenceId=null, interfaceIDs_backing=null, timestamp=1637564864192, utcOffset=3600000, duration=0, amount=28.0)

```

amdaini:
[...]

PR #991, opened 2021-11-25
Add Possibility to adjust Correction in Percent in Wizard

Philoul:
This possibility is regularly requested in issues (https://github.com/MilosKozak/AndroidAPS/issues/2462, #300, #977), what do you think of this proposal...

The Idea is to only add a CheckBox in current WizardDialog to select Units for correction (U or %) and to have an automatic convertion between both:

Current correction in Unit : CalculatedBolus = ( ∑ InsulinFromXxx + insulinFromCorr ) * PercentageFromSettings
Additional correction in % : CalculatedBolus = ( ∑ InsulinFromXxx ) * PercentageFromWizard
- Default value for PercentageFromWizard = PercentageFromSettings

When we switch from U to % (or % to U), CalculatedBolus remains identical, you immediatly see the global correspondence between both units see screenshots below:
- Deliver this part of bolus is set to 80% in my preferences for screenshots below (it works fine for standard days but before sport training, period of resistance, I generally adjust percentage and I think it's more complicated with Units, but for other reasons Units are easier than percentage to make corrections...)
- When I enter a correction in one unit (U or %), it is automatically converted in the other unit if I check or uncheck the box, then I can adjust value in selected unit
- Min/Max for percent is set to 10% to 250% (hard coded but these values can be added in Constant file and of course be changed)
- InsulinFromCorrection is set to 0.0 if percentage unit is selected
- Last selection of units (% or U) is memorized
- default value for percentage is the value set in preferences

image

Default values when you open wizard (according to last selection)
image


Andries-Smit:
For users that will not use this feature, will the checkbox be hidden, just like the blue text? (Preference: General > Overview > Advanced > Deliver this part of the bolus wizard result [%])

Philoul:
We can add some new parameters in Advanced settings of Overview but @MilosKozak doesn't want too many settings to not fall into "xDrip syndrome", so I prefer other solutions without additional settings...

The other idea would be to propose a short click on wizard button to have a very simplifed User Interface, and a long click with full interface...

Something like that with a Short Click (Photoshop coding, sorry for french interface...)
image

We can also memorize last selection of Advanced settings to show Simplified or advanced by default according to user selection on short click, same as View detailled information...

Philoul:
@Andries-Smit 
> For users that will not use this feature, will the checkbox be hidden, just like the blue text? (Preference: General > Overview > Advanced > Deliver this part of the bolus wizard result [%])

You are right, if percentage is not set in preference We don't see le blue box it in WizardDialog, So I have to fix that

avereha:
Something that we learned in https://github.com/nightscout/AndroidAPS/issues/976 is that if we give people too many (small)checkboxes, they are eventually going to click on them, even if by mistake.

If a user decides to use this feature, are they always going to use percentage correction(or absolute values) or they will need the ability to switch between percentage and absolute values on each action? If they always want to use one way or another for the correction, maybe a setting would simplify this dialogue a bit.

While I don't know exactly how we could make this better, I wanted to share that adding more options to a dialogue that already has a lot of options and is used often can lead to mistakes.

Philoul:
> If a user decides to use this feature, are they always going to use percentage correction(or absolute values) or they will need the ability to switch between percentage and absolute values on each action? If they always want to use one way or another for the correction, maybe a setting would simplify this dialogue a bit.

I don't know, some users prefer Unit, but the question to also have % was asked several times

> Something that we learned in #976 is that if we give people too many (small)checkboxes, they are eventually going to click on them, even if by mistake.

I think there are no risk of mistake, because when you click on it, it doesn't change the result of calculated bolus, it only change and convert the unit of correction.
- I set 80% in my settings, but with Unit, I never know when I adjust Correction (especially positive value), if I'm still below 100% or above, I find usefull to immediatly check (or make adjustment quicker) to switch from U to % or on the other direction

> While I don't know exactly how we could make this better, I wanted to share that adding more options to a dialogue that already has a lot of options and is used often can lead to mistakes.

In wizard, I never use several fields : BG value (it's in overview in background, Profile selection),
My photoshop picture (before working on development) was proposals to simplify User Interface (without any additional settings...)

PR #1037, opened 2021-12-05
Allow/reenable the setting of TempTargets while pump is suspended or …

swissalpine:
…disconnected

Unfortunately, it is currently no longer possible to set a TempTarget when the pump is disconnected or suspended (TBR: 0%). This seems logical at first sight (why a TempTarget when the pump is not connected or the loop suspended!), but prevents many useful possibilities, namely when the TempTarget should be set immediately after the disconnection or suspension has expired.
An example of this are f. e. sport activities: To reduce IOB, I set a 0% TBR half an hour before sport for one hour with the function "pump disconnected", so that the associated 0% TBR continues for a while while sport starts. I also set a three hour Activity TempTarget. This continues to be active when the pump loops again with the pump disconnected time expired.
The fix allows/reenables setting a TempTarget even if the pump is suspended or disconnected.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
@@ -361,13 +361,10 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
                     activity,
                     ProtectionCheck.Protection.BOLUS,
                     UIRunnable { if (isAdded) CarbsDialog().show(childFragmentManager, "Overview") })
-                R.id.temp_target         ->
-                    if (loop.isDisconnected) OKDialog.show(activity, rh.gs(R.string.not_available_full), rh.gs(R.string.smscommunicator_pumpdisconnected))
-                    else
-                        protectionCheck.queryProtection(
-                            activity,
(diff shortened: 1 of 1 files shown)


MilosKozak:
well I think your problem is a result of misusing "Pump disconnection" feature. 
I'm still fan of keeping it as it is now because you have more options to resolve your issue (for exaple set TT before TBR 0% or instead of disconnection set very low profile) @AdrianLxM

AdrianLxM:
TT also have the functionality to prevent SMB.
So setting them while disconnected would be a good feature for a scenario like a shower after sports. There you might want to not have them for say one or two hours... But afterwards enable them again automatically.

Enabling TT while disconnected would be straight forward imho. Adding something like temporary preferences to disable SMB for an hour after sports seems overkill for something we could solve with TT. 

But maybe there are other solutions?

AdrianLxM:
An addition to "set very low profile": Wouldn't that cause other issues like if you run away from your phone or it runs out of battery? The pump would have the low profile run for a much longer time than it should?

MilosKozak:
if you set it with duration (like you set disconnection with duration) I don't see huge difference

AdrianLxM (approved):
Looks good to me. I like it
👍

swissalpine:
Thanks for your support (@AdrianLxM) and merging (@MilosKozak)!

PR #1048, opened 2021-12-07
Add Possibility to adjust Correction in Percent in Wizard

Philoul:
**Sorry I made a wrong synchro on PR #991 so I closed it and opened a new PR with a clean branch**

This possibility is regularly requested in issues (https://github.com/MilosKozak/AndroidAPS/issues/2462, #300, #977), what do you think of this proposal...

The Idea is to only add a CheckBox in current WizardDialog to select Units for correction (U or %) and to have an automatic convertion between both:

Current correction in Unit : CalculatedBolus = ( ∑ InsulinFromXxx + insulinFromCorr ) * PercentageFromSettings
Additional correction in % : CalculatedBolus = ( ∑ InsulinFromXxx ) * PercentageFromWizard
- Default value for PercentageFromWizard = PercentageFromSettings

When we switch from U to % (or % to U), CalculatedBolus remains identical, you immediatly see the global correspondence between both units see screenshots below:
- Deliver this part of bolus is set to 80% in my preferences for screenshots below (it works fine for standard days but before sport training, period of resistance, I generally adjust percentage and I think it's more complicated with Units, but for other reasons Units are easier than percentage to make corrections...)
- When I enter a correction in one unit (U or %), it is automatically converted in the other unit if I check or uncheck the box, then I can adjust value in selected unit
- Min/Max for percent is set to 10% to 250% (hard coded but these values can be added in Constant file and of course be changed)
- InsulinFromCorrection is set to 0.0 if percentage unit is selected
- Last selection of units (% or U) is memorized
- default value for percentage is the value set in preferences

image

Default values when you open wizard (according to last selection)
image


swissalpine:
I have tested it: It's really great!
A small note: If I activate percent and then deactivate again or vice versa, then the percentage or the amount of insulin changes slightly (1-2%) after the first back conversion. This is certainly not a big problem, but also not quite optimal and could irritate users - it is probably due to rounding errors ...

Philoul:
> A small note: If I activate percent and then deactivate again or vice versa, then the percentage or the amount of insulin changes slightly (1-2%) after the first back conversion.

It's because each time you click on button the calculation is updated in realtime (slight modif of iob between 2 calculations).
Percentage used for convertion is not rounded (double value used and not int) to keep exact value in Units, so if you quickly switch between first calculation and converted one, you get exactly same value because not enough time to have modification of iob, but if you wait only a few seconds (for ex 15/20s) between first calculation and convertion, then you can get slight modification.

swissalpine:
No: I can reproduce this several times again!
I enter 90% and 35g, uncheck %, getCorr 0.4 IE, check %, get 91%
Change this back to 90%, uncheck %, get Corr 0.4 IE, check %, get 91%
I've done this 6 times in a row ...

swissalpine:
Here is a similar video with the reproducable change from 90% to 89%:

https://user-images.githubusercontent.com/12517693/145419478-455734aa-f7b9-4169-823b-43f59a62832f.mp4

Philoul:
Ok, I'll check it (my min increment for my pump is 0.01U, and it seems that for you it's only 0.1U)
On my tests percentage never changed only Units of insulin was slightly modified during convertion.
In your video it's Unit that remains unchanged and percentage that is modified during convertion.
Thanks

swissalpine:
Thanks. My pump is a spirit combo and the combo has only 0.1 increment.

Philoul:
@swissalpine Can you test again with my latest commit (I removed rounding of calculated correction so now it should be fixed)

swissalpine:
@Philoul The rounding problems are fixed! Thank you! I'm happy to have this feature!

PR #1080, opened 2021-12-16
Fix dialogs and remove extra CONFIRM. Fixes #1078

### LINKED ISSUE #1078: Wear OS confirmation screen is blank
Version: 3.0-beta11 for both phone & watch

Steps to reproduce

1. Go into app on wear os
2. Select Wizard/eCarb/Bolus/TempT
3. Select any inputs in the steps and confirm
4. There is a confirmation coming back from the phone
5. Expected result: Confirmation shows info and allows the user to confirm the action
6. Actual result: The confirmation screen shows "Title" at the top and a blank/black screen without the option to confirm anything

substars:
Fixes #1078 

Reproduced and tested on a virtual Wear OS Round API 30 and my Galaxy Watch 4.

I also saw that the confirm dialogs said CONFIRM twice, removed one so we can see more of the sometimes very long confirmation message.

Wizard confirmation on current dev:
Screenshot_1639637001

Same confirmation after removing the extra `setContentView()`:

Screenshot_1639635309

Same confirmation after removing title from the layout:

Screenshot_1639636403

If this is too much change for 3.0, there is a one-line fix that I'll call out in the comments.

Thank you for this amazing project!

### DIFF
--- a/wear/src/main/java/info/nightscout/androidaps/interaction/actions/AcceptActivity.java
+++ b/wear/src/main/java/info/nightscout/androidaps/interaction/actions/AcceptActivity.java
@@ -21,7 +21,6 @@ import info.nightscout.androidaps.data.ListenerService;
 
 public class AcceptActivity extends ViewSelectorActivity {
 
-    String title = "";
     String message = "";
     String actionstring = "";
     private DismissThread dismissThread;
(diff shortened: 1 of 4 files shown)


wear/src/main/java/info/nightscout/androidaps/interaction/actions/AcceptActivity.java
```diff
@@ -43,7 +41,6 @@ protected void onCreate(Bundle savedInstanceState) {
             return;
         }
 
-        setContentView(R.layout.grid_layout);
```
substars, line 46: Removing this line fixes #1078

PR #1082, opened 2021-12-16
Wear: add feature: simplified watch face during charging

Andries-Smit:
During the night, while charging, it would be helpful, if the display could stay “always on” and show my blood glucose. The Android developer option “Stay awake when charging” can be used to do so. However,  watch-faces are a bit too bright, has too much information, and the details are hard to read with sleepy eyes. Therefore I added an option for the watch-face, to simplify the UI only during charging when set in the configuration.
Implemented in all watch faces that inherit WatchFaceBase: Home, Home2, Home Big, Digital, Steampunk and cockpit.

(Includes a fix for the trend arrows showing as a blue emoticon)

image

image

image

image

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -100,6 +100,9 @@
             <intent-filter>
                 <action android:name="android.service.wallpaper.WallpaperService" />
                 <category android:name="com.google.android.wearable.watchface.category.WATCH_FACE" />
+
+                <action android:name="android.intent.action.ACTION_POWER_CONNECTED" />
+                <action android:name="android.intent.action.ACTION_POWER_DISCONNECTED" />
             </intent-filter>
         </service>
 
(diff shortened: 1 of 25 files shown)


dlvoy:
I've tested it and it switch to charging mode on next data resent. While for changing from standard to charging mode it is fine, i got confused why watchface do not get back to normal when i took the watch from charger back on the wrist. "Send all data" option of wear plugin helps in that case to force face update.

@Andries-Smit can you add broadcast receiver to listen for charging status change and force update? 
https://developer.android.com/training/monitoring-device-state/battery-monitoring
https://stackoverflow.com/questions/50912439/wearos-how-to-check-is-battery-is-plugged

Andries-Smit:
@dlvoy yes, that might be a good idea. let me check that

Andries-Smit:
@dlvoy implemented a battery receiver, so now the updates are instant

dlvoy:
Retested, works like a charm!

rICTx-T1D:
I like the feature-request, but why not include in global namespace for all watchfaces?

StephenBrown2:
How can I get this on my watch? I've tried Wear Installer, and it says "Wear component not found" when I select AndroidAPS.

Andries-Smit:
> I like the feature-request, but why not include in global namespace for all watchfaces?

For no specific reason, other than each watch face has his own layout. And i implemented it on my favorite first. Each layout requires its own implementation for now. I can have a look if it can be implemented for other watchfaces too.

Andries-Smit:
@rICTx-T1D The feature is not implemented in watch faces that inherit WatchFaceBase: Home, Home2, Home Big, Digital, Steampunk and cockpit. 
@dlvoy could you please check the other watch faces too? Thanks

PR #1094, opened 2021-12-24
Fix mmol Units for Therapy Event in NS

### LINKED ISSUE #1085: [3.0-beta9] NSClient - Incorrect vertical position of BG check on graph
Hi, this is probably only displaying issue on NS client side, details in screenshots.

Here is bg check of 5.8 correctly displayed on aaps graph, also entry in treatments shows value corrrectly. (Sorry for non-english screenshots)
AndroidAPS_01

AndroidAPS_02

But on NS Client side the value is positioned on the bottom of the graph, not on correct y-axis (vertical) position.

NSClient_01

And also in entries screen the value is not shown in the BG check entry.

NSClient_02

Philoul:
Fix #1085
But it's a workaround (I suppose root cause is a bug in NS or somewhere else in AAPS but I can't find it)
=> See informations below to help fix root cause (I don't like this kind of workaround...)

When BG value (`glucose`) associated to a therapyEvent is in mmol unit, there are 2 additional entries in json string (`targetTop` and `targetBottom`).
These 2 entries are null but when added, glucose units (`mmol`) is replaced by `mg\/dl` in json string so units is wrong for `glucose` value in received json string.

json string received for BG Check if units is mgdl:
```
{
	"_id":"61c599ee1d52ca0004ae55f7",
	"eventType":"BG Check",
	"isValid":true,
	"created_at":"2021-12-24T09:58:49.000Z",
	"enteredBy":"AndroidAPS",
	"units":"mg\/dl",
	"notes":"test bg",
	"glucose":144,
	"glucoseType":"Sensor",
	"mills":1640339929000,
	"mgdl":144
}
```
json string received for BG Check if units is mmol:
```
{
	"_id":"61c599251d52ca0004ae55f2",
	"eventType":"BG Check",
	"isValid":true,
	"created_at":"2021-12-24T09:55:28.000Z",
	"enteredBy":"AndroidAPS",
	"units":"mg\/dl",
	"notes":"test mmol",
	"glucose":8.055555555555555,
	"glucoseType":"Sensor",
	"mills":1640339728000,
	"mmol":8.055555555555555,
	"targetTop":null,
	"targetBottom":null
}
```

### DIFF
--- a/core/src/main/java/info/nightscout/androidaps/extensions/TherapyEventExtension.kt
+++ b/core/src/main/java/info/nightscout/androidaps/extensions/TherapyEventExtension.kt
@@ -69,12 +69,18 @@ fun therapyEventFromNsIdForInvalidating(nsId: String): TherapyEvent =
     )!!
 
 fun therapyEventFromJson(jsonObject: JSONObject): TherapyEvent? {
-    val glucoseUnit = if (JsonHelper.safeGetString(jsonObject, "units", Constants.MGDL) == Constants.MGDL) TherapyEvent.GlucoseUnit.MGDL else TherapyEvent.GlucoseUnit.MMOL
     val timestamp = JsonHelper.safeGetLongAllowNull(jsonObject, "mills", null) ?: return null
     val type = TherapyEvent.Type.fromString(JsonHelper.safeGetString(jsonObject, "eventType", TherapyEvent.Type.NONE.text))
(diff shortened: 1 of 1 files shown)


MilosKozak:
I analyzed the problem and it is in NS. I reported it to @sulkaharo .
We should wait for NS patch instead of making workarounds.
Thanks!

PR #1100, opened 2021-12-27
confirm dialog scroll + title fixes

substars:
#1080 changed AcceptActivity to always have `CONFIRM` as its top title (with no secondary title), but I later realized that pump and loop status use AcceptActivity and had a secondary title that no longer showed up -- so after that PR they just said `CONFIRM` on top.

I still like a single title better since there is not much room on the screen for text, but this restores the  `title` Intent Bundle attribute and will use it to set the top title if it is supplied. Loop status, for example, will now set the title to `STATUS LOOP` (which I think used to be the secondary title). I tried to look through the code to see if the secondary title needed the primary title for context and I can't find an example of that.

I also made the message text for all confirm dialog screens scrollable using the rotating bezel since it already worked on the plus/minus screen. Feels very nice on a Galaxy Watch 4 Classic.

I have been using this branch build on my live rig for the last few days and haven't had any issues. I'm trying not to look at my phone so much 🙈

### DIFF
--- a/wear/src/main/java/info/nightscout/androidaps/interaction/actions/ViewSelectorActivity.java
+++ b/wear/src/main/java/info/nightscout/androidaps/interaction/actions/ViewSelectorActivity.java
@@ -29,7 +29,7 @@ public class ViewSelectorActivity extends Activity {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.grid_layout);
 
-        setTitleBasedOnScreenShape(String.valueOf(getTitle()));
+        setTitleBasedOnScreenShape();
 
         pager = findViewById(R.id.pager);
         DotsPageIndicator dotsPageIndicator = findViewById(R.id.page_indicator);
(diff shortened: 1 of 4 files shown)


wear/src/main/java/info/nightscout/androidaps/interaction/actions/ViewSelectorActivity.java
```diff
@@ -59,7 +59,12 @@ public void setAdapter(GridPagerAdapter adapter) {
         pager.setAdapter(adapter);
     }
 
-    private void setTitleBasedOnScreenShape(String title) {
+    private void setTitleBasedOnScreenShape() {
+        // intents can inject dynamic titles, otherwise we'll use the default
+        String title = String.valueOf(getTitle());
+        if (getIntent().getExtras() != null)
+            title = getIntent().getExtras().getString("title", title);
```
Andries-Smit, line 66: Prefer if statement without prentices, please use a single line, or add the conditional section {}

wear/src/main/res/layout/action_confirm_text.xml
```diff
@@ -11,6 +11,9 @@
         app:boxedEdges="all">
```
Andries-Smit, line 11: Let's make the padding smaller to android:padding="5dp" , this will make the text much easier to read on round devices

Andries-Smit:
@substars works like a charm! awesome!

Issue #1105, opened 2021-12-29
[enhancement] - calculate and record carbs based on food inputs


erobinson:
Would like to be able to:

- Select a food from the FOOD tab and have it open the bolus wizard/calculator w/ the correct number of carbs
- Include fat & protein into carb calculations
- Record when specific foods were eaten (and then be able to see that in a history browser and BG levels after the meal)
- Have a meal calculator where multiple foods & portions can be added/adjusted and then the meal can record carbs or start the bolus wizrd/calculator
- nice to have - default the list of foods to foods frequently consumed at that time of day (i.e. default to show breakfast foods in the morning)

erobinson:
Initial concept:
![image](https://user-images.githubusercontent.com/1564297/147697465-437c6873-6a44-46bc-b241-6dd07d65935a.png)

swissalpine:
The ideas are great and the design looks good! I would use it!

pnharris02:
I like this idea of using the foods database to do bolus calculations.  I discussed something similar in one of the forums several months ago.  
Note that there is a relatively new ticket https://github.com/nightscout/AndroidAPS/issues/3761 that talks about entering carbs/100g directly from the bolus calculator.

PR #1106, opened 2021-12-29
Added the ability to trigger the calculator from the food tab

erobinson:
Added the ability to bring up the bolus calculator from the food tab. This would allow users to quickly bolus for frequently consumed foods/meals. Working on making a more robust meal calculator where you can add foods & portions, etc. Wanted to create an initial PR for the small changes to get feedback and add initial enhancements.

Screenshot_20211228-224253

Screenshot_20211228-224325

https://github.com/nightscout/AndroidAPS/issues/1105

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/MyPreferenceFragment.kt
@@ -28,6 +28,7 @@ import info.nightscout.androidaps.plugins.bus.RxBus
 import info.nightscout.androidaps.plugins.configBuilder.PluginStore
 import info.nightscout.androidaps.plugins.constraints.safety.SafetyPlugin
 import info.nightscout.androidaps.plugins.general.automation.AutomationPlugin
+import info.nightscout.androidaps.plugins.general.food.FoodPlugin
 import info.nightscout.androidaps.plugins.general.maintenance.MaintenancePlugin
 import info.nightscout.androidaps.plugins.general.nsclient.NSClientPlugin
 import info.nightscout.androidaps.plugins.general.nsclient.data.NSSettingsStatus
@@ -96,6 +97,7 @@ class MyPreferenceFragment : PreferenceFragmentCompat(), OnSharedPreferenceChang
     @Inject lateinit var statusLinePlugin: StatusLinePlugin
     @Inject lateinit var tidepoolPlugin: TidepoolPlugin
     @Inject lateinit var virtualPumpPlugin: VirtualPumpPlugin
+    @Inject lateinit var foodPlugin: FoodPlugin
     @Inject lateinit var wearPlugin: WearPlugin
     @Inject lateinit var maintenancePlugin: MaintenancePlugin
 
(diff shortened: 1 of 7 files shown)


MilosKozak:
code quality is very good. What I'm not sure about is that fat and protein is usually treated as eCarbs and not accounted into a bolus @AdrianLxM

MilosKozak:
and for easy usability we should consider adding # of portions next to calc icon

AdrianLxM:
I do like the idea of handling fat/protein but I think we should discuss it a bit more into detail how we should handle it.

I also like the idea of having the calculator directly accessible. 👍 
Could you maybe split this into two MRs, in case out discussion how to handle fat/protein takes longer?

erobinson:
The thought for the calculator there was, for snacks rather than for meals, where you can quickly bolus. I.e. bolus for a banana or protein bar are you head out the door in the morning. So wanted to keep it quick and simple. Then to build a more robust meal process where you can add foods/portions/bolus/ecarbs/etc. For a meal.

I could add another preference to allow the user to choose whether or not to include fat/protein in the calculator shortcut and have it default to false. Then users can choose to turn it on. 

Otherwise, I can just split that part out and can discuss further.

AdrianLxM:
The caveat I see with this calculation of fat/protein is, that it would re-define `I:C` ratio as many understand it (have tested it) and as their diabetes team would see it.

It would work if we define the `I:C` as _**pure carb factor**_. Meaning that the factor was tested on a meal only consisting of carbs or was calculated back to such a factor by subtracting the insulin needed for fat/protein when testing it.

I fear that in reality people will test their factors with say _toast with a bit of cream cheese_ which would mean their `I:C` would already include some insulin for fat/protein.
Couldn't this lead to too much insulin if we now add some?
With a closed loop system I'd rather err on the low side of a bolus as the loop itself can take care of some of the fat/protein.

I personally therefore calculate with "excess fat/protein". So fat/protein exceeding my standard mixture.

I do have two proposals up for discussion - maybe there are better solutions though.
1) Find a way to calculate the "excess fat/protein" by adding settings for the default mixture?
2) Setting the default percentage for fat/protein to 0% as it's already covered by `I:C` tested on a food mixture.

To me 2) would feel like a bad workaround though as it wouldn't scale for higher amounts of fat. On the other hand, if people would go with a _**pure carb factor**_ they would need to add fat/protein always - even with their default mixture. So I'm leaning towards 1).

AdrianLxM:
> @MilosKozak: What I'm not sure about is that fat and protein is usually treated as eCarbs and not accounted into a bolus 

If you have mostly fat/protein, a direct bolus might be "too fast". If there still is a larger amount of carbs it is fine in my experience to add it directly to the bolus. Even modern insulins have quite some "rat's tail" that would cover the slower food parts. In energy (kCal) I don't use eCarbs up until 2:1 (fat/protein : carbs).

AdrianLxM:
@erobinson are you in our discord channel? Can we do some brainstorming there?

erobinson:
Will do another PR with just the calculator shortcut & w/o the fat/protein included in the carb count.

PR #1119, opened 2022-01-02
Seconds to Int for AlarmClock Intent

AdrianLxM:
The expected datatype for the `AlarmClock.EXTRA_LENGTH` extra is `Integer`.

As we usually use Long to represent time, I'd convert it directly before passing it to the Intent.
In the case that the alarm is extremely in the future (technically possible with `Long`, practically not), it is capped to `Integer.MAX_VALUE`


MilosKozak:
sorry i missed this PR

Issue #1122, opened 2022-01-03
[V3 Beta 11] Wear OS devices SDK < 25 are no longer working


Andries-Smit:
Dear team,

Older Wear OS devices SDK < 25 are no longer working. This is issue occurred after commit https://github.com/nightscout/AndroidAPS/commit/17b4d6e7ed14a0f602678251ee7d8ace162f06fe. The commit introduced decency injection for Wear OS. 

And on start up cause the following Error

> 2021-12-16 20:08:32.947 3293-3293/info.nightscout.androidaps W/System: ClassLoader referenced unknown path: /data/app/info.nightscout.androidaps-1/lib/x86
> 2021-12-16 20:08:32.951 1313-1388/? D/gralloc_ranchu: gralloc_alloc: format 1 and usage 0x933 imply creation of host color buffer
> 2021-12-16 20:08:32.957 3293-3293/info.nightscout.androidaps D/AndroidRuntime: Shutting down VM
>     
>     
>     --------- beginning of crash
> 2021-12-16 20:08:32.957 3293-3293/info.nightscout.androidaps E/AndroidRuntime: FATAL EXCEPTION: main
>     Process: info.nightscout.androidaps, PID: 3293
>     java.lang.NoClassDefFoundError: Failed resolution of: Landroid/net/wifi/hotspot2/pps/HomeSp;
>         at info.nightscout.androidaps.di.DaggerAppComponent.mapOfClassOfAndProviderOfAndroidInjectorFactoryOf(DaggerAppComponent.java:119)
>         at info.nightscout.androidaps.di.DaggerAppComponent.dispatchingAndroidInjectorOfObject(DaggerAppComponent.java:123)
>         at info.nightscout.androidaps.di.DaggerAppComponent.injectAaps(DaggerAppComponent.java:232)
>         at info.nightscout.androidaps.di.DaggerAppComponent.inject(DaggerAppComponent.java:228)
>         at info.nightscout.androidaps.di.DaggerAppComponent.inject(DaggerAppComponent.java:51)
>         at dagger.android.DaggerApplication.injectIfNecessary(DaggerApplication.java:63)
>         at dagger.android.DaggerApplication.onCreate(DaggerApplication.java:38)
>         at info.nightscout.androidaps.Aaps.onCreate(Aaps.kt:15)
>         at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1024)
[...]

PR #1127, opened 2022-01-04
Wear: fix import error api 25

Andries-Smit:
Fix error #1122 to support API 25


AdrianLxM:
LGTM :+1: like discussen on Slack

Issue #905, opened 2021-11-11
Feature request : button to "clean" main graph


robertrub:
Sometimes, the main graph is so filled with different informations that the BG curve is not easy to see.

If possible, it would be nice to have a _tick box_ that toggles between main BG view with carbs and insulin vs the actual graph with all the information on it.

![Screenshot_20211109-074035_AndroidAPS](https://user-images.githubusercontent.com/76626474/141339110-3d44186f-dae6-40ac-8d64-a6b3af4e9a4c.jpg)

robertrub:
Hi @MilosKozak 

@Philoul did a pull request some time ago that "solves" this issue. Could you please have a look and validate it (if you're OK with it) ?

https://github.com/nightscout/AndroidAPS/pull/1142

Tnx

Robby

robertrub:
If possible, 8r would be nice to have "cleaning" levels... 1 / clean orange messages  2/ clean also profile changes

Having the checkbox **on the graph** is much more practical.

PR #1142, opened 2022-01-09
Add a Checkbox to simplify main graph (hide TherapyEvents from graph)

Philoul:
What do you think of this proposal (see #905)
Only therapyEvents are hidden if checkbox unchecked... (I think it's important to keep all other events)
image

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
@@ -186,6 +186,14 @@ class HistoryBrowseActivity : NoSplashAppCompatActivity() {
             ).show()
         }
 
+        binding.simplifyCheckbox.isChecked = sp.getBoolean(R.string.key_therapy_events_visible, true)
+        binding.simplifyCheckbox.setOnCheckedChangeListener { _, isChecked ->
+            run {
+                sp.putBoolean(rh.gs(R.string.key_therapy_events_visible), isChecked)
+                updateGUI("SimplifyCheckbox Change")
+            }
+        }
+
         val dm = DisplayMetrics()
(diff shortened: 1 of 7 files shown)


vanelsberg:
Yes. Think this is a good idea. Would love to see that option in.
Could it maybe also hide the profile changes? There are times I want to see them, but most of the time for me they are cluttering the view.

Philoul:
@vanelsberg hiding Profile Switch at the same time is easy to do, but I'm not convinced it's a good idea:
-> Profile Switches should be quite rare in every looper's time line (temporary PS for exercises or for special events and one-time PS for updating profile) and Selected Profile at any time is one of the most important information concerning user's loop (all main parameters are defined in your profile)
Can you explain to me why you have so many PS that your main graph becomes unreadable?
@MilosKozak , do you have any comments concerning this PR (with or without @vanelsberg's additional request)?

vanelsberg:
@Philoul In case of the screen print I had been experimenting/adhusting my profile.
In general I do use Profile switches:

- For special meal (e.g: pizza) - setting a 2 hour 130% and AAPS/SMB handles that perfectly.
- My insulin sensitivity changes a lot and unpredictably by day and even by day time (morning/midday/evening/night). So much that even oRef's autosense feature can not keep up. So when I have my "off-day" I use a switch profile to 120% for hours to keep on target. 

I know when I intentionally do a profile switch. So I have no need to see them in the overview all the time unless I am looking for an event or improvement. And then profile switches and other info are be far better viewable under treatments.

rustymonkey:
The Overview screen/graph is already quite 'busy'. This functionality is useful, though, so would it be feasible for it to instead appear as an option within the graph menu to keep the graph clean? e.g.
[x] Predictions
[x] Basals
[ ] Activity
[ ] Events <--------

MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/8fee28bcb7226aaae8b467391747bdd3e0ee9796

PR #1155, opened 2022-01-11
Swap cob and iob icons in wizard dialog

### LINKED ISSUE #1154: Re-order icons in Bolus Wizard
I think it would be nice if the icons in bolus wizard would follow the right order of the detailed list, the actual icons order is BG - TREND - INSULIN - CARBS. I think Insulin should swap with carbs.

photo 100 18-11-49

Philoul:
Fix #1154

### DIFF
--- a/app/src/main/res/layout/dialog_wizard.xml
+++ b/app/src/main/res/layout/dialog_wizard.xml
@@ -289,17 +289,6 @@
                 app:srcCompat="@drawable/ic_fortyfiveup" />
 
             <ImageView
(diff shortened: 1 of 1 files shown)


MilosKozak:
hmmm i thought IOB first is more logical

parapenT1sta:
> hmmm i thought IOB first is more logical

I made this suggestion just to follow the order from the detailed list. Just to be more precise.

MilosKozak:
if to change this I'd consider order in bottom part of dialog

PR #1163, opened 2022-01-12
Wizard Icons -> Allow both: Enable and Disable

swissalpine:
So far it is only possible (thanks to @Philooul) that a click on the icon disables the function. This PR also allows to enable bg, trend, carbs and iob if disabled while clicking on the icon.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/dialogs/WizardDialog.kt
+++ b/app/src/main/java/info/nightscout/androidaps/dialogs/WizardDialog.kt
@@ -175,10 +175,33 @@ class WizardDialog : DaggerDialogFragment() {
             }
             dismiss()
         }
-        binding.bgEnabledIcon.setOnClickListener { binding.bgCheckbox.isChecked = !binding.bgCheckbox.isChecked }
-        binding.trendEnabledIcon.setOnClickListener { binding.bgTrendCheckbox.isChecked = !binding.bgTrendCheckbox.isChecked }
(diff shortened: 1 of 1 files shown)


Philoul:
Not needed, my PR already allow Enable And Disable...

MilosKozak:
I agree with @Philoul . His code is more elegant

swissalpine:
Shure, I know the behaviour of the iob Check Box.
But I couldn't reactivate BG etc. after disabling it first. Would you see a screen video top beliebte me?

Philoul:
Yes, icons are small so I'm not surprised it could be difficult to click on them.
Can you send a screenshot of your small screen (to check if we can increase a bit size, margin or padding of these icons on low res screen...)

swissalpine:
Too late. I have long since adjusted the for me. But if you like, I restore an old version ...
Screenshot_20220114-062916
The size of the icons is 0.7, so you can roughly estimate that there is hardly any space.
"Cancel" is then abbreviated to "Canc".

Philoul:
We have a Low res skin in AAPS...
It's better if we use low res skin to adapt these kind of user interface that doesn't work on small phones..

swissalpine:
What are you trying to tell me: that I shouldn't change it like that for myself - or that I should do a PR with this adjustment?
There is much more changed here: e.g. fixed font sizes, but also the alpha values for the icons, because otherwise the icons are hardly to be seen ...

Philoul:
I think, each time you detect something that need improvements for low res screen, it's better if you open an issue with your proposal (screenshot before/after)
Some of your modifications can be included for all users that have small phones with "low res skin" selection in preferences...

PR #1160, opened 2022-01-12
Wear: tile support pump actions and temporary target

Andries-Smit:
Wear OS Tiles provide easy access to the information and actions users need in order to get things done.  Using the Tiles API requires targeting API level 26 or higher.

Tiles can be used to quickly access actions on the AAPS application, without going through the menu of the watch face.
The tiles are optional and can be added and configured with 1 to 4 actions by the user.  (Long press the tile, will show the configuration options)

Action tile can request the pump for: 
1. Calculation, based on carb input
1. Insulin 
1. Treatment: Insulin & Carbs 
1. Carbs
1. Temporary Target (target and time)
action_tile_preview

The Temp Target Tile can request a temporary target based on the phone app presents.
1. Activity
1. Hypo
1. Eating soon
1. Manual (target and time)
1. Cancel
image

Furthermore some small changes
1. Align action/menu naming with phone app (Bolus =>Treatment, eCarbs => Carbs, Wizard => Calculator)
1. Toast for shown after action request

**Compatibility note:**
- The watch-phone API is extended, making the change backward compatible. Watches that are still on 2.8.2.1 keep working
- Watch Minimum API is staying 23, Tested by @dlvoy on Sony SWR50
- Tiles can only be used watches with API level 26 or higher.

PR #1195, opened 2022-01-18
Wear: sync settings units and wizard percentage 

Andries-Smit:
When a user wants to control the AAPS pump through a watch, when using mmol/L, it needs to set the units on the watch configuration.
This is no longer needed as this can be done by synchronizing the settings from the phone to the watch. The watch setting is removed and now, the Units are configured on the phone preference "General > Units"

With this PR the `percentage` for the calculation wizard is also sent to the watch.
The percentage can be set on the phone preferences: "Overview > Advanced Settings > Deliver this part of bolus wizard result [%]" The default value is 100%

The PR is depending on #1160


vanelsberg:
Welcome additions to PR#1160.
**Tested and found no issues.**
Well done Andries!

(Wear app was tested for Samsung Watch 4 Classic)

PR #1196, opened 2022-01-18
Add menu for direct rescale of Main graph

Philoul:
The idea is to use existing button
- Short click to select subgraphs
- long click to show a rescale menu

image

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
@@ -198,7 +198,7 @@ class HistoryBrowseActivity : NoSplashAppCompatActivity() {
         binding.bgGraph.gridLabelRenderer?.reloadStyles()
         binding.bgGraph.gridLabelRenderer?.labelVerticalWidth = axisWidth
 
-        overviewMenus.setupChartMenu(binding.chartMenuButton)
+        overviewMenus.setupChartMenu(binding.chartMenuButton, overviewData, ::updateGUI)
         prepareGraphsIfNeeded(overviewMenus.setting.size)
         savedInstanceState?.let { bundle ->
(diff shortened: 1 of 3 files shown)


vanelsberg:
@Philoul Yes, I do think the direct  rescale menu would be an improvement 

Not sure about the short/long press.
Would not be affected by low responsiveness of rescaling?

Philoul:
This PR allows direct selection (it's another way to accelerate if you want to go back from 12h to 6h, you can select it without full "loop" 12=>18=>24=>6...
I tested short/long press and it works fine.
But when aaps is temporary frozen (for ex if you try to scale several times quickly or if you try to rescale graph during the countdown of new bg value, nothing works (short, long click or anything else).
You can test it and how it works on your phone.

vanelsberg:
Ok. Unless this is difficult to implement I would say: go for it and lets see how it works?
Willing to test if that helps!

Philoul:
PR Updated (rescale menu included in HistoryBrowserActivity and OverviewFragment)

Philoul:
@Andries-Smit and @osodebailar , can you tell me if you think this PR is a good idea? (for UX and concerning Material compatibility for menu...)

osodebailar:
> @Andries-Smit and @osodebailar , can you tell me if you think this PR is a good idea? (for UX and concerning Material compatibility for menu...)

The idea is good in my opinion.
In addition googles designer guide says if  we had an intuitive  function , like swippping for example, its a good idea
to have the same function in a menu.

Andries-Smit:
> @Andries-Smit and @osodebailar , can you tell me if you think this PR is a good idea? (for UX and concerning Material compatibility for menu...)

The old way was "hidden" too, so moving a hidden feature to a less hidden place is not a problem. The menu is a better an more explicit interaction and, better accessible if people find it.

MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/d3c3ef45c3c89536e188b3fec38537912c8758f2

Issue #732, opened 2021-10-15
Missing Accessibility Info


kevlon300:
The "trend arrow" and all the buttons at the locale profile screen have no accessibility text. This is missing to use talkback.

Andries-Smit:
Thank you for reporting @kevlon300 we should keep our app open for everyone.
I am working on a fix for the 2 uses your reported, please contact me to discuss your need and if you are able to validate and help to test screenreader accessibility

PR #1200, opened 2022-01-19
Improve screenreader accesibility

Andries-Smit:
Improve screen reader accessibility solves issue #732 
More work is needed but will be different pull request


Andries-Smit:
Related PRs #1251 #1252  and #1253

diana1710:
i am very happy!!!
This test has brought me a very big relief.

On the home screen, you now only have to do half of the gestures, because the images are now assigned to the values. So everything is much clearer.

I am especially happy that the unlabeled buttons in the profile now have names. Changing an existing profile is now no problem at all.
When creating a new profile I have a single problem, if I want to enter another time in a profile, I can select the desired time from the drop-down list. It will then be displayed to me correctly. If I save everything, it does not contain the selected time, but the time that would follow next. If the previous time was 12:00 and I want to set 17:00, it says 13:00 after saving.

Issue #1159, opened 2022-01-12
[3.0 RC1] Dash pod initialization after discarding pod is failing on a wrong bluetooth address


filipwoz:
Repro Steps:
- try to deactivate dash pod
- break up the procedure and click on Discard Pod
- try to activate a new pod
Result:
Unhandled java Error message: 
![image](https://user-images.githubusercontent.com/20597551/149184646-5ccea16f-932d-44e3-a760-734bbce94162.png)

Expected:
After discarding a pod a new pod should be possible to activate without issues

Seems that by discarding a Pod cleanup of the buetooth connection address does not happen.

Logs attached:
[AndroidAPS._2022-01-12_07-01-11_.zip](https://github.com/nightscout/AndroidAPS/files/7856366/AndroidAPS._2022-01-12_07-01-11_.zip)
Note: do not wonder I did quite a strange workaround after that to make it running again (probably was not needed):
- Downgraded to 3.0 beta 11
- tried to activate a new pod (again with the same error)
- tried to deactivate the old pod and again discarded it
- tried to activate a new pod again - then it worked!
- Upgraded back to 3.0 RC1
I suppose what really helped was probably the re-discarding of the old POD and not the version downgrade.

avereha:
How did you "break up the procedure"?

filipwoz:
> How did you "break up the procedure"?

I am not sure as my son did it without me looking at than. I thought you could read it from the logs.
I think he was telling that it was taking a long time (or even showing a timeout message) and then he pressed the button to discard the pod.

PR #1212, opened 2022-01-21
fix race on pod activation

avereha:
Set Bluetooth address only once, after pod scanning. 
Do not set it on each connect.
Fixes: https://github.com/nightscout/AndroidAPS/issues/1159


vanelsberg:
2.8.2.20-dev, build 1d093bb64: Pod activated without problems.

Issue #1241, opened 2022-01-29
3.0 RC3 Unable to activate profile


HuismanErik:
Hey guys,

Today I tried to adjust something in my profile. Unfortunately, that didn't work out. And when activating AndroidAPS did a profile switch.
And the pulldown didn't let me select the right profile anymore. That's why I deleted the unwanted profiles. In the end, it seems to me that it is no longer necessary to activate the profile. When I press the activate profile button nothing happens. Below I have posted a number of screenshots and posted logging. I hope you can help me!

![Screenshot_20220129-143301_AndroidAPS](https://user-images.githubusercontent.com/7250929/151663360-65426a25-4660-43db-acd2-b4fa4743fb16.jpg)
![Screenshot_20220129-143657_AndroidAPS](https://user-images.githubusercontent.com/7250929/151663376-97c111dd-fea3-4cca-a205-2598c286296e.jpg)
![Screenshot_20220129-142824_AndroidAPS](https://user-images.githubusercontent.com/7250929/151663368-d4243ae2-bdd8-47b8-a100-60ca01ec2e3a.jpg)

[AndroidAPS_LOG_1643463038940.log.zip](https://github.com/nightscout/AndroidAPS/files/7963798/AndroidAPS_LOG_1643463038940.log.zip)

Sincerely,

Erik

HuismanErik:
Okay, I managed to reactivate the profile by making a backup and then restoring a backup from a week ago. Then reload the latest backup. I'm glad it's working again, just don't dare to adjust the profile anymore.

MilosKozak:
if dialog for profile doesn't appear no valid profile is detected. If you cannot find the bug in profile you can always remove it and create new

HuismanErik:
A new profile could also not be activated with the activate button. I have tried everything including a new profile and finally a database reset. After restoring a backup, the profile was active again. I was pretty desperate.
It is not clear to me what is wrong with the profiles I create. I can't find anything in the log file about the profile being disabled.
In the screen the profile seems to be good, there is no red background. The profile has also worked well for at least a month.

MilosKozak:
you can sent me non working configuration for more analysis

HuismanErik:
I have looked at the code and my problem has been solved by making a small change to the code. See the PR. Anyway, thanks for your offer.

PR #1246, opened 2022-01-30
all targets should be in range instead of any

HuismanErik:
Yesterday my profile was not valid. But this was not visible in the application. It was because one of the targets was too low. 
Other targets were correct, it was not visible in the application.
This PR fixes [#1241](https://github.com/nightscout/AndroidAPS/issues/1241)


MilosKozak:
good catch, thanks! it was missing on more places than in your PR so I created https://github.com/nightscout/AndroidAPS/commit/f3c83ef582be65661ff0e4e14b38339264fd027c

HuismanErik:
No problem. I'm happy to help!

PR #1251, opened 2022-01-31
Improve a11y table structure with row focus for screen readers

Andries-Smit:
A common pattern for AAPS is to create a table with information <label> : <value>. Screen readers have a hard time cycling through them.
A pattern to make life easier for screen reader users is by grouping related information. This PR makes the rows focusable, and let the screen reader announce the label and value combined.

image

### DIFF
--- a/app/src/main/res/layout/careportal_stats_fragment.xml
+++ b/app/src/main/res/layout/careportal_stats_fragment.xml
@@ -8,7 +8,8 @@
 
     <TableRow
         android:layout_width="fill_parent"
-        android:layout_height="wrap_content">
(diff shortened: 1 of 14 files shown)


vanelsberg:
Fortunately for me I am not very familiar with this way of interacting with the UI.
For I can judge this just works. Screen reader finds all information items on the AAPS tabs as will as on menu's.

diana1710:
The row selection is much better than cell selection for screenreader. Only in Open APS Ama it is difficult to follow the lines in the section from PredBGS. I can use ther a special key from Talkback to read only a part from the 3 hours.

PR #1252, opened 2022-01-31
Improve a11y hide dialog header image for screen reader

Andries-Smit:
Most dialogs contain an image a title in the header. The image is for decoration purposes only and should be hidden for the screen readers.

image


diana1710:
Everything works as expected. Thank you for all the improvements. Everything works much easier this way.

PR #1253, opened 2022-01-31
Improve a11y wizard dialog

Andries-Smit:
Wizard dialog is complex to navigate, improving accessibility by
* All checkboxes use a label, labels are clickable, making the touch target larger
* Number picker 
  * Number support custom description
  * Announce value change on +/ - button
  * Text input uses labelFor
* Checkbox with image label, uses the image as a label, making the touch target larger
* Converted the image controls for calculation parameters into real checkboxes, leveraging on standard a11y features
* Converted the ok cancel button to okcancel layout, fixed left linking of cancel button to close the dialogue
* Show confirmation "no action" dialog, when no action is taken (0 carbs, 0 insulin)

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/dialogs/WizardDialog.kt
+++ b/app/src/main/java/info/nightscout/androidaps/dialogs/WizardDialog.kt
@@ -4,11 +4,7 @@ import android.content.Context
 import android.os.Bundle
 import android.text.Editable
 import android.text.TextWatcher
-import android.view.LayoutInflater
-import android.view.View
-import android.view.ViewGroup
-import android.view.Window
-import android.view.WindowManager
+import android.view.*
 import android.widget.AdapterView
 import android.widget.AdapterView.OnItemSelectedListener
 import android.widget.ArrayAdapter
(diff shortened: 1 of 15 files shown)


app/src/main/java/info/nightscout/androidaps/dialogs/WizardDialog.kt
```diff
@@ -241,6 +246,26 @@ class WizardDialog : DaggerDialogFragment() {
             }, fabricPrivacy::logException)
         )
 
+        setA11yLabels()
+    }
+
+    private fun setA11yLabels() {
+        val bgInputTextId = binding.bgInput.editText?.id
```
MilosKozak, line 253: kotlin way is:
`binding.bgInput.editText?.id?.let { binding.bgInputLabel.labelFor = it }`

Andries-Smit: Fixed. Thanks for the feedback, I like to learn.

diana1710:
I usually use the text field and enter the values via the keyboard. Especially for beginners this is difficult: First you have to find the keyboard, enter everything and switch to ready, then find the text field again and check if everything was entered correctly....

The possibility to already know the step size before selecting plus or minus and then hear what value was entered is great.
It is very nice to have the result moved to a front position.
As it is now, it is wonderful. I have the choice of entering the value directly or going to plus or minus.

diana1710:
Hello,

The list of blood glucose values is now quite easy to read.
All the dialogues you listed I have tested extensively. It's all great and very helpful. !

Using the local profile was the most complicated thing for me so far. I feel very safe there now.

I am really happy about the great changes, although I was also very satisfied before.

PR #1269, opened 2022-02-03
fix: rounding cob, basal in SMB AMA reason

### LINKED ISSUE #165: [2.7.1-dev] [cosmetic-issue] AAPS show carbs in change request in extrem precision
AAPS show carbs in change request in extrem precision.

Andries-Smit:
The values in the reasons of AMA and SMB should be appropriately rounded. This makes the text easier to read, and open-loop users will not see a strange suggestion. (Personally, It gave me a bad feeling and doubts about Android APS the first time I was using it)

Besides this, it is not funny to hear on your screen reader trying to utter o o o o o o o o o o o o o o o 3 

This PR round COB  to 1 digit, rounds the bases to 2 digits (Fixes #165 rounding COB only occurred with AMA)

image

Note to the reviewer: this change should only be limited to the string `reason` and should not impact any calculations.


MilosKozak:
We discussed this before if it's good idea to modify original JS code. Maybe now when there are no updates to OAPS we could do it. @AdrianLxM ?

AdrianLxM:
No objections for AMA. This is basically a fork from an older snapshot.

PR #1292, opened 2022-02-08
profile editor add DIA and Target Chart

Andries-Smit:
Improvement profile configuration
* Add chart for DIA.
* Add chart for Aria Target BG.
* Prevent invalid input with + and minus button, by correct rounding min/max of ISF  and Target BG (issue occurred only  with mmol )
* Support different ranges for target low and target high.
* Move profile name input field out of the DIA tab.
* Prevent the wrong target high and low.
  * set low target equal to high target, when the high target is lower than the low target. 
  * set high target equal to low target, when the low target is higher than the high target. 
* Chart chart is also available in the profile dialog, (click on current profile home) and when comparing profiles in the profile helper
* For usability, make use of the full label of the time range editor

Image

Image

Remark; looking at the java code and UI design, there is much room for improvement, a full redesign should be considered


vanelsberg:
Yes. Looks great Andries :-)

core/src/main/res/values/strings.xml
```diff
@@ -106,11 +106,11 @@
...
-    <string name="dia_label">DIA</string>
-    <string name="ic_label">IC</string>
-    <string name="isf_label">ISF</string>
-    <string name="basal_label">Basal</string>
-    <string name="target_label">Target</string>
+    <string name="dia_label">Duration of Insulin Action</string>
+    <string name="ic_label">Insulin to Carb ratio</string>
+    <string name="isf_label">Insulin Sensitivity Factor</string>
+    <string name="basal_label">Basel rate</string>
+    <string name="target_label">Blood Glucose Target</string>
```
MilosKozak, line 113: at least `target_label` is used elsewhere. Making the string long breaks something

core/src/main/java/info/nightscout/androidaps/utils/ui/TargetBgProfileGraph.kt
```diff
@@ -0,0 +1,99 @@
...
+        val units = profile.units
+        for (hour in 0..23) {
+            val valueLow = Profile.fromMgdlToUnits(profile.getTargetLowMgdlTimeFromMidnight(hour * 60 * 60), units)
+            val valueHigh = Profile.fromMgdlToUnits(profile.getTargetHighMgdlTimeFromMidnight(hour * 60 * 60), units)
+            maxValue = max(maxValue, valueHigh)
+            targetArray.add(DoubleDataPoint(hour.toDouble(), valueLow, valueHigh))
+            targetArray.add(DoubleDataPoint((hour + 1).toDouble(), valueLow, valueHigh))
+        }
+        val targetDataPoints: Array<DoubleDataPoint> = Array(targetArray.size) { i -> targetArray[i] }
+        val targetSeries: AreaGraphSeries<DoubleDataPoint> = AreaGraphSeries(targetDataPoints)
```
MilosKozak, line 36: ISF has backround, IC has not. Voting for graph without

Andries-Smit: ISF should have not background, basal and target range area should

Andries-Smit: fixed

PR #1301, opened 2022-02-10
Fix WizardInfoDialogMmol

### LINKED ISSUE #1298: [AAPS 3.0.0] Bolus Wizard calculation Settings page BG value shown as x/18 mmol/l.
Using Bolus Wizard for bolus calcultation and the results are shown in Treatments tab. Openig the results from green "Calc" link that shows the calculation settings. The first line has BG value left to ISF value but BG is not shown correctly but as a value x/18 mmol/l .

Philoul:
Fix #1298
I propose to convert `BolusCalculatorResult` values in MGDL in `BolusWizard` and back to selected unit in `WizardInfoDialog`

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/dialogs/WizardInfoDialog.kt
+++ b/app/src/main/java/info/nightscout/androidaps/dialogs/WizardInfoDialog.kt
@@ -69,13 +69,15 @@ class WizardInfoDialog : DaggerDialogFragment() {
         binding.close.setOnClickListener { dismiss() }
         val units = profileFunction.getUnits()
         val bgString = Profile.toUnitsString(data.glucoseValue, data.glucoseValue * Constants.MGDL_TO_MMOLL, units)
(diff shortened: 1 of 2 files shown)


app/src/main/java/info/nightscout/androidaps/dialogs/WizardInfoDialog.kt
```diff
@@ -69,13 +69,15 @@ class WizardInfoDialog : DaggerDialogFragment() {
         binding.close.setOnClickListener { dismiss() }
         val units = profileFunction.getUnits()
         val bgString = Profile.toUnitsString(data.glucoseValue, data.glucoseValue * Constants.MGDL_TO_MMOLL, units)
+        val isf = Profile.toUnits(data.isf, data.isf * Constants.MGDL_TO_MMOLL, units)
+        val trend = Profile.toUnitsString(data.glucoseTrend * 3, data.glucoseTrend * 3 * Constants.MGDL_TO_MMOLL, profileFunction.getUnits())
```
MilosKozak, line 73: use `units` here too

PR #1307, opened 2022-02-11
Update XdripPlugin.kt

Florian7784:


MilosKozak:
this would enable SMB always for all sensors which is not what we wants

PR #1310, opened 2022-02-12
A11y/apply labels number picker

Andries-Smit:
Set the labels for screen reader for all number pickers based on the #1253 
- Add content description, when focus on - and + button the increment step size is announced
- Announce value change on +/ - button
- Set "labelFor" to announce the label when text input is focused

For the forms and dialogs
- profile helper
- calibration
- carbs
- care
- extended bolus
- fill
- insulin
- profile switch
- temp basal
- temp target
- treatment
- local profile


Andries-Smit:
Commit was placed other PR https://github.com/nightscout/AndroidAPS/pull/1253#issuecomment-1038007795

> The list of blood glucose
>
>  values is now quite easy to read.
> All the dialogues you listed I have tested extensively. It's all great and very helpful. !
> 
> Using the local profile was the most complicated thing for me so far. I feel very safe there now.
> 
> I am really happy about the great changes, although I was also very satisfied before.

PR #1314, opened 2022-02-12
Dynamic isf dev

tim2000s:
Update to AAPS Determine Basal using Chris WIlson's Dynamic ISF code. Can be created as a separate branch to allow testing for users.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.kt
@@ -268,7 +268,7 @@ class DetermineBasalAdapterSMBJS internal constructor(private val scriptReader: 
         this.mealData.put("lastCarbTime", mealData.lastCarbTime)
 
         tddAIMI = TddCalculator(aapsLogger,rh,activePlugin,profileFunction,dateUtil,iobCobCalculator, repository)
-        this.mealData.put("TDD_AIMI7", tddAIMI!!.averageTDD(tddAIMI!!.calculate(7)).totalAmount)
+        this.mealData.put("TDDAIMI7", tddAIMI!!.averageTDD(tddAIMI!!.calculate(7)).totalAmount)
         this.mealData.put("TDDPUMP", tddAIMI!!.calculateDaily().totalAmount)
 
         if (constraintChecker.isAutosensModeEnabled().value()) {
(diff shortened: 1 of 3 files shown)


hartlijn:
this good be a game changer for a lot of people!
Hope I didn't miss some documentation (I am not too techy to be able to find the exact code and therefore don't know where to read in-code comments); is there any?

1. WIll this fully replace the ISF value in the Profiles?
2. Will Autosense still increase/decrease ISF in combination with this new ISF calculation?

app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/DetermineBasalAdapterSMBJS.kt
```diff
@@ -249,7 +264,13 @@ class DetermineBasalAdapterSMBJS internal constructor(private val scriptReader:
         this.mealData.put("slopeFromMaxDeviation", mealData.slopeFromMaxDeviation)
         this.mealData.put("slopeFromMinDeviation", mealData.slopeFromMinDeviation)
         this.mealData.put("lastBolusTime", mealData.lastBolusTime)
+        this.mealData.put("lastBolusNormalTime", lastBolusNormalTime)
```
MilosKozak, line 267: It's always passing zero?

app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
```diff
@@ -76,7 +76,55 @@ class TddCalculator @Inject constructor(
...
+            val tbr = iobCobCalculator.getTempBasalIncludingConvertedExtended(t)
+            val profile = profileFunction.getProfile(t) ?: continue
+            val absoluteRate = tbr?.convertedToAbsolute(t, profile) ?: profile.getBasal(t)
+            tdd.basalAmount += absoluteRate / 60.0 * 5.0
+
+            if (!activePlugin.activePump.isFakingTempsByExtendedBoluses) {
+                // they are not included in TBRs
+                val eb = iobCobCalculator.getExtendedBolus(t)
+                val absoluteEbRate = eb?.rate ?: 0.0
+                tdd.bolusAmount += absoluteEbRate / 60.0 * 5.0
```
MilosKozak, line 113: here should be be used `calculationStep` too to be clear

MilosKozak:
I'd probably treat this as new plugin to not everyone switch to it by default and enable it in engineering mode for now. It's easily achieved in AAPS

MilosKozak:
based on what i've seen Chris is making his ideas patented. This makes it hard usable in opensource without proper licensing

MilosKozak:
If AS should be disabled and ISF from profile not used at all we should cover it in code too

tim2000s:
Chris is code is patented except for use in WeAreNotWaiting, so that's covered. 

I'll make someodanon the other branches, but I think this benefits the community, so if you want me to run as a  Standalone model for a while in beta then submit for a Dev update (possibly as a separate plugin), happy to figure that out.

PR #1328, opened 2022-02-16
Feature: show/hide battery status light

### LINKED ISSUE #937: Configurable warning lights including Orangelink battery level [enhancement]
Warning lights on the overview screen are really useful, but can contain indicators that aren't used (e.g. Sensor battery level shows '-' for BYOD users) and doesn't show others that are used (e.g. Orangelink battery for Eros/Medtronic users).

It would be useful:
- if the user could enable/disable warning indicators and for Orangelink to be added as a configurable item
- OR if AAPS automatically displayed relevant warning lights based on the system's configuration

Andries-Smit:
The overview status lights can show battery information, however, it is not always supported or needed, for those cases, we should hide them:
* Patch pumps discards batteries and do not need to show it in the status battery light except Omnipod Eros
* Only show battery level if the action button change battery is available
* Only show battery level when supported by pump or link

In the action layout, we hide the battery row when is not needed at all
*  Patch pumps discards batteries and do not need to show it in the status lights except Omnipod Eros

image

Fixed #937


MilosKozak:
Again. I'd not mention Eros, G8 etc but add property to describe pump and this flag

Andries-Smit:
Feedback @MilosKozak processed

m-scha1337:
How come `app/src/main/res/layout/careportal_stats_fragment_lowres.xml` got fully axed?
Since this is still referenced in [app/src/main/res/layout/actions_fragment_lowres.xml:129](https://github.com/nightscout/AndroidAPS/blob/d5aece1ed841db6aeda5da59cb4834db55ba37f9/app/src/main/res/layout/actions_fragment_lowres.xml#L129), the build seems to fail (resource not found)..

Just wanted to let you know.. ¯\\\_(ツ)_/¯

Issue #1330, opened 2022-02-16
Proposal: wizard remove on click icons handling


Andries-Smit:
The Bolus Calculator wizard got a new option since version 3.0. There are icon indicators on the bottom section that shows what options are used in the calculation. 
![image](https://user-images.githubusercontent.com/6724749/154282334-4b2d4a92-1b17-4480-934e-5801315cc9ea.png)
On click of an item it toggles the setting, however, I see some risk with the implementation:
* From a UX/UI perspective, users will not intuitive understand that it is actual checkboxes
* They are close the cancel button and can be clicked accidentally
* When toggled, the setting will be stored, and used in the next calculation too

We risk that the user will accidentally change the parameters and only will figure out after many wrong boluses.

My proposal is to remove the click interaction and only show the settings, so the settings can only be changed when the show checkbox is checked. 

Please let me know what your perspective is?

swissalpine:
You're right, you have to know that these are buttons - but that's also the case in other places.
More important fo me: I love the new "buttons"! They are really handy!
However, I had to adjust the display for my small smartphone, so now there is no danger of clicking them by mistake (and I changed so that they no longer disappear when the calculation is shown) :
![Screenshot_20220114-062916](https://user-images.githubusercontent.com/12517693/154305973-bb258249-a936-4c60-819b-09c01111ccee.png)

Andries-Smit:
@swissalpine Your layout is nice, it has a bigger touch area and less change for accidental clicking

Philoul:
I think it's a good idea to keep the click in these icons (that's why I made PR #1156 😉) and I identified the risk of too small icons not easy to click when I made this PR so I'm not surprised.
I think it's a good idea to increase the size of Icons on a seperate line (there is enough place on all screen when detailled information is hidden)

I think we can also improve the logic between IOB and COB icons/checkbox (it's another subject but linked because we don't see on Icons if a click is enable or not, checkbox is greyed when disabled...).
- Currently IOB checkbox is checked and disabled when COB is checked (for icons, no action is done if you click on IOB icon while COB icon is "colored"). For safety it's dangerous to have cob checked and iob unchecked.
- Currently when both are unchecked, if you check COB, then IOB is automatically checked and disabled (IOB icon automatically colored and no action available on click)

My proposal would be to always keep both icons (and checkboxes) enabled, check both when COB is checked, uncheck both when IOB is unchecked, and allow IOB checked alone if the user check IOB or uncheck COB...

[...]

PR #1333, opened 2022-02-17
On click icons in a row above cancel/ok

swissalpine:
If you like it, you can merge it.
This should be the layout shown in
https://github.com/nightscout/AndroidAPS/issues/1330


Andries-Smit:
@swissalpine Thank you for the PR.
The icons were replaced by checkboxes with background images, to support screen readers.

swissalpine:
Thanks for the hint with the checkboxes

PR #1340, opened 2022-02-18
Background images needs another sizing

swissalpine:
Sorry for the renewed PR - but I was not aware that the size ratios of the background images differ from the "normal" images.


Philoul:
Size is better, but icons are not be centered and BG icon is truncate on the left...
image

Issue #1349, opened 2022-02-20
DynamicISF safety limiter bug


robertrub:
@MilosKozak If I select the DynamicISF option, I have a safety limiter at 1.5 (mine is 10U) which may (or may not) come from my basal rate (0.5) x max multiplier (3).

![Screenshot_20220220-100849_AndroidAPS.jpg](https://user-images.githubusercontent.com/76626474/154836862-3db20f8e-1651-4171-bb06-57bccb451132.jpg)

This is a game stopper of course ;) No more SMBs, no more high basals etc.

If I select OpenSMB, the same limiter goes to 22.

![Screenshot_20220220-102857_AndroidAPS.jpg](https://user-images.githubusercontent.com/76626474/154836921-c6e01ea1-6bb9-41a1-97e2-fe80f516b774.jpg)

philipgo:
Same problem with latest dev and AutoISF here.

bchris21:
Same here (3.0.0.1-dev-a) but with different value, maybe because it refers to "child" profile.

Screenshot with SMB last run on 10:56 is with Dynamic ISF

![Screenshot_20220220_112134.jpg](https://user-images.githubusercontent.com/4725280/154838234-b350529c-a3da-42b3-a090-8a869dd74b1a.jpg)

Screenshot with SMB last run on 11:00 is with OpenAPS SMB plugin.

![Screenshot_20220220_112154.jpg](https://user-images.githubusercontent.com/4725280/154838254-bff588ff-63df-4a44-aa04-3bd6af3b3e2c.jpg)

Alamo04:
Confirm, same here.

elgrupo123:
unfortunately also here.  IOB max 4IU

vanelsberg:
Confirming,
Duplicate of https://github.com/nightscout/AndroidAPS/issues/1346

swissalpine:
Please note: The first commit (https://github.com/swissalpine/AndroidAPS-1/commit/a0c3fad06ea3dad73e3819794e7889e427ccb664) is wrong, the second works for me:
https://github.com/swissalpine/AndroidAPS-1/commit/6a961eb4d8b949ff8564bcd17f0f3b2469e0beda

Alamo04:
@MilosKozak @swissalpine 
There are 2 DynamicISF errors in building the App, looks like kotlin errors.
Think they are not importand and self fixed in this version.
[...]

Issue #1346, opened 2022-02-19
OpenAPS SMB Dynamic ISF, SMB reports limiting max_iob to 1.5U (not following preference settings?)


vanelsberg:
**AndroidAPS version**: 10
**Build number**: 3.0.0.1-dev-8, build fef8421e7
**Flavor**: fullrelease

## Phone: ##

Samsung Note 9

## Problem description: ##
While testing OpenAPS SMB Dynamic ISF I noticed staying on high BG's for far to long while 0-Temping and SMB not kicking in. Looking into this found the SMB tab reporting constraints "_Safety: Limiting IOB to 1.5U because of max value in preferences_". 

When switched back to standard "OpenAPS SMB" constraint reported "_Safety: Limiting IOB to 8.0U because of max value in preferences_" as expected.

Looks like OpenAPS SMB Dynamic ISF is not following the settings from preferences?
Unless this is part of the "Dynamic ISF" functionality, I would say that is a bug?

_Additional information and reproduction steps:_

Using "Sensitivity Oref1" or "OpenAPS AMA" 

1. Make sure "_OpenAPS SMB_" is selected for APS
2. Under preferences, make a note of max_iob settings
3. Check the SBM tab under constraints. It should report a limiting value as configured in setting  under (2)
4. Switch to the new "_OpenAPS SMB Dynamic ISF_" plugin
5. Under preferences, check if the max_iob settings is showing?
6. Check the SBM tab under constraints. Confirm It now reports a limiting value for max_iob, different form settings.

In general, I notice SMB is not kicking in while IOB is over 1.5U

![Screenshot_20220219-123339_AndroidAPS](https://user-images.githubusercontent.com/45144129/154800489-a9a0d059-9aab-4f68-9f29-5b5f954e5660.jpg)
![image](https://user-images.githubusercontent.com/45144129/154802993-0f4d1410-da33-4d0b-a2f0-dce78570b251.png)

## Logfiles: ##
Around 13:25h
Restarted AAPS
Run SMB
Run SMB
Select "_OpenAPS SMB Dynamic ISF_"
Restarted AAPS
Run SMB
Run SMB
Export logfiles

[AndroidAPS._2022-02-19_13-26-05_.0.zip](https://github.com/nightscout/AndroidAPS/files/8102160/AndroidAPS._2022-02-19_13-26-05_.0.zip)
[AndroidAPS.log](https://github.com/nightscout/AndroidAPS/files/8102161/AndroidAPS.log)

PR #1352, opened 2022-02-20
Fix for DynamicISF safety limiter bug

swissalpine:
https://github.com/nightscout/AndroidAPS/issues/1349
https://github.com/nightscout/AndroidAPS/issues/1346


vanelsberg:
Think this solves the problem 👍

PR #1355, opened 2022-02-20
Fix Layout and Size for TT Icon in Wizard

Philoul:
same than #1340 for TT Icon


swissalpine:
Nearly the same as https://github.com/nightscout/AndroidAPS/pull/1355
👍

PR #1364, opened 2022-02-23
Fix Google's NTP can't use in china.

LuckyJame:
Due to the firewall, Google's NTP website can't be accessed in China，
In Objective Plugin,  We have a error" Failed retrieve time".
It's work good at “ntp.aliyun.com”...
And we translation for objectives,exam,core,wear,dana etc.


MilosKozak:
* this method would mean add translation for all languages
* better would be detect `Locale` and use appropriate time server
* translations must go through Crowdin otherwise they will be overwritten
* with new PR you must target dev branch

LuckyJame:
ok~  i will do it soon.
first  join in crowdin ~tks

PR #1374, opened 2022-02-25
Reverted Dynamic ISF code to previous following user feedback

tim2000s:
Rolling 24 hour code can be more aggressive following a site failure, Previous code was less so and users generally prefer this behaviour.


mick711:
why not add possibity to use one off them ?

MilosKozak:
@tim2000s i reverted to 1.3 as this code is crashing

PR #1412, opened 2022-03-05
Fix notification color contrast

### LINKED ISSUE #446: White text on red : not easy to read
The alert text is not easy to read. It would be easier to read if the contrast was higher.

Andries-Smit:
Color contract is imported for some users, but making for all users text easier to read.
Current 1.86:1, Suggested 4.5:1
To fix this, the background color has to change or the text color. Because the background color has a meaning, it should not be changed to much. The PR fixes #446 by suggesting to change:
* Text color to black, 
* Text style bold
* Tweak green notificationInfo color for background with black text

image


osodebailar:
Good idea. But can you do me the favour and set the color by attribute and not by  direct in the color table.
This is one of the point that make my themeswitcherbranch greater and greater, because merging is not enough. i had to change the colors and code to set the color by attribute. 
Setting colors by attribute is one of the necessary main things that are necessary to change themes. in this case from light to dark

Andries-Smit:
@osodebailar Done

PR #1422, opened 2022-03-08
Fix Wizard Dialog Layout

Philoul:
When TT icon is visible, the Carb icon was outside the screen on my loop phone...
Fixed by this PR
image

### DIFF
--- a/app/src/main/res/layout/dialog_wizard.xml
+++ b/app/src/main/res/layout/dialog_wizard.xml
@@ -242,7 +242,7 @@
         </LinearLayout>
 
         <LinearLayout
-            android:layout_width="wrap_content"
+            android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_gravity="center|center_vertical"
             android:layout_marginTop="10dp"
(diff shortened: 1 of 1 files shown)


Philoul:
@Andries-Smit Why do you ask to replace `ImageView` by `CheckBox` in Layout ?
=> Now images are not centered... (I think it was better before, do you know other parameters to improve layout ?)

swissalpine:
They are nearly centered if you set android:layout_width="20dp"
That's what I've done to fix both issues: images not centered and icons aren't competly shown.

Philoul:
No, it seems to be the same with 20dp...
Results is different when the layout_weight parameter is used.

Andries-Smit:
> why?
@Philoul for Accessibility. images are not explained by screen reader, blind people don't know what to do with them. When changing an image into a button, you can add description, however now we use them as a checkbox, and then you have to tell the user the check state on or off. Manually adding that code is 30 line. Whe using checkboxes you get that for free, and it is consistent what the expected UI

Philoul:
@Andries-Smit Thanks for the explains
Can we use an ImageButton instead of a Checkbox for accessibility ? (it's probably easier to have image centered)

Andries-Smit:
@Philoul image buttons do not have a state "checked" and "unchecked" we can make the checkbox work. Ill get back to you

Andries-Smit:
@Philoul toggle buttons are the better option for this case. I opened an alternative PR #1427 If this is good enough that one can be merged. If you like to improve on it, you can copy it.

PR #1435, opened 2022-03-11
Updates to Dynamic ISF Plugin

tim2000s:
Adjustment factor for Dynamic ISF added to allow modification of the ISF outcome;
Austosens based on Rolling 24 hour TDD/7-day Average TDD in Determine Basal to replace Autosens outputs; is used to modify basal rate and to adjust target when Autosens switch is enabled (for testing before rejigging profile and Autosens components).
When glucose levels are flat and eventual BG is predicted below current bg, use an average of current and expected BG to determine dosing ISF.


app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMBDynamicISF/DetermineBasalAdapterSMBDynamicISFJS.kt
```diff
@@ -194,9 +194,9 @@ class DetermineBasalAdapterSMBDynamicISFJS internal constructor(private val scri
         this.profile.put("current_basal_safety_multiplier", sp.getDouble(R.string.key_openapsama_current_basal_safety_multiplier, 4.0))
 
         //mProfile.put("high_temptarget_raises_sensitivity", SP.getBoolean(R.string.key_high_temptarget_raises_sensitivity, SMBDefaults.high_temptarget_raises_sensitivity));
-        this.profile.put("high_temptarget_raises_sensitivity", false)
```
MilosKozak, line 197: @AdrianLxM do you remember why this was always false?

tim2000s:
If I remember correctly, it was because it caused issues with regular profile switching resulting in very small ISF numbers. 

I think there's an additional change in the code required to fully enable it though, as in this push, the switches in preferences remain greyed out.

MilosKozak:
what happened there is so many commits?

MilosKozak:
@tim2000s could you squash these commits?

PR #1436, opened 2022-03-12
Fix nightscout client big time text

### LINKED ISSUE #1308: Large NSClient time display
Reporting bugs
--------------
- **Note the precise time the problem occurred** and describe the circumstances and steps that caused
  the problem
- Note the Build version (found in the About dialog in the app, when pressing the three dots in the
  upper-right corner).
- Obtain the app's log files, which can be found on the phone in
  _/storage/emulated/0/Android/data/info.nightscout.androidaps/_
  See https://androidaps.readthedocs.io/en/latest/EN/Usage/Accessing-logfiles.html
- Open an issue at https://github.com/nightscout/AndroidAPS/issues/new

Andries-Smit:
When running the nightscout client, the time is shown, this was too big and made part of the UI unusable
With the PR the font size is made smaller.

image

Fixes #1308

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/skins/SkinInterface.kt
+++ b/app/src/main/java/info/nightscout/androidaps/skins/SkinInterface.kt
@@ -48,8 +48,10 @@ interface SkinInterface {
 
             if (isTablet) {
                 binding.infoLayout.apply {
-                    val texts = listOf(bg, time, timeAgoShort, iob, cob, baseBasal, extendedBolus, sensitivity)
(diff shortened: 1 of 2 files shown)


MilosKozak:
in tablet mode it should stay large

Andries-Smit:
We can do dimentions and resource qualifiers.

Andries-Smit:
@MilosKozak fixed with Skin tabled landscape the text will stay the same, is that good enough?

PR #1464, opened 2022-03-20
Tidepool Plug-in back on.

emmatovar27:
Compared to the original version of the plugin, the main change is only uploaded finished TBR to avoid overlapping Temp basals in the Tidepool website. 
One recommendation is only to activate and upload once a day. (Simulate the Tidepool uploader behavior of one dump of data instead of continuous stream information)


MilosKozak:
this is not the reason why Tidepool plugin is disabled. It was not possible to upload profiles properly thus data in Tidepool was wrong

PR #1485, opened 2022-03-22
Notes displayed differently in graph to make them more visible 

kristinsmotz:
I had trouble at some points viewing the bg graph, especially around the time a profile switch was enacted, after the profile text paint.style was changed from stroke to fill_and_stroke around v.3.0 launch. Tried playing with the settings to see if I could improve the visibility of all the variables in some way. This is where I found it looks good to me. The text size is reduced for all three variables I have changed to allow more contrast from background and other values, and more of the profile name to be visible when using a longer profile name.
Notes are now displayed differently in the main graph area of the Overview tab to make them more visible, while not hiding BG values. Changed colors of Exercise (now magenta) and Notes (now cyan). Changed Paint.Style and scaled.TextSize for Exercise, Notes, and Profile.
Screenshot_20220322-133437_NSClient
.


nicoleloop:
I really like this way to "de-clutter" the graph, has there been any more progress towards including this commit into /dev anytime soon?

kristinsmotz:
Will reassess this when I have tested a bit more with color options in new day/nightmode styles

Issue #1483, opened 2022-03-22
Background reverted from black to gray


MilosKozak:

Alamo04:
Thanks for report this issue Milos.
Hope this will be fixed soon, love black and hate this lighted grey. :(

PR #1487, opened 2022-03-23
set darkmode background

osodebailar:
Background in darkmode is set to real black . 

TODO: Set the dialogs background color , eg. Carbs Dialog , Wizard dialog, Insulin dialog , progress dialog etc.
            I make a proposal in the next days. 

@MilosKozak : The dialogs in the future , with or without rounded dialogs ? 
                         Latest in material 3 rounded dialogs are standard. I can make a proposal too. 

This will fix: https://github.com/nightscout/AndroidAPS/issues/1483


MilosKozak:
do you have screenshots to see the difference?

PR #1494, opened 2022-03-23
rounded dialogs like in android 12

osodebailar:
@MilosKozak 
Here as as proposal rounded dialogs for aaps. Latest in android 12 standard.

rounddialog_01
rounddialog_02
rounddialog_03
rounddialog_04


Andries-Smit:
@osodebailar this is part of Material 3. https://m3.material.io/components/dialogs/overview
We are making big steps moving towards material 2. While doing this it feels hasty to move already to 3 features manually and we we do we might get most of the material 3 features for free, when we really do the update, including corners and spacing.

When we are redesign to redesigning the dialogues, we might want to follow the full material 3 guidelines https://m3.material.io/components/dialogs/guidelines and remove the header color, and using the same spacing.

osodebailar:
@Andries-Smit the rounded dialogs are introduced by Google before material 3 , exact since SDK28. In Material 3 they are fully astablished.
I am working for over 1,5 years with my gui branches and the users who uses my branches like them very much . 
A complete redesign of the dialogs can be done afterwards, piece by piece .

Andries-Smit:
@osodebailar indeed, the rounded dialogs are from before MD3, and with MD3 they become more pronounced. (Funny, I never noticed) And I do not doubt some people will like rounded corners.  

The changes we do to the styling should aim to **simplify** the code and styling and to improve **(code) quality**, the feature is a result a good programming, not a goal in itself. The proposed changes can be done without code modification, I will DM you for some details.

osodebailar:
@Andries-Smit  @MilosKozak : tested here so far and seems to be good.

PR #1511, opened 2022-03-28
chore: Unified experience for sort and delete in QuickWizard, BG Source, Automation and treatments

Andries-Smit:
Unified experience for sort and delete in QuickWizard, BG Source and Automation. Based on feedback @Philoul 
- When used from the hamburger menu, the action delete and sort (if applicable) are already shown
- Removed the Remove and Sort item from from the overview option menu, to prevent glitches and make the menu consistent across all tabs
- When BG Source, Automation are in the tabs, the Long press of an item will start the action, showing remove and sort button. when only one action (BG source only supports a remove) the action will started.
- Refactor to use a shared ActionModeHelper to manage the option menu (QuickWizard, BG Source and Automation, treatments)
- Refactor to use a shared DragHelper for QuickWizard and Automation.
- Make treatment options visibility and remove part of the title base if there is room
- Fix refresh of list when show/hide is selected


Philoul:
Latest version is great for automation, quickwizard and BG plugin (in tab or in Hambuger menu).
Only Treatments still need to be improved a bit I think: 
in several tabs we miss some options
- Carbs and Bolus : delete future treatments and refresh from Nightscout
- Temp Target : refresh from Nightscout
- CarePortal : refresh from Nightscout
- User Entries : I think we can remove menu and use same logic to show/hide loop entries and add an icon for export (both are with a button + an icon in current master)

Andries-Smit:
@Philoul  The other options are conditional, and only visible under a condition
- Delete future treatments (only when you have scheduled a treatment in the future
- Refresh from Nightscout (only when you run the NS client
See: https://github.com/nightscout/AndroidAPS/pull/1511/files#diff-47b0a0103893e5c21c0ad3ba835bd9c13dd0277fa1607cf8370b463993f4f3f7R321-L310

Implemented  
- User Entries show/hide loop entries feature

PR #1537, opened 2022-04-02
Theme switcher preparation:  graphdata treatmentdata by attribut and final preperations

osodebailar:
@MilosKozak : This is the last main pull request.
I will further work on minor things in the fragments and dialogs for coloring.
Its up to you that we can start a wider test for the main developers by open the theme switching in preferences.

Graphdata and treatmentdata colors now accessed by attribute
Nearly all Fragments and dialogs can now be used in light theme !

TODOS:
- Testing, testing, testing.
- Working on colors to establish a very well looking light theme. 
- After this work is finished: cleaning the colortable, remove duplicates and discuss naming of colors.The main meaning is given
 by the attribut-name
- establish theme switching in pref_general

Colors in objectives @MilosKozak : Color handling here is a little bit strange hardcoded. Maybe someone can help

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
@@ -205,7 +205,7 @@ class HistoryBrowseActivity : NoSplashAppCompatActivity() {
         binding.bgGraph.gridLabelRenderer?.reloadStyles()
         binding.bgGraph.gridLabelRenderer?.labelVerticalWidth = axisWidth
 
-        overviewMenus.setupChartMenu(binding.chartMenuButton)
+        overviewMenus.setupChartMenu(context, binding.chartMenuButton)
         prepareGraphsIfNeeded(overviewMenus.setting.size)
         savedInstanceState?.let { bundle ->
             rangeToDisplay = bundle.getInt("rangeToDisplay", 0)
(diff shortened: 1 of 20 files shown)


swissalpine:
I am currently having a very hard time reading the rather dark gray font in the tabs.
My attempts to change this in the colors.xml files fails (defaultext, tabtext etc.).
Is this problem related to the changeover and will it solve itself in a few days with your latest PRs?
Is there a individual hotfix that I can implement?
Thanks for your work!

osodebailar:
The PRs are all done in the moment .
the default fonts are given by Google . We did not have here special things, so that the theme switching can act . 
Maybe you have problems by the darker background. Or do you have problems I the light theme in the moment? 
The tab fonts Color’s I think were implemented by another one . Normal I would not touch them

swissalpine:
Thank you for your reply.
Maybe a screenshot clarifies what I mean:
Partly the fonts are colored in the combo tab (warncolors.kt). The not colored ones are in a very dark gray, especially on a small display (Jelly2) the openapsama_fragment.xml is therefore only with great difficulty to read.
Screenshot_20220402-221353

swissalpine:
I think I have found a solution:
In night/styles.xml I added the following line:
<item name="android:textColor">@color/white</item>

osodebailar:
Please do not use Color’s direct !!! It is important, very important to use only attributes .
?attr/defaultTextcolor. 
otherwise the light theme would not be work .
And please test the changes also in light theme 
Or better , Wait until I make a PR in the next days !

osodebailar:
That’s the solution. But in each case , you have to change the default styles.xml too and Test it . Otherwise the default Color in the light theme can also the gray Color

swissalpine:
Thanks for the hint! I will have a look then!

swissalpine:
Looking at the default styles.xml I find the lines:
```
<!---Text Colors -->
        <item name="android:textColorPrimary">@color/black</item>
        <item name="android:textColor">@color/black</item>
```
Proposal: Why not add this lines to the night theme? More contrast for the text is a benefit for many elder people like me.

osodebailar:
Thanks . The work seems to be done .
If you like it you can test the light mode in your mini phone and mail me your experience and the locations where we can improve the Color’s.

PR #1580, opened 2022-04-06
Chore remove redundant style

Andries-Smit:
After setting the new primary color the TextInputLayout border and hint can be removed
- reuse the notes layout in Wizard Dialog


osodebailar:
@Andries-Smit after removing , have you test the elements also in the light mode ?

PR #1592, opened 2022-04-09
color and theme cleaning

osodebailar:
- Begin with cleaning attributes , styles and colors not really needed. Not ready yet , will be continued.
- In addition fix again the gray text colors in overview info section ( now white colors again, because they are better readable )
- Working on light theme colors , build with the new material theme builder 

Open for  wishes or propsals for better coloring
@MilosKozak : I think we can start open theme switching and testing under developers

AAPS_Dark
AAPS_light


Philoul:
Among the little issues in light them:
- Deviation bars grey and black in dark have the same color in light theme.
- Sens curve is white in light theme.

osodebailar:
@Philoul the deviation bars are ok or have you another proposal ?
Yes the sens graph is white.
We can set background to the graph ( light gray ) or change the sense to a gray Color .
What do you think about it ?

osodebailar:
@Philoul change color to a gray in light theme for ratio

Philoul:
@osodebailar Sens curve is white in dark theme, I propose Black in light theme
=> I tried to change ratiocolor, but it only change Sens background in menu (sens curve is not linked to attr)
=> And Black color is not readable in menu because text is in black in light theme

PR #1600, opened 2022-04-10
style event date time to material input

Andries-Smit:
In the dialogs the Event time inputs are text views to set date and time, giving the materiel input bordered appearance, makes the users aware they are editable.
- Add border stroke
- Add click ripple effect

image


MilosKozak:
btw time picker has wrong colors

Andries-Smit:
@MilosKozak noticed the time picker color too, it is on my list

PR #1601, opened 2022-04-10
fix issue ratio color in light theme and make graph colors changable

osodebailar:
- ratio color in light theme and textcolor in popupmenu fixed and now configurable.
- fixed optical issue for background span.
@Philoul in additional graphs background is set to midgray for better visualisation.
- graph colors for treatments , devs, iob and so on, now  changeable  if theme is switched.
  In the moment color access by attribut not working here. 
@MilosKozak : after theme change colors will be changed after next workflow recalc. 
                          if we want it faster, we had to implement an additional event after theme switch.


app/src/main/java/info/nightscout/androidaps/workflow/PrepareBasalDataWorker.kt
```diff
@@ -116,27 +118,27 @@ class PrepareBasalDataWorker(
         // create series
         data.overviewData.baseBasalGraphSeries = LineGraphSeries(Array(baseBasalArray.size) { i -> baseBasalArray[i] }).also {
             it.isDrawBackground = true
-            it.backgroundColor = rh.gc(R.color.basebasal)
+            it.backgroundColor = ContextCompat.getColor(ctx, R.color.basebasal)
```
MilosKozak, line 121: why not `rh.gac()` here?

osodebailar: After 1 day of work i found the solution.
i had to fix gac function. Now it works also her in graph data.

osodebailar: > why not `rh.gac()` here?
@MilosKozak : fixed problem. ready for merge

PR #1607, opened 2022-04-11
Fix style date time dialog

Andries-Smit:
Styling Date and Time picker theme does not needed to be overwritten when using proper DayNight Time and correct colors.

Before, the theme of the dialog was not the day/night theme. (Implemented manually)

It fixes the wrong color of the time picker mentioned in https://github.com/nightscout/AndroidAPS/pull/1600#issuecomment-1094360416

Before, 
after

image

@osodebailar What is your perspective on this change? (Are we ok, with the accentColor in the light theme as header, as default from Android?)
The other question should the Date and Time ticker be converted to the Material TimePicker https://material.io/components/time-pickers/android#using-time-pickers

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/HistoryBrowseActivity.kt
@@ -169,7 +169,7 @@ class HistoryBrowseActivity : NoSplashAppCompatActivity() {
             val cal = Calendar.getInstance()
             cal.timeInMillis = overviewData.fromTime
             DatePickerDialog(
-                this, R.style.MaterialPickerTheme,
+                this,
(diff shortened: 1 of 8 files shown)


osodebailar:
I designed the light and dark timepicker as a proposal.
I am free, but i like the darker titlebar in the dark theme  looks like the other dialogs, we have no other dialog with the secondary color as titlebar color and the gray titlebar in the light theme because both of them looks more elegant.

The secondary color was automatic generated by the new themebuilder from google. I am free also about the color , but please not the 'crying green ' we have in the dark theme.

Andries-Smit:
I agree that your desinged headers looks better; however, this proposal removes complexity and sticks to the standard. and we might want to reconsider our color accent. I'll send you a PM to discuss this topic offline.

PR #1617, opened 2022-04-14
Feature statuslight opens caredialogs

osodebailar:
As a new feature to make the handling better the care dialogs for:
prime fill, sensor change, battery change can also be opened in overviewfragment.
So for normal operations only overview fragment is necessary.
The click behavior is the same as for the textfields top open the date/timepicker dialogs.

prime-fill

statuslight_click

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
@@ -207,6 +207,10 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
         binding.infoLayout.apsMode.setOnClickListener(this)
         binding.infoLayout.apsMode.setOnLongClickListener(this)
         binding.activeProfile.setOnLongClickListener(this)
+        binding.statusLightsLayout.cannulaOrPatchLayout.setOnClickListener(this)
+        binding.statusLightsLayout.insulinLayout.setOnClickListener(this)
+        binding.statusLightsLayout.sensorLayout.setOnClickListener(this)
+        binding.statusLightsLayout.batteryLayout.setOnClickListener(this)
     }
 
     @Synchronized
@@ -333,6 +337,7 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
         // https://stackoverflow.com/questions/14860239/checking-if-state-is-saved-before-committing-a-fragmenttransaction
         if (childFragmentManager.isStateSaved) return
         activity?.let { activity ->
+            val pump = activePlugin.activePump
             when (v.id) {
                 R.id.treatment_button    -> protectionCheck.queryProtection(
                     activity,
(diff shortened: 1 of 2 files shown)


Andries-Smit:
The feature could be helpful, though personally, I think we should not add too many click behaviours in the overview. With the change of accidental interaction. Actions could be more explicit and are used less often. The overview should be reserved for daily interactions.
Personally, the extra lines in the background make the user interface more chaotic.

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
```diff
@@ -425,6 +429,22 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
                         }.show(childFragmentManager, "Overview")
                     })
                 }
+
+                R.id.cannula_or_patch_layout -> {
+                    activity.let { protectionCheck.queryProtection(activity, ProtectionCheck.Protection.BOLUS, UIRunnable { FillDialog().show(childFragmentManager, "FillDialog") }) }
```
Andries-Smit, line 434: Patch pumps do not use the prime.

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
```diff
@@ -425,6 +429,22 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
...
+                }
+
+                R.id.insulin_layout -> {
+                    activity.let { protectionCheck.queryProtection(activity, ProtectionCheck.Protection.BOLUS, UIRunnable { FillDialog().show(childFragmentManager, "FillDialog") }) }
```
Andries-Smit, line 438: patch pumps do not fill, they only replace

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
```diff
@@ -425,6 +429,22 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
...
+                }
+
+                R.id.sensor_layout -> {
+                    CareDialog().setOptions(CareDialog.EventType.SENSOR_INSERT, R.string.careportal_cgmsensorinsert).show(childFragmentManager, "Actions")
+                }
+
+                R.id.battery_layout -> {
+                    CareDialog().setOptions(CareDialog.EventType.BATTERY_CHANGE, R.string.careportal_pumpbatterychange).show(childFragmentManager, "Actions")
```
Andries-Smit, line 446: patch pump dont change batteries

osodebailar:
The feature to handle all necessary actions on one page is one of the main features for the users of my own branch with my own surface .
We can remove your background border.

swissalpine:
I also don't see the benefit of offering click options for things that are not used several times a day in the Overview.
Especially on small devices, erroneous click actions are very easy to happen.

MilosKozak:
I personaly use only fill canula once in 3 days. and battery change once a month. the rest is handled automatically
I don't see huge benefit too. 
and border makes it less readable

osodebailar:
> 
ok. For me , i do not like to open the action page. But as you like

osodebailar:
> 
Small devices are special things ;-) 
But i repeat , for me its one of the main feature to handle all in one page.

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
osodebailar: add conditions now

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
osodebailar: add conditions

app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewFragment.kt
osodebailar: add conditions

Issue #1648, opened 2022-04-19
The last layout changes are inconvenient for small devices


swissalpine:
Yes, I use a smartphone with a small screen, but I think I'm not the only one. From there, you should be careful with layout changes that take up space and don't add functionality. That's what the changes Material Card View and Material Button do.
Maybe it is possible to change the margin-top for small devices to 0 dp in the Material Card View (I changed it from 4 dp to 1 dp so I can see the graph again), how to adjust the material buttons is beyond me. Please revert this changes!
New Matieralbuttons:
![Screenshot_20220419-154927](https://user-images.githubusercontent.com/12517693/164031188-c50ebcf9-469a-48b7-bbb6-cd2d614e1fa2.png)
![Screenshot_20220419-154943](https://user-images.githubusercontent.com/12517693/164031505-8349053c-3e7a-429e-a2f6-42ff732cbdf8.png)

----

Old buttons:
![Screenshot_20220419-155236](https://user-images.githubusercontent.com/12517693/164031404-4b7ac49d-bfc6-41b0-8a39-609f76489d3c.png)
![Screenshot_20220419-155134](https://user-images.githubusercontent.com/12517693/164031625-bfba30f6-a129-41e7-aa61-599d1034c507.png)

Please undo at least the material button commit!
And have an eye on the impact on small Screens as well for future layout changes (btw I liked the old layout!). Thanks!

osodebailar:
@swissalpine  We can change the style to smaller text , so it fits again .

Andries-Smit:
There are 2 types of users, users with a personal phone running AAPS, do have, in general, a large screen.
And users with a designated device, which is quite often as small as possible.

To cater for both, we can easily distinguish using qualified resources. and then we can tweak for smaller screens, for example, setting some padding/margins/corner radius to 0.

https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes

Philoul:
Even if it's not a small device, I think marging in material button should be reduced (close than half of the width is lost...)
![Screenshot_20220420-075324_AndroidAPS.jpg](https://user-images.githubusercontent.com/52934600/164160640-cef909c7-a814-4c76-a099-fd68f6438f20.jpg)

osodebailar:
[...]

PR #1649, opened 2022-04-19
small font for material button

osodebailar:
@swissalpine : Please test this on your mini phone

Issue: https://github.com/nightscout/AndroidAPS/issues/1648

### DIFF
--- a/app/src/main/res/layout/dialog_loop.xml
+++ b/app/src/main/res/layout/dialog_loop.xml
@@ -82,7 +82,6 @@
                     android:drawableTop="@drawable/ic_loop_closed"
                     android:text="@string/closedloop"
                     android:textAppearance="@style/TextAppearance.AppCompat.Medium"
-                    android:textSize="11sp"
                     android:visibility="gone" />
 
                 <info.nightscout.androidaps.utils.ui.SingleClickButton
@@ -97,8 +96,6 @@
                     android:paddingEnd="10dp"
                     android:drawableTop="@drawable/ic_loop_lgs"
                     android:text="@string/lowglucosesuspend"
-                    android:textAppearance="@style/TextAppearance.AppCompat.Medium"
-                    android:textSize="11sp"
                     android:visibility="gone" />
 
                 <info.nightscout.androidaps.utils.ui.SingleClickButton
@@ -113,8 +110,6 @@
                     android:paddingEnd="10dp"
                     android:drawableTop="@drawable/ic_loop_open"
                     android:text="@string/openloop"
-                    android:textAppearance="@style/TextAppearance.AppCompat.Medium"
-                    android:textSize="11sp"
                     android:visibility="gone" />
 
                 <info.nightscout.androidaps.utils.ui.SingleClickButton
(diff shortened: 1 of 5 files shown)


swissalpine:
Thank you for taking the point.
Unfortunately the font size does not change after merging your branch. The problem still exists.

osodebailar:
I set the font size from 10 to 8sp and the padding to 0. maybe we can try „dp“ instead of „sp“

swissalpine:
Yes, that is probably the reason. I actually changed to dp almost everywhere because my system-wide large font size destroys any layout otherwise.

osodebailar:
> Yes, that is probably the reason. I actually changed to dp almost everywhere because my system-wide large font size destroys any layout otherwise.

@swissalpine  Lets try again . change to dp and make letter spacing smaller

swissalpine:
Thank you for your continued work.
There are still two lines too many in dialog_loop.xml which should be deleted (300-301):
```
android:textAppearance="@style/TextAppearance.AppCompat.Medium"
android:textSize="11sp"
```
Beside it fits _for me_, if the following changes in the styles.xml are still added:
The font can be 12 dp (10dp is already too small ...), maybe <... padding-top>0dp</item> is another possibility to get the buttons a bit smaller. But it would be even better if the distance between icon and text would be smaller, at least that's how it was before. But I don't know how to do that ...
Here is my excerpt from the style.xml:
```
    <style name="ButtonSmallFontStyle" parent="GrayButton">
        <item name="android:letterSpacing">0</item>
        <item name="android:paddingTop">0dp</item>
        <item name="android:paddingStart">0dp</item>
        <item name="android:paddingEnd">0dp</item>
        <item name="android:textSize">12dp</item>
    </style>
```
What do you think?

osodebailar:
> Thank you for your continued work. There are still two lines too many in dialog_loop.xml which should be deleted (300-301):
> 
> ```
> android:textAppearance="@style/TextAppearance.AppCompat.Medium"
> android:textSize="11sp"
> ```
> 
> Beside it fits _for me_, if the following changes in the styles.xml are still added: The font can be 12 dp (10dp is already too small ...), maybe <... padding-top>0dp is another possibility to get the buttons a bit smaller. But it would be even better if the distance between icon and text would be smaller, at least that's how it was before. But I don't know how to do that ... Here is my excerpt from the style.xml:
> 
> ```
>     <style name="ButtonSmallFontStyle" parent="GrayButton">
>         <item name="android:letterSpacing">0</item>
>         <item name="android:paddingTop">0dp</item>
>         <item name="android:paddingStart">0dp</item>
>         <item name="android:paddingEnd">0dp</item>
>         <item name="android:textSize">12dp</item>
>     </style>
> ```
> 
> What do you think?

Thanks. 
I removed the lines that i overlooked add top padding and set the fontsize to 11dp. For longer texts ( quickwizard button ) its better to act with 11 instead of 12.

@swissalpine : can you test please again  and say ok for merging ?

swissalpine:
No, it all fits back on the pages, that's really o.k. I think - but reading the writing is difficult for an old man like me. But better this way than the other.
Quickwizzard I don't use, so 12 dp would have fit for me, but I think this word blows everything up.
Here are the current screenshots:
Screenshot_20220420-181224
Screenshot_20220420-181211
Thanks for finding a solution!

osodebailar:
@swissalpine thanks for testing .
@MilosKozak we can merge the solution.

PR #1652, opened 2022-04-20
Fix cardview and optimze for small displays

osodebailar:
Due to the change to material buttons the former margin workaround for the singleclick buttons to fit for small displays can be
reduced.
Card for the loop button removed > no extra cards for actions.
Card in overview was darken a little bit in overview for better optic. If its ok, we darken the rest cardviews (only in dark mode).

overview


MilosKozak:
there is a conflict. still valid?

osodebailar:
> there is a conflict. still valid?

@MilosKozak : conflict solved. its valid and tested

PR #1665, opened 2022-04-22
optimize overview buttons layout

osodebailar:
The sytling of overview buttons have many disadavantages. 
On of these occures when the text is to long. 
I optimize layout , except Quickwizard, with ellipsis in the end if text s to long.
Additional to make the buttons look better and spare space the icon padding is reduced.
After optimizatrion the look and feel is more aesthetic and professional

Overview buttons now :+1: 
overview_buttons_now

Overview buttons after optimization:
overview_buttons_ellipsis_reduced_padding


Philoul:
This update is great.
Many thanks!

osodebailar:
@Philoul thank you very very much.

PR #1666, opened 2022-04-22
Wear

MilosKozak:


Philoul:
Do you keep compatibility with old version of wear aaps, or is it necessary to update wear app if aaps plugin is updated?

MilosKozak:
> Do you keep compatibility with old version of wear aaps, or is it necessary to update wear app if aaps plugin is updated?

most of the stuff is new or reactored
communication is different (primary reason for refactoring along with unify the code with AAPS)

so NO

PR #1671, opened 2022-04-24
Feature add swipe refresh for loop and APS fragments

osodebailar:
To save space and to have a better look on the loop and all APS Fragments buttons
to re run calculation are removed and a swipefrefreshlayout was added.
You can see the swipefresh also in apps like google drive.

For users that have not the ability to swipe additional an option menu entry is created.
Text entry is the same as in fromer button

see also:
https://developer.android.com/training/swipe/add-swipe-interface
APS_ISF
APS_ISF_menu
APS_ISF_swipe


MilosKozak:
works nice 👍

PR #1674, opened 2022-04-25
Adding the ability to email exported preferences

erobinson:


MilosKozak:
the problem with this is, that it only opens messaging app and not send an email automaticaly

MilosKozak:
the problem with this is, that it only opens messaging app and not send an email automaticaly

erobinson:
> the problem with this is, that it only opens messaging app and not send an email automaticaly

yeah, it's definitely not perfect or ideal, but better than the current implementation right? in that it makes it much easier for end-users to backup the file off the phone.

lee-b:
I think it would be much better to regularly and automatically save to a user-selected folder.  That way users can automatically backup that folder to an SDCard or remote service, or just sync it automatically using Dropbox/Google Drive/Onedrive/Resilio/Syncthing or set up their own regular task to email it automatically, without prompts.

mordechi-david:
I did with a friend a version of the smartphone just for the loop to suit even kids whose parents do not want to have a smartphone, and there we made an app that zips the preferences folder to zip and sends by email but we did it in a codolar with almost no code but it is possible

Issue #1597, opened 2022-04-10
Overlapping Times in basal Tab leading to wrong iob? Dev-i


Ede1975:
It seems tbrs have not been cancelled, Times are overlapping in basal tab
Possibly aaps was out of reach of dash a couple of Times during the period from 14:53 to 16:53

![IMG-20220410-WA0000.jpg](https://user-images.githubusercontent.com/88888180/162627614-4d46bc2e-aa0b-454b-a5bb-f0dc0b09acf4.jpg)

![AndroidAPS_LOG_1649605049645.log.zip](https://github.com/nightscout/AndroidAPS/files/8459360/AndroidAPS_LOG_1649605049645.log.zip)

MilosKozak:
is it possible you changed pump meanwhile?

Ede1975:
You mean from dash to some other pump or virtual pump? No, definately not.

MilosKozak:
do you have logs prior this one?

MilosKozak:
note: where is this from
`17:02:30.229 [RxCachedThreadScheduler-53] D/DATABASE: [PumpSyncImplementation.syncTemporaryBasalWithPumpId():274]: Updated TemporaryBasal TemporaryBasal(id=28452, version=3, dateCreated=1649602950162, isValid=true, referenceId=null, interfaceIDs_backing=InterfaceIDs(nightscoutSystemId=null, nightscoutId=6251f4b9215b0200041e4d4c, pumpType=OMNIPOD_DASH, pumpSerial=4241, temporaryId=null, pumpId=-8976659072097442112, startId=null, endId=-7142919318749436032), timestamp=1649602949159, utcOffset=7200000, type=NORMAL, isAbsolute=true, rate=1.5, duration=1800000)
`

Ede1975:
> do you have logs prior this one?

No, unfortunately not. I am still on the same Version, did not happen again. In case it occurs again i will add new logs.

emmatovar27:
Hi There! I have the same issue. This happens every time the TBR is set for 30 min. The rest of TBR looks to work fine.
![Screenshot_20220419-175209](https://user-images.githubusercontent.com/39507448/164126009-186010bd-c76c-4cb3-a94e-5c3c46640a10.png)
[AndroidAPS_LOG_1650416536289log.zip](https://github.com/nightscout/AndroidAPS/files/8517378/AndroidAPS_LOG_1650416536289.log.zip)

 Let me know if there is anything additional information is need. 

[...]

PR #1715, opened 2022-05-09
Fix Dash historyIds

avereha:
Before this change, we used `ULID` for history Ids.
For `pumpId`, we would take the entropy part of the history id.

It turns out that on some phones the default Random() implementation does not have enough entropy and we are getting duplicate values.
Should fix https://github.com/nightscout/AndroidAPS/issues/1597

After update, the Dash history will be empty.


jbr7rr:
Will this fix https://github.com/nightscout/AndroidAPS/issues/1597 ? Or is this a separate issue? I'm currently looking into that one but have been unable to pinpoint the exact issue there.

avereha:
@jbr7rr : yes, I think so!

vanelsberg:
> It turns out that on some phones the default Random() implementation does not have enough entropy 

Wow. Nice find Andrei.
Will test!

vanelsberg:
> After update, the Dash history will be empty.

@avereha Also lost the active Pod (which was beyond 72 hours anyway)
Could that be or is this "unexpected"..

avereha:
This is very possible!
I changed the Dash configuration.

Issue #1726, opened 2022-05-12
Wear OS DigitalStyle watchface time is missing minutes


tswire:
On the latest 3.0.0.2 dev k build 49163c43d, the digitalstyle watch face seems to have font issues (too big when set to AM/PM vs 24 hr) or missing info in the time.  See pic below.
![Screenshot_20220512_100432_sysui](https://user-images.githubusercontent.com/30097654/168147056-3c808384-2596-495c-bc05-50606d010bcb.png)

vanelsberg:
Using 24 hrs format. For me is shows ok

tswire:
Must be the AM PM view only then. Hopefully someone can look at it as all these wear os faces were working perfectly before 3.0.0.2 build

jgslade:
I've also got the big font for the am/pm on digital style.
![Screenshot_20220514_115706_sysui](https://user-images.githubusercontent.com/29897815/168443386-0bfbee0a-c7ce-4af3-899c-76ba212336a9.png)

jgslade:
I was looking at the code to see how it all works (duct know Java/kotlin but have done some c, python, html, and sql).
I adjusted the font sizes in DigitalStyleWatchface.kt and adjusted the layout a bit on round/activity_digitalstyle.xml. It looks like the am/pm info is being passed with the hour info. 

![Screenshot_20220514_164119_sysui](https://user-images.githubusercontent.com/29897815/168450459-a5b1d22c-1899-4607-a1b1-4d0d3af82bfa.png)

jgslade:
I think I found a fix:

If this class
```
fun hourString(mills: Long): String {
        var format = "hha"
        if (android.text.format.DateFormat.is24HourFormat(context)) {
            format = "HH"
        }
        return DateTime(mills).toString(DateTimeFormat.forPattern(format))
    }
```

in
info/nightscout/androidaps/utils/DateUtil.kt

is changed to 
```
fun hourString(mills: Long): String {
        var format = "hh"
        if (android.text.format.DateFormat.is24HourFormat(context)) {
            format = "HH"
        }
        return DateTime(mills).toString(DateTimeFormat.forPattern(format))
    }
```

The display looks correct.
![Screenshot_20220514_174908_sysui png](https://user-images.githubusercontent.com/29897815/168451636-5de61a93-5df0-4394-a42b-a3717be8faa3.jpg)

tswire:
> I think I found a fix:
> 
> If this class
> 
> ```
[...]

PR #1737, opened 2022-05-15
Fix for Wear OS DigitalStyle watchface time is missing minutes #1726 

jgslade:
This fixes the display issue on the digital style watchface. I haven't noticed any regressions.


MilosKozak:
there should be an extra element for AM/PM 
image

PR #1750, opened 2022-05-19
Update to DynamicISF and addition of user settab;e low glucose suspend value, as found in Loop

tim2000s:


MilosKozak:
default value of LGS is 60 but allowed range is 65-100

MilosKozak:
@AdrianLxM I'm not sure we should change other algs. your opinion?

tim2000s:
FOund a bug when running this morning.

The default value of LGS is supposed to trigger standard behaviour. Users enter 65-100.
Standard behaviour of LGS threshold is to be min_bg-0.5*(min_bg-40). At glucose levels below 100, this creates a low glucose suspend level that's below 70, which I think a few people think is too low. 

However I also realised that the PR misses the logic to set the threshold to use standard logic for threshold when I merged it, so I'm closing the PR for now.

MilosKozak:
What I mean is when you allow range 65-100 user cannot revert to 60 once changed default value

PR #1753, opened 2022-05-20
Fix lint layout warnings

Andries-Smit:
Fix various lint warning
- hardcoded text
- moved the test to the string file for translations
- Improve RTL in tables (loop, dana, combo etc.)
- removed extra nested layouts that were not required.
- Formatted the edited files

The change is easier to review in "spit" mode ignore whitespace
https://github.com/nightscout/AndroidAPS/pull/1753/files?diff=split&w=1

### DIFF
--- a/app/src/main/res/layout/bgsource_item.xml
+++ b/app/src/main/res/layout/bgsource_item.xml
@@ -45,6 +45,7 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:paddingStart="10dp"
+                android:paddingEnd="0dp"
                 android:textStyle="bold"
                 tools:text="Name" />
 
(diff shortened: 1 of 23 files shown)


app/src/main/res/layout/bgsource_item.xml
```diff
@@ -39,52 +37,52 @@
...
+                tools:text="NS" />
 
             <TextView
                 android:id="@+id/invalid"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:paddingStart="10dp"
-                android:text="@string/invalid"
-                android:textColor="?attr/alarmColor" />
+                android:textColor="?attr/alarmColor"
```
MilosKozak, line 72: see this https://github.com/nightscout/AndroidAPS/commit/3d729820638f807f311e51e480f2c04a45ae23a0
tools:text is displayed only in IDE but not in app

Andries-Smit: Indeed, your right. In this cases it should not just in the IDE. (same for NS label). Will fix it

Andries-Smit: I see now how I broke it last time.

PR #1755, opened 2022-05-20
Update DynamicISF and add in user configurable low glucose suspend

tim2000s:
Dynamic ISF update includes the new model based on insulin peak time.

The update to low glucose suspend threshold replaces the threshold value with a user specified one only when that value is greater than the threshold defined by the algorithm.

By default, it uses the built in low glucose suspend calculation.

This allows users to set a lower target and have a higher LGS threshold, ie. when you set 90 as your target, oref1 would normally set a 65 lgs value. With this, users can have a 90 target and an 80 lgs threshold, which is safer.

This is applies across all three algorithms and is a safety enhancement.


MilosKozak:
@AdrianLxM I'm not sure we should change other algs. your opinion?

MilosKozak:
default value of LGS is 60 but allowed range is 65-100
when you allow range 65-100 user cannot revert to 60 once changed default value

tim2000s:
> default value of LGS is 60 but allowed range is 65-100 when you allow range 65-100 user cannot revert to 60 once changed default value

It's actually a soft limit. Users can enter values below 65, and if they do, it will revert to the standard threshold model in oref1. Similarly, if theyvspacify a value that's too high, it reverts to the standard threshold model.

tim2000s:
I'd also add that you're not really changing anything major about the logic of the decision making process. You're simply allowing a user to set the log glucose threshold to kick in at a higher value than it would normally, especially where a low target is in use. Any temp target adjustments that would set a higher threshold will still do so, if the threshold is higher than that of the user. 

Initial user feedback is that this does help reduce low glucose events when using the OpenAPS SMB code.

app/src/main/assets/OpenAPSAMA/determine-basal.js
```diff
@@ -182,9 +182,21 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
...
 
-    // min_bg of 90 -> threshold of 70, 110 -> 80, and 130 -> 90
-    var threshold = min_bg - 0.5*(min_bg-50);
+    // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85, or if specified by user, take that value
+ var lgsThreshold = profile.lgsThreshold;
+        var threshold = min_bg - 0.5*(min_bg-40);
+        var oldThreshold = threshold;
 
+        if(lgsThreshold < 65 || lgsThreshold > 120) {
+            threshold = threshold;
```
MilosKozak, line 191: this code is strange

app/src/main/assets/OpenAPSAMA/determine-basal.js
```diff
@@ -182,9 +182,21 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         return rT;
     }
 
-    // min_bg of 90 -> threshold of 70, 110 -> 80, and 130 -> 90
-    var threshold = min_bg - 0.5*(min_bg-50);
+    // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85, or if specified by user, take that value
+ var lgsThreshold = profile.lgsThreshold;
+        var threshold = min_bg - 0.5*(min_bg-40);
```
MilosKozak, line 187: why is default formula changed?

tim2000s: Due to a typo. It shouldn't be.

MilosKozak:
```
        <info.nightscout.androidaps.utils.textValidator.ValidatingEditTextPreference
            android:defaultValue="60"
            android:inputType="numberDecimal"
            android:dialogMessage="DDDDDDDD"
            android:key="DDDDDDDDDD"
            android:title="DDDDDDDDDDDDDDDDDDD"
            validate:maxMgdl="100"
            validate:minMgdl="60"
            validate:testType="bgRange" />
````
you need to merge dev back to make it working
https://github.com/nightscout/AndroidAPS/commit/6e7b5ab7dbebb98d53e662ab866de4a31568c32b

app/src/main/assets/OpenAPSAMA/determine-basal.js
tim2000s: I've checked back against openaps/oref0 master, and the min_bg-50 value is incorrect. In openAPS it's min_bg-40. sp the deletion is where the typo is.

tswire:
Any update?

PR #1761, opened 2022-05-22
Fix regression in medtronic cannula change detection

p5nbTgip0r:
It seems an erroneous `lastPrimeRecordTime!=0L && ` was added during the kotlin refactorings of this:
https://github.com/nightscout/AndroidAPS/commit/59a3bf88834a8433676a78e893ee5694166f0949#diff-b3c8866292a4a902f69ba7dc1b82892bae6fc93a2dce97749c34f1fcf9f1f751R400
(you'll likely have to scroll down to `MedtronicHistoryData.kt` and expand it to see the linked change due to github collapsing large diffs)

This PR just reverts that line, which restores the previously-working functionality of automatic cannula change detection in medtronic pumps.


MilosKozak:
@andyrozman

andyrozman (commented):
I am not sure this is correct... What is lastPrimeRecordTime is 0.

andyrozman (approved):
It seems that error was made with refactoring and this should fix it...

PR #1777, opened 2022-05-28
Communication moved from Gitter to Discord

kaindl:
As of last comment in Gitter-Channel communication moved to Discord, but in CONTRIBUTING and in the exams the Gitter-Channel is still referenced.


MilosKozak:
PR must be dev-based otherwise there will be conflicts like in this case
Change only EN verison. The rest will be updated via Crowdin

PR #1806, opened 2022-06-04
Fix Autotune User Entry if launched by automation rule

Philoul:
Replace Autotune Source by Automation Source (+ add note)


Philoul:
Wrong target, sorry

PR #1811, opened 2022-06-06
Autotune UI Improvement

Philoul:
Save last run results (PumpProfile, TunedProfile and settings)
Load Last Run Result after aaps restart
Add a link on LastRun Date to load last result


MilosKozak:
there is a conflict

PR #1817, opened 2022-06-08
Autotune feature : Tune Specific days of the week

Philoul:
I update UI and algo to be able to tune specific days of the week (for example Weekend days versus Working days)
Note, the development of this feature was started on OAPS project but never finished, so I don't have the equivalent in OpenAPS to compare results...

After several months of tests, I noticed that it is necessary to increase a lot the number of selected days (minimum 21 or 28 days selected for tuning "Weekend days" to get smooth tuning results calculated with only 6 or 8 remaining days...).

image


Philoul:
No, I don't think so, for ex I tune "training days" that are mondays and wednesdays (sport training between 19h and 20h30), so it's not a range.
The other profiles are "working days" are tuesdays, thurdays and fridays (I also work on training days 🤪), and "weekend days" so day selection is more flexible I think.
And for autotune, the idea is to tune the more recent days to adjust profile. That's why in user interface we only select a number of days back (without this pr it's all days up to yesterday, like autotune web)

Philoul:
I'm testing this version for more than one month now and it works good.
Same days selected for automation rule and autotune action, then I automatically have my profile tuned with selected days and applied.

PR #1830, opened 2022-06-13
Updated Dynamic ISF 

tim2000s:
This push consists of the following changes:

1. DYnamic ISF Plugin uses combination of 8 hour weighted average, last full day and 7 day average TDDs.
2. 2. TDD calculator adjusted to provide this information.
3. Model is the reworked 1800 rule using log values and insulin peak.

LGS Threshold moved in preferences to the DynamicISF prefs module and only available in the Dynamic ISF plugin.


tim2000s:
Wrong version of Dev.

PR #1831, opened 2022-06-13
Dynamic ISF update to model

tim2000s:
This push consists of the following changes:

DYnamic ISF Plugin uses combination of 8 hour weighted average, last full day and 7 day average TDDs.

TDD calculator adjusted to provide this information.
Model is the reworked 1800 rule using log values and insulin peak.

LGS Threshold moved in preferences to the DynamicISF prefs module and only available in the Dynamic ISF plugin.


app/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
```diff
@@ -501,9 +510,41 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
...
+            lgsThreshold = (18 * lgsThreshold);
+            lgsThreshold = round(lgsThreshold, 2);
+            console.error("LGS Threshold converted to"+lgsThreshold+"; ");
+        }else{
+            lgsThreshold;
+        }
+        var threshold = min_bg - 0.5*(min_bg-40);
+        var oldThreshold = threshold;
+        if(lgsThreshold < 65 || lgsThreshold > 120) {
+            threshold;
```
MilosKozak, line 527: It does nothing

app/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
```diff
@@ -501,9 +510,41 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
...
+            }
+        else if( lgsThreshold < threshold){
+            threshold;
```
MilosKozak, line 530: It does nothing

app/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
```diff
@@ -501,9 +510,41 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         return rT;
     }
 
-    // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85
-    var threshold = min_bg - 0.5*(min_bg-40);
+    // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85, or if specified by user, take that value
+        var lgsThreshold = profile.lgsThreshold;
+        console.error("Profile LGS Threshold set to"+lgsThreshold+"; ");
+        //if(lgsThreshold < 5.6)
+        if(lgsThreshold > 3.2 && lgsThreshold < 5.6){
```
MilosKozak, line 517: All JS code code works in mgdl. I believe we must continue with this assumption and do conversion in kotlin

app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
```diff
@@ -161,6 +161,102 @@ class TddCalculator @Inject constructor(
...
+        tdd.totalAmount = tdd.bolusAmount + tdd.basalAmount
+        //}
+
+
+        aapsLogger.debug(LTag.CORE, tdd.toString())
+        return tdd
+    }
+
+
+    fun calculate8Gap():TotalDailyDose {
```
MilosKozak, line 213: this is copy-paste
replace with 1 function with params

app/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
tim2000s: Yes. If the user LGS threshold is outside the limits that we allow, including the calculated threshold, we leave the calculated threshold. It's only when user defined value is higher than cacluated value, that we replace the calculated value.

app/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
tim2000s: I've tried that. The conversion doesn't pick up using the Kotlin code. It only changes the display value and doesn't get passed to determine basal. 

Happy to have it in Kotlin, and would prefer it that way, but so far haven't been able to make it work.

app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
tim2000s: If you allow this to be used, then I can try and work that out, but right now that will take me some time, and I'd like to get the updated model into wider circulation.

MilosKozak:
my cleanup https://github.com/nightscout/AndroidAPS/pull/1832/commits/2d6de1970b637222ce1951864748f520cf7cee72

PR #1839, opened 2022-06-18
Add Unit test for AutotuneCore

Philoul:
Unit Test for PreppedGlucose class added to test correct import of OpenAps Autotune categorization file

Unit test (passed) for AutotuneCore file
- files of oref0-autotune run are included in res/autotune/test?
- Information concerning oref0 command line and input files in Oref0-Autotune_Command.txt

Still remain unit test for AutotunePrep file (the most complicated...) I propose to do it in another PR


MilosKozak:
One test is failing in automated build

Philoul:
What is "automated build" and what should I do to fix this issue ?
When I run unit tests on my computer, everything works fine I think (120 of 120 tests passed)
Is it due to the errors I have on 3 key strings in `AutotuneCoreTest.kt` file (see screenshot below)? I think it's because these 3 strings are in core module and not in app module, but during the run, the build is Ok and it works as expected... (and I don't know how to fix the "red" key strings...)

image

MilosKozak:
I agree it's working fine localy
image
image
it's hard to debug. add logging. the problem is usually parallelism

Philoul:
It fails during `Run tests with max tries of 2`
here (it fails on both tries)
```
<============-> 99% EXECUTING [52s]> :app:testFullDebugUnitTest > Executing test info.nightscout.androidaps.plugin> :app:testFullDebugUnitTest > Executing test info...signatureVerifier.Signatur> :app:testFullDebugUnitTest > 8 tests completed> :app:testFullDebugUnitTest > Executing test info.nightscout...general.autotun<============-> 99% EXECUTING [53s]<============-> 99% EXECUTING [54s]
info.nightscout.androidaps.plugins.general.autotune.AutotuneCoreTest > autotuneCoreTest FAILED
    java.lang.AssertionError: expected:<1.002> but was:<1.112>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:555)
        at org.junit.Assert.assertEquals(Assert.java:685)
        at info.nightscout.androidaps.plugins.general.autotune.AutotuneCoreTest.autotuneCoreTest(AutotuneCoreTest.kt:61)
```
It fails on first verification of basal rate in the loop `for (i in 0..23)` below (`basal[0]` is wrong)
=> AutotuneCore is "pure algo" with no external call other than preferences, and with the same input, it should allways return the same output...

Here `isf` and `ic` verification passed just before basal verification, so in `prep` that signify that `crData`, `csfGlucoseData` (used for ic calculation)  and `isfGlucoseData` (for isf) and Input Profile (ic / isf) are correct.

```
        val outProfile = autotuneCore.tuneAllTheThings(prep, inputProfile, inputProfile)
        OapsOutputProfile?.let {
            Assert.assertEquals(OapsOutputProfile.isf, outProfile.isf, 0.0)
            Assert.assertEquals(OapsOutputProfile.ic, outProfile.ic, 0.0)
            for (i in 0..23)
                Assert.assertEquals(OapsOutputProfile.basal[i], outProfile.basal[i], 0.0)
        }
```

My only idea concern usage, in `autotuneCore` algo, during basal calculation of:
```
val BGTime = Calendar.getInstance()
...
val myHour = BGTime.get(Calendar.HOUR_OF_DAY)
```
=> Could it be a bad conversion between UTC and local hour during unit tests (Unit tests build with local time = UTC + 2h) ? (but I don't know what to do to fix it in unit test if it's the reason...)

@MilosKozak any idea?

MilosKozak:
yes, common isssue as well.
Instead of `Calendar` use dateUtil.now() which is easy to mock

Philoul:
In AutotuneCore calculation must be done in localtime (from 4AM to 4AM the day after), 
(Calendar is used to get local hour of a basalGlucose date)
So I propose to fix it with
```
TimeZone.setDefault(TimeZone.getTimeZone("GMT+2"))
```
included in UnitTest

MilosKozak:
👍

PR #1842, opened 2022-06-20
fix #1835 : missing`notes` field for carbs entry

### LINKED ISSUE #1835: bug: Notes field not delivered to nightscout (at leats for carb correnctions)
**Steps to reproduce:**

- open `carbs` dialog, 
- enter carbs amount, 
- enter `test note` text in `Notes` field
- press `ok`, then `ok` to send
- wait event to be delivered to NS
- open nightscout -> reports -> treatments. there will be your new `carb correction` (or `meal bolus`) treatment with empty `Notes` column
image
in nightscout's db field `notes` is empty, like if it wasn't even sent to nighscout

**Expected behavior**
`test note` text in `Notes` field

**versions used**
Checked in nsclient 3.0.0.2 (downloaded from releases in github) and aaps 3.0.0.2 (built myself).
Nightscout 14.2.5

additional screenshots:
(this is add carbs screen from aaps, but in russian, sorry :))
image
note correctly shown in treatments (but still missing in nightscout)
image

this is 2.8.2 nsclient behavior:
image

image

justmara:
fixes #1835

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/activities/fragments/TreatmentsBolusCarbsFragment.kt
+++ b/app/src/main/java/info/nightscout/androidaps/activities/fragments/TreatmentsBolusCarbsFragment.kt
@@ -268,6 +268,10 @@ class TreatmentsBolusCarbsFragment : DaggerFragment() {
             }
 
             holder.binding.calculation.tag = ml
+
+            var notes = ml.carbs?.notes ?: ml.bolus?.notes ?: ""
+            holder.binding.notes.text = notes
+            holder.binding.notes.visibility = if (notes != "") View.VISIBLE else View.GONE
         }
 
         override fun getItemCount() = mealLinks.size
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/nsclient/NSClientAddUpdateWorker.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/nsclient/NSClientAddUpdateWorker.kt
@@ -88,7 +88,7 @@ class NSClientAddUpdateWorker(
                             .also { result ->
                                 result.inserted.forEach {
                                     uel.log(
-                                        Action.BOLUS, Sources.NSClient,
+                                        Action.BOLUS, Sources.NSClient, it.notes,
                                         ValueWithUnit.Timestamp(it.timestamp),
                                         ValueWithUnit.Insulin(it.amount)
                                     )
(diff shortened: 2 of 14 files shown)


MilosKozak:
it is not so simple. this will clear Carbs table

justmara:
sorry, im not an android dev :)
how to add migration then? or someone can use this as a base and simply add migration? I've checked it and it works fine. don't know about db clear - it is refilled from NS anyway.

justmara:
@MilosKozak better now? :)

justmara:
`versionCode` wasn't changed even on 2.8->3 upgrade, why it is required here? And how could it lead to crashes on downgrade? It's an extra field in database, not the missing one :) Older versions wont use it overall. It is nullable, so wont even break writes to this table by older apps.

okay, added same for boluses :)

MilosKozak:
> `versionCode` wasn't changed even on 2.8->3 upgrade, why it is required here? And how could these changes lead to crashes on downgrade? It's an extra field in database, not the missing one :) Older versions wont use it overall. It is nullable, so wont even break writes to this table by older apps.
> 
> okay, added same for boluses :)

ad) 2.8 -> 3.0 database was changed thus it's still possible to use old one (without data migration)
on start app is verifying database version and unchanged db schema. test it yourself

MilosKozak:
i believe this is from you:
image

justmara:
>i believe this is from you:

It must be my nightly experiments :) there was moment when I've built intermediate version with db changes in `21`, and then installed newer one with fix for bolus too, but with same db version. And now it. Must be migration that fails because of column already present in table. Since Room database is persisted even after uninstall of app - I cannot get rid of it to check for clean update.

justmara:
Now I'm running nsclient built with all these changes without issues. Sorry, but don't know how to check the clean update process.

justmara:
I've found also weird behavior with custom index that were resulting in crashes on update. Now I've passed update from clean 3.0.0.2 built from master to the one built from my branch in emulator without any issues.

justmara:
final touch: added notes to treatmens lists - cabrs/bolus and user-entry
image

justmara:
... and rebased on dev to fix conflicts

justmara:
this is how it looks merged to beta1 :)
image
image

MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/pull/1999

PR #1845, opened 2022-06-21
Wear mute alarm

Andries-Smit:
As a user, I want to mute the alert sounds with the Wear OS smartwatch app. 
image

The user can bind a hardware button to the Quick Snooze Activity. 
> Watch 4: Settings -> Advanced features  -> Customize keys -> Double press -> Select AAPS Snooze Alert

The activity tied to the button alarms can be muted without looking at your phone or watch.
Similar to the xDrip QuickSnooze function. https://github.com/NightscoutFoundation/xDrip/blob/master/wear/src/main/java/com/eveningoutpost/dexdrip/QuickSnooze.java

When xDrip is installed with the QuickSnooze activity off xDrip is called snoozed too, so user can mute both xDrip and AAPS alarms.

The snooze activity only mutes the sound but does not close the dialog on the phone. At a later stage, the user can see and act on the issue.

A follow-up story on the feature; show the alert as a notification similar to bolus progress with a snooze button so it can be handled through the display as well. (or for people that don't have a hardware key available)

@Philoul Could you create a nice app icon for this?


Philoul:
Hi @Andries-Smit 
Yes, I'll take a look and propose something 😉

wear/src/main/java/info/nightscout/androidaps/interaction/actions/QuickSnoozeActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+                Log.e("WEAR", "failed to snooze xDrip: ", e)
+            }
+        } else {
+            Log.d("WEAR", "Package $xDripPackageName not available for snooze")
+        }
+
+        finish()
+    }
+
+    private fun isPackageExisted(targetPackage: String?): Boolean {
```
MilosKozak, line 44: remove ? and !!

Sjoerd-Bo3:
@Andries-Smit How do you enable the quicksnooze for xdrip+ via the AAPS snooze alert?

Andries-Smit:
@Sjoerd-Bo3 you need to have xDrip also installed in your watch
See: https://github.com/jamorham/xDrip-plus/blob/master/Documentation/WatchGuide.md#enable-xdrip-android-wear-integration

Sjoerd-Bo3:
> @Sjoerd-Bo3 you need to have xDrip also installed in your watch See: https://github.com/jamorham/xDrip-plus/blob/master/Documentation/WatchGuide.md#enable-xdrip-android-wear-integration

Aah didn't know that, it works like a charm now!

Sjoerd-Bo3:
@Andries-Smit Is it also an option to let the snooze also silence the Omnipod Dash alarms? That would be really phenomenal

PR #1851, opened 2022-06-24
Add Unit test for AutotunePrep

Philoul:
Add Unit test for `AutotunePrep` to test correct categorization (the most complicated part of Autotune Algo is in this file)
To be able to compare categorization, I took from OpenAPS log file the exact iob/activity calculation (values available in OpenAPS autotune log file, I used an excel file to create a json for unit test)
`AutotuneIob` is replaced by `TestAutotuneIob` with an override of `getIOB(...)` function to replace aaps iob calculation by OpenAPS iob calculation
autotuneIob.boluses is also built with OpenAPS categorization (crData only) (`Dosed(...)` function included in AutotunePrep sum all boluses between start and end perdiods of crdata, but this sum is depends on the calculation of iob (especially for tbr splitted differently between AAPS and OpenAPS)
=> For tests I create only one bolus at end Timestamp of each crdata period with total dosed calculated by OpenAPS (so correct value is only included in unit test only if correct categorization is calculated)

This unit test verify that aaps algo works exactly the same way than OpenAPS for categorization of all data


MilosKozak:
👍

PR #1861, opened 2022-06-28
Wear multiple plus buttons

Andries-Smit:
Wear OS input for carbs and insulin can be improved by providing multiple plus buttons. The values of the plus buttons are based on the settings that are used in the Carbs and Insuline dialog of the phone.

image


MilosKozak:
ready?

Issue #1862, opened 2022-06-28
Wear - eCarbs start time recorded in hours instead of minutes


jgslade:
Here I am trying to start the eCarbs in 15 minutes.
![photo_2022-06-28_09-00-30](https://user-images.githubusercontent.com/29897815/176212919-c317be93-e515-4db3-936c-2542d8ebd14c.jpg)

Here it's telling me that it'll start at 11:53PM.
![photo_2022-06-28_09-00-24](https://user-images.githubusercontent.com/29897815/176212958-8bc6ae77-20d2-42da-9f92-6eea4264b820.jpg)

But it's on 8:53AM at the time, so it should be saying it'll start at 9:08AM.
![photo_2022-06-28_09-00-34](https://user-images.githubusercontent.com/29897815/176212981-a9a0484b-405c-40df-9bff-035210fcd7d8.jpg)

I am currently on dev-n
![photo_2022-06-28_09-05-33](https://user-images.githubusercontent.com/29897815/176213971-c712b903-3be9-4944-8a97-38bb9753c15a.jpg)

Don't have any logs because I never committed the transaction.

jgslade:
I was able to build dev-n with my fix and it is working
![photo_2022-06-28_16-15-50](https://user-images.githubusercontent.com/29897815/176311848-b93c438e-b376-4a1e-a444-a03f38e94d2c.jpg)
![photo_2022-06-28_16-15-55](https://user-images.githubusercontent.com/29897815/176311952-4aeaf3d2-7a6e-45b4-aca0-67b667127532.jpg)
![photo_2022-06-28_16-15-59](https://user-images.githubusercontent.com/29897815/176311995-7ceb952d-96b8-4345-bf16-296c4ede361a.jpg)

MilosKozak:
https://github.com/nightscout/AndroidAPS/commit/9b6477381bc96f073b874b209f342b0e815a3a3f
thanks!

PR #1864, opened 2022-06-28
Fix for issue  #1862

jgslade:
Fix for eCarb start time from wear being hours instead of minutes


MilosKozak:
it's already in dev

Issue #1878, opened 2022-07-03
3.1-beta1: autosens/dynamicisf not propagated to nsclient


justmara:
Using aaps and nsclient built from 3.1-beta1 branch the autosens value (the correction %) is never shown by nsclient. 
Neither is the DynamicISF value - frozen on 'n/a' state forever.

PR #1881, opened 2022-07-04
dynisf value on overview for nsclient

justmara:
partial fix for #1878
autosens value is now hidden on nsclient (it was never updated, so better hide it)
variableIsf icon/value refreshment code moved to `updateSensitivity` method
modified layout a bit so hidden sensitivity text does not make variable-isf to stick to top


justmara:
rebased on dev and retargeted PR

justmara:
here is what it looks like in nsclient
Screenshot_20220723-094056_NSClient

MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/b38cb58338752473df4c7b3cc22eb46d40a503ae

Issue #1882, opened 2022-07-04
EROS: Pod is not discarded properly


vanelsberg:
**AAPS version 3.0.0.1 release and higher**
**Driver: Omnipod EROS**

**Problem description:**
The Omnipod EROS driver has a bug that can lead to a situation where users get stuck with a pod registration that can not be continued or discarded. Several users reported running into this problem as a result of Pod activation failure.

Problem was introduced by [Eros: try to resolve fake TBR alert](https://github.com/nightscout/AndroidAPS/commit/4cea57acf6d74baffef83e1f04376b10bb5c1978). As a result of this "fix" pod state  can no longer be discarded through code. Pod may also not be assigned a new random Pod address on activation.

As a direct result user is unable to Discard Pod on failures during activation where the Pod can not be succesfully activated.
See also: https://github.com/nightscout/AndroidAPS/issues/1853

**Solution:**
Revise code logic for discarding a Pod.

vanelsberg:
Possibly related:
https://github.com/nightscout/AndroidAPS/issues/1627
https://github.com/nightscout/AndroidAPS/issues/1129

vanelsberg:
@MilosKozak Working on a PR, currently under test

PR #1896, opened 2022-07-07
Fixing issue with EROS Pod state causing pod not being discarded properly

vanelsberg:
Fixing issue with EROS Pod state causing pod not being discarded properly
https://github.com/nightscout/AndroidAPS/issues/1882


vanelsberg:
Tested OK:
- Pod activation
- Pod discarding (developer mode)
- Pod de-activation

PR #1908, opened 2022-07-12
Feat: local profile select active

Andries-Smit:
When the local profile is opened, it should default show the currently active profile. 
Previously it selected always the first.


Philoul:
Before when you just cloned a profile from treatments (or created from autotune), the new created profile was selected by default...
I think each time you clone or copy a new profile, you edit it immediatly (at least to cjange default name or adjust some parameter)
With this PR new created profile is no more selected by default in local profile, even temporary 😞.
I agree having the first profile selected when no action was done is not the best and current active profile is better, but could we improve default selection when a new profile has been created or updated recently?

Andries-Smit:
@Philoul thanks for your input. I did not think about that case. As we say, a bug for one, can be a feature for another.
We could fix it by passing with the intent the prefered selected profile.

PR #1910, opened 2022-07-13
Feat Wear OS Crashlytics

Andries-Smit:
Forward Wear OS crashes to phone, where they show in the log and register by Crashlytics


MilosKozak:
I was trying to do the same but without success. This is working??

Andries-Smit:
@MilosKozak sure! see crashlytics you will see a view "Testing Wear OS Crash ...." from various sources, the app, tile and activity.
image

There are two routes to log error, catch store in file system -> send on startup For example https://github.com/eBay/WearCrashReporter Or catch and send directly  https://gist.github.com/mauimauer/c6f40ec89863906e3b7a 
Any preference for the AAPS implementation?

app/src/main/java/info/nightscout/androidaps/plugins/general/wear/wearintegration/DataLayerListenerServiceMobile.kt
```diff
@@ -115,12 +114,37 @@ class DataLayerListenerServiceMobile : WearableListenerService() {
...
                     val command = EventData.deserialize(String(messageEvent.data))
                     rxBus.send(command.also { it.sourceNodeId = messageEvent.sourceNodeId })
                 }
+
+                exceptionPath -> logWearException(messageEvent)
+            }
+        }
+    }
+
+    private fun logWearException(messageEvent: MessageEvent) {
```
MilosKozak, line 128: I'd move this to FabricPrivacy and do not expose `instance()`

wear/src/main/java/info/nightscout/androidaps/comm/ExceptionService.kt
```diff
@@ -0,0 +1,110 @@
...
+        transcriptionNodeId?.also { nodeId ->
+            messageClient
+                .sendMessage(nodeId, path, data).apply {
+                    addOnSuccessListener { }
+                    addOnFailureListener {
+                        Log.d("WEAR", "sendMessage:  $path failure")
+                    }
+                }
+        }
+    }
```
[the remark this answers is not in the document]
Andries-Smit: Sure it is duplication, I tried to re-use it, but I got trouble starting the service in the background.

The benefit of the code being independent allows us to log issues when there are issues in DataLayerListener too.

app/src/main/java/info/nightscout/androidaps/plugins/general/wear/wearintegration/DataLayerListenerServiceMobile.kt
Andries-Smit: When moving the whole function, it feels a bit dirty to pass MessageEvent to FabricPrivacy... not separation of concern?

MilosKozak: you can decompile it here and then pass to FabricPrivacy as Exception

wear/src/main/java/info/nightscout/androidaps/comm/ExceptionService.kt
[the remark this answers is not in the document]
Andries-Smit: Update with re-use DataLayer and rx

MilosKozak:
and still working?

Andries-Smit:
> I was trying to do the same but without success. 
What failed?

Issue #1920, opened 2022-07-19
3.1.0 White Theme Problems


m4nnihuber:
When using Light Theme on a Pixel 5 App crashes and can't be restarted.
Keeps on crashing and AAPS doesnt open again

osodebailar:
What version before updating?
Try a clean install and reimport your settings.

m4nnihuber:
Came from 3.0.0.1 

clean installation tried but no solution.
app keeps crashing when switching to light theme

osodebailar:
If you install it clean without import settings and than switch to white theme it also crashes ?

m4nnihuber:
yes. same problem

osodebailar:
That’s very strange .
You can send a logfile and maybe write me in messenger for further discussion if you want

MilosKozak:
https://console.firebase.google.com/u/0/project/androidaps-c34f8/crashlytics/app/android:info.nightscout.androidaps/issues/e7837a05364e6f62fcace2ec3536caef?time=last-thirty-days&versions=3.1.0%20(1500);3.0.0.2-dev-o%20(1500)&sessionEventKey=62D70FE100B000017F956356B90FD450_1700624401145498450

pimeys:
Can confirm same behavior on a OnePlus 8 Pro. Already downgraded, so sadly no logfile. Ping me in Discord if I can add something to the issue.

lucasvreis:
Since people on discord were reporting the issue, I ran v3.1with logcat and the first time I changed the theme to white, it gave me this:
```log
--------- beginning of crash
07-19 18:54:43.827 19270 19270 E AndroidRuntime: FATAL EXCEPTION: main
07-19 18:54:43.827 19270 19270 E AndroidRuntime: Process: info.nightscout.androidaps, PID: 19270
07-19 18:54:43.827 19270 19270 E AndroidRuntime: android.view.InflateException: Binary XML file line #68 in info.nightscout.androidaps:layout/overview_fragment: Binary XML file line #68 in info.nightscout.androidaps:layout/overview_fragment: Error inflating class <unknown>
07-19 18:54:43.827 19270 19270 E AndroidRuntime: Caused by: android.view.InflateException: Binary XML file line #68 in info.nightscout.androidaps:layout/overview_fragment: Error inflating class <unknown>
07-19 18:54:43.827 19270 19270 E AndroidRuntime: Caused by: java.lang.reflect.InvocationTargetException
[...]

PR #1926, opened 2022-07-20
fix white theme crash

osodebailar:
fix for issue https://github.com/nightscout/AndroidAPS/issues/1920

@MilosKozak 
The white theme crash was caused due to the merge of the pump-common module.

 removed style from pump common and moved the spinner style to the global noght and light styles.xml including better naming


MilosKozak:
how did you locate it?

osodebailar:
> 

not easy  about some hours to find it . As i know that it can't be a direct coding problem i checked the commits and found the commit number where the problem occurs.

PR #1930, opened 2022-07-20
#1929: fix menu in historybrowser

### LINKED ISSUE #1929: 3.1: history browser charts menu without labels
Chart selector menu shown fine on overview, but has empty labels in History browser

image
Screenshot_20220720-152747_NSClient

justmara:
fixes #1929
these labels had `0` color for both foreground and background spans when called from `HistoryBrowseActivity`

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewMenus.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/overview/OverviewMenus.kt
@@ -134,8 +134,8 @@ class OverviewMenus @Inject constructor(
                         val item = popup.menu.add(Menu.NONE, m.ordinal + 100 * (g + 1), Menu.NONE, rh.gs(m.nameId))
                         val title = item.title
                         val s = SpannableString(" $title ")
-                        s.setSpan(ForegroundColorSpan(rh.gac(context, m.attrTextId)), 0, s.length, 0)
-                        s.setSpan(BackgroundColorSpan(rh.gac(context, m.attrId)), 0, s.length, 0)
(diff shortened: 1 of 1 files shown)


osodebailar:
Sure that's a good solution to remove context ?
Have you tested it with dark and white theme ?
Generally we need the context to get the right access tongue correct white and dark color table .

justmara:
light theme crashes on my phone (and in android 12 emulator) too even on clean 3.1, even without this fix :)

osodebailar:
> light theme crashes on my phone (and in android 12 emulator) too even on clean 3.1, even without this fix :)

It’s solved in dev .
And you can’t merge in master .
Solutions are made in a branch from dev and merged in dev.

justmara:
ok, I'll rebase it a bit later and push here again

justmara:
light theme built from dev.
image
image

justmara:
rebased on dev and changed PR target to dev also

osodebailar:
How the sensitivity and deviation menu item in light and dark look like ?

justmara:
image
image

osodebailar:
Seems to work

justmara:
>Sure that's a good solution to remove context ?

this method has an overload, so when context is not supplied - it uses ResourceHelper's own context instead

image

osodebailar:
> > Sure that's a good solution to remove context ?
> 
> this method has an overload, so when context is not supplied - it uses ResourceHelper's own context instead
> 
> image

Lol , I know that’s my function .
But it takes the default context . it’s usaslly only if gac is used in classes where you can’t get access to the actual fragment context . 
In most cases it is important to to call the function with the “actual” fragment context . Otherwise you get always the default theme and Color’s and that’s the dark mode themes and Color’s .

justmara:
in history browser there is no fragment, but activity :)

Issue #1960, opened 2022-07-24
3.1.0.2: Important info truncated in main screen


Arne1346:
Hi,

with 3.1. some UI improvements have been applied. Unfortunately now a 6 inch phone (Xioami Mi A2) with a resolution 1080 x 2160 pixels is no longer able to display important information on the main screen due to truncation:

![signal-2022-07-24-134333_002](https://user-images.githubusercontent.com/72160775/180645623-0a009cf9-1d94-43f6-988d-3787d4509e65.png)

The red 1 in the middle of the line is the first digit of the remaining insulin units. At the moment the screenshot was taken it was 19 IU (which makes quite a different to 1 IU as displayed).

Can you rethink and enhance the new layout to make sure that all information can be displayed as before in 3.0?

thank you,
Arne

Alamo04:
Confirmed
It's because the german translation is to long (Std) instead the english (h).

Solution for development: Don't translate this information and let it stay in english.

Solution for user: If needed, set AAPS language to english.

cr582:
This issue is also seen in treatment tab among other locations. See also discussion and screenshots in #1394 (which was unfortunately and incomprehensibly closed).

jgordon5:
I have the same problem with my Honor 50 5g. Sometimes it shows full information on home screen, at other times it's truncated. I put in the screen/pixel dimensions in Android Studio but it's still often truncated and doesn't show COB in full or IOB.

MilosKozak:
> I have the same problem with my Honor 50 5g. Sometimes it shows full information on home screen, at other times it's truncated. I put in the screen/pixel dimensions in Android Studio but it's still often truncated and doesn't show COB in full or IOB.

screenshot?

jgordon5:
I'll send it to you as soon as it does it again.
Jill Gordon

jgordon5:
Hello
Here you are, the IOB is truncated and the sensitivity on the right hand
side.
Another screenshot shows insulin actually on board.

Best wishes
Jill

jgordon5:
![Screenshot_20220805_195429](https://user-images.githubusercontent.com/38868358/183264281-96f527b1-f612-4014-8e7b-eb75f76def37.jpg)
![Screenshot_20220805_195415](https://user-images.githubusercontent.com/38868358/183264289-6ea9fec7-70ff-40f3-a0bb-23b21bf5e653.jpg)
[...]

PR #1968, opened 2022-07-27
new statuslight layout and position

osodebailar:
Propsal for style change of statuslight.

Due to many problems depending on different smartphones and custom fonts, i change the style and position of the statuslight information.
Now the information is more clearer viewable without any sizing problems.

Thx @swissalpine  for inspiration

Fix issue https://github.com/nightscout/AndroidAPS/issues/1960

Status_light_proposal_dash
Status_light_proposal_otherpump
Status_light_proposal_otherpump_light

### DIFF
--- a/app/src/main/res/layout/overview_fragment.xml
+++ b/app/src/main/res/layout/overview_fragment.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:orientation="vertical"
     tools:context=".plugins.general.overview.OverviewFragment">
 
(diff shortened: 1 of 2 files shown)


swissalpine:
Looks good. 😉

Philoul:
I don't like it, we loose a lot a space for graphs when we have enough place on a single line to show everything.
 
The idea (with @Andries-Smit ), was to use a single line when we can, and use a double line only if text is too long.

In this proposal, we always have the equivalent of 3 lines one for icon, and 2 lines of information...👎

osodebailar:
But the information is now much more clearer.
Scrolling is easy and easy to handle to see al the graphs . 
I like the idea from @swissalpine , maybe because it’s nearly the same of my older gui replacement .

To use one line seems to be better , but use dynamic 2 lines if the text size is greater , I don’t like it . 
One line for all or 2 line for all , but not a dynamic change .

Philoul:
it's not a dynamic change, but very very static...
99.5% of the time the information fits to available space or not.
During the last 2 months, I got hidden information only one or 2 times (1 hour max) because of combination of long numbers (max number of digits without any "1")...

Philoul:
And what's the result in landscape orientation?

Philoul:
> One line for all or 2 line for all , but not a dynamic change .

In your proposal don't see 2 lines but 3 lines for all, one line for icon and 2 lines for informations...

osodebailar:
If we do this than without dynamic. Dynamic changes are not very user friendly and should always be avoided.

lets do the challenge and make the other proposal and let Milos decide what we do.

osodebailar:
> > One line for all or 2 line for all , but not a dynamic change .
> 
> In your proposal don't see 2 lines but 3 lines for all, one line for icon and 2 lines for informations...

Yes indeed . 
One line for all or 2 line for all , I mean your proposal . Let’s see the result of your proposal

koelewij:
I agree with Philoul! With the Omnipod one line is more than enough to show all information. Why waste a line when that is not required.
When not everything would fit, a 2nd line could be an option. But also the people that do the translation should be instructed to keep text in that part of the UI as short as possible.

koelewij:
@osodebailar i would not know if it is possible in "the translation tool". I never used it. But when this is especially for making different translation of an app, this seems a useful feature.
Maybe time to dig into the translations so i can support the development with Dutch.

swissalpine:
As far as I know, the translation tool does not know any text length defaults here. In the past, before collaborative translation, there were always considerations in German translation as to how it would look in the GUI. For example, terms were once intentionally separated incorrectly if they were too long for the buttons in the care portal: Pumpenbatteriewechsel -> Pump Batterie Wechsel.
Now there are too many people working on it who don't have such considerations in mind (and especially don't think about small displays).
Long story short: From the translation side, no support is to be expected here - and even if it were solved once now, a next - well-intentioned - would make it worse again.

I see again and again on screenshots, not only on very small screens, that the status lights do not look good: Squashed text, hardly any padding, just ugly. This is what a multi-line solution speaks for.
Conversely, I also understand that each line more also wastes space (but why this should be a problem on large smartphones, with the enormous lengths, I wonder).

My original concept of status lights were exclusively icons that change color (white, yellow, red - hence the name); you could see the exact values with a swipe on the status page. This is how it was initially in AAPS. But some wanted to see all the information on one side, i.e. every minute of battery life ... Hence the problems now: a less important desert of numbers in the middle of the screen.
Again, long story short: I would be very happy with an option that hides the numbers and reintroduces the discoloring icons instead. For those who like it that way (be it due to a small screen, be it due to the perceived superfluity of the detailed informations).
I also like the two/three line layout of @osodebailar because it takes up the rhythm of IOB etc. and makes the screen look tidier and more structured (unfortunately the graph would be cut off on my Jelly 2, but maybe there would be the colored icon alternative again).
I also like that the status lights move to the top, then they are somewhat out of sight and do not distract so much from the important numbers (BG, IOB etc.) and the graph.

Andries-Smit:
@osodebailar I did worked on a similar PR. With the idea to use screen size to determine small or big devices. https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes
But I did not finshed it and I found is hard te make it work and determine the right sizes
https://android-doc.github.io/guide/practices/screens_support.html

ArthurusDent:
> Now there are too many people working on it who don't have such considerations in mind (and especially don't think about small displays). Long story short: From the translation side, no support is to be expected here - and even if it were solved once now, a next - well-intentioned - would make it worse again.

I've been using AndroidAPS in German for years. I can't remember this specific part of the GUI ever having that problem. Also, if I remember correctly, you can leave comments in Crowdin, why a certain translation has been chosen. Careful translators should read those comments. I know that Achim is gone, who was the main translator for German, and I haven't been on Crowdin for quite a while, but usually not every new translator has the power to just change translations. They have to be signed off by one of the senior translators. This seems to have gone wrong here and hopefully this can be fixed and avoided in the future.

> I see again and again on screenshots, not only on very small screens, that the status lights do not look good: Squashed text, hardly any padding, just ugly. This is what a multi-line solution speaks for. Conversely, I also understand that each line more also wastes space (but why this should be a problem on large smartphones, with the enormous lengths, I wonder).

Here is an example of why I oppose this change. For context, I'm on a 5.5 inch screen with a resolution of 1080x1920. My paint skills are bad but I think you see that the first custom graph below the main graph would be cut off if the new status lights were now being tripled. 

Screenshot_20220727-092959_AndroidAPS

Screenshot_20220727-092959_AndroidAPS

Alamo04:
@ArthurusDent 
Your screenshots are all with not translated Status Lights. In that case there is no problem.
If you checkout a newer version, with translated status lights, you will not longer see all information, because instead (h) it is now (Std.) in german version.

PR #1974, opened 2022-07-29
small debug logging optimization

justmara:
As you can see - most of loadBgData function work time is taken by `toISOString()`, that is called for debug messages, that are (near) never written to log because of disabled tag. But string is still combined before logger call.
These logger calls are (mostly) inside loops, so they greatly affect function's performance.
So I've added another `debug()` method to logger, that takes lambda for message composition and only invokes it in case of real log write.

image


MilosKozak:
👍

PR #2005, opened 2022-08-10
Add notes to DB

MilosKozak:


MilosKozak:
Break DASH code at the moment

justmara:
My code is used for about two months already by people with Dash too. So that must be the latest pump driver's changes that causes crashes :(

Issue #2016, opened 2022-08-15
Dash issue. f7951ec


elgrupo123:
Same issues with the Dash as before. No bolus delivery possible. Build f7951ec
![PXL_20220815_175939683](https://user-images.githubusercontent.com/92778761/184697285-5da1b04b-1b13-45b8-9f3c-868611c68e5d.jpg)

![PXL_20220815_180106362](https://user-images.githubusercontent.com/92778761/184697301-ecc15af3-0161-4ddd-908c-8d4126db9441.jpg)

The older version - works fine!
![Screenshot_20220815-204647](https://user-images.githubusercontent.com/92778761/184697570-5156c5e1-f5fd-4c36-8994-d6b1912695b5.png)

mick711:
Same issue go back to old dev

vanelsberg:
@justmara Tnx!!!
Will see if I can test this...

MilosKozak:
reverted

PR #2018, opened 2022-08-16
fix for Dash database

justmara:
fix for #2016 : the dash database maps entries as columns with prefixes so it must be migrated or recreated after [adding](https://github.com/nightscout/AndroidAPS/commit/4b3861d523e6c3d28eb1c7153f9e6204b2915194#diff-5d2a6c189c8fdefc38ebfebf607f93f40a7c7762659f30177987a95d41a33091R8) a field to db entity


justmara:
@vanelsberg Milos reverted dev, removing failing code, so this fix is no more actual for dev. So if you take this change - you must be careful on next update , because it can cause dash database cleanup again.

vanelsberg:
> @vanelsberg Milos reverted dev, removing failing code, so this fix is no more actual for dev. So if you take this change - you must be careful on next update , because it can cause dash database cleanup again.

Thanks for the update! The code revert on dev appeared to be working ok for me.
Will delay testing this for now.

PR #2024, opened 2022-08-17
bypasses the NSServer certificate check in engineering mode

lvlv:
We currently need a full functional NSServer for AAPS to connect even in debug/engineering mode.
This change bypasses the NSServer certificate check in engineering mode, making it easier for developers to build a local server.

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/general/nsclient/services/NSClientService.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/general/nsclient/services/NSClientService.kt
@@ -3,6 +3,7 @@ package info.nightscout.androidaps.plugins.general.nsclient.services
 import android.annotation.SuppressLint
 import android.content.Context
 import android.content.Intent
+import android.net.SSLCertificateSocketFactory
 import android.os.*
 import androidx.work.OneTimeWorkRequest
 import com.google.common.base.Charsets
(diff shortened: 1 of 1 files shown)


MilosKozak:
i don't think this should be in master. Do it for yourself

lvlv:
> i don't think this should be in master. Do it for yourself

That will be a burden for those who build local ns server. I have to patch over every time I upgrades.
May I know your concern?

lvlv:
We do see a lot of requests on discord:
image

If disabling cert checks in engineering mode is not a good choice, will a new option to "allow insecure https server" be better?

swissalpine:
> I have to patch over every time I upgrade.

You can commit your changes to keep them! Multiple patches aren't necessary.

lvlv:
> > I have to patch over every time I upgrade.
> 
> You can commit your changes to keep them! Multiple patches aren't necessary.

Sorry, I can't catch it, even if I committed my changes, we have to rebase on new versions every 90 days when we are forced to upgrade.

swissalpine:
No, just pull the changes. I modified a lot of the code to my needs. That's no problem.

lvlv:
> No, just pull the changes. I modified a lot of the code to my needs. That's no problem.

You are betting no conflicting changes. This is certainly not friendly to end users.

swissalpine:
No, but normaly I'm capable to solve them.

vanelsberg:
> i don't think this should be in master. Do it for yourself

Agree. As far as I can tell this disables SSL checking always when in developer mode? 
I do not think we should want that?

Running dev versions 90% personally I would not! For a developer there are other ways to get by, for instance with a self-signed cert or using a free certificate from Let's Encrypt.

lvlv:
I think use an option to whitelist a specific cert should be a better
choice, but I didn't try that out.

Theodoor ***@***.***> 于 2022年11月19日周六 上午6:45写道：

savek-cc:
Maybe even "just"(?) install the self signed NS-cert into the android certificate store (so independent of AAPS)?
https://coderwall.com/p/wv6fpq/add-self-signed-ssl-certificate-to-android-for-browsing

PR #2043, opened 2022-08-24
fix usage of SMB munites instead of UAMSMB minutes on negative iob

### LINKED ISSUE #2042: negative iob causes to use of SMB minutes instead of UAMSMB
This [line of code](https://github.com/nightscout/AndroidAPS/blob/master/app/src/main/assets/OpenAPSSMB/determine-basal.js#L1061) determines what limit will be applied to SMB size. IOB can be negative - for example overnight, after compensating some lows. Then on any rise this block will fall to maxSMBBasalMinutes instead of maxUAMSMBBasalMinutes, which will result in higher instant insulin delivery.

When does it matter? For example when sensor is squeezed while sleeping so it lowers the bg data. When released it will rise fast to original level. So high delta will lead to UAM trigger, but this UAM will use regular SMB minutes as a limit. So defensive low-minutes limit for UAM minutes wont protect you.

justmara:
fixes #2042

uses `meal_data.mealCOB > 0 && iob_data.iob <= mealInsulinReq` as a switch to SMB minutes, if it fails - falls back to UAM minutes


szantos:
May be related: #2096

PR #2056, opened 2022-08-30
Fix duration calculation affecting overlapping TBRs

laurakcaron:
# Fix negative basal IOBs from overlapping entries (Medtronic)

## Issue

I believe this is may be related to nightscout/AndroidAPS/#1597 and nightscout/AndroidAPS/#1724

The issue can be replicated by running MedtronicHistoryDataUTest.kt.

The troublesome temp basal (which has the overlapping duration) begins at 1:55:05. Then there is a zero duration temp basal at 1:59:43 and then a gap in time before the next one at 2:04:45. Here is the relevant data:

```

 {
	"entryType": "TempBasalCombined",
	"DT": "31.10.2021 01:55:05",
	"atechDateTime": 20211031015505,
	"decodedData": {
		"Object": {
			"durationMinutes": 30,
			"insulinRate": 0.0,
			"isPercent": false
		}
	}
}, {
	"entryType": "TempBasalCombined",
	"DT": "31.10.2021 01:59:43",
	"atechDateTime": 20211031015943,
	"decodedData": {
		"Object": {
			"durationMinutes": 0,
			"insulinRate": 0.0,
			"isPercent": false
		}
	}
}, {
	"entryType": "TempBasalCombined",
	"DT": "31.10.2021 02:04:45",
	"atechDateTime": 20211031020445,
	"decodedData": {
		"Object": {
			"durationMinutes": 30,
			"insulinRate": 0.55,
			"isPercent": false
		}
	}
}, 

```

The problem appears in the calculation of the duration of zero temp basals, which occurs in MedtronicHistoryData.kt. The condition 

```

(previousItem.durationAsSeconds > initialDuration) 

```

compares the time between the start of the temp basal at 1:55 with the start of the next valid temp basal at 2:04 (which is previousItem.durationAsSeconds) with the time between the start of the temp basal at 1:55 and the zero duration temp basal at 1:59 (initialDuration). Finding it to be larger, the duration is set to 30 minutes (1800 seconds) and we get a value like this: 

```

ItemOne: PumpHistoryEntry [type=TempBasalCombined    [-2 , 0xFE], DT: 31.10.2021 01:55:05, length=0,0,0(0), rawData=], ItemTwo: null, Duration: 1800, ObjectType: TemporaryBasal

```

with duration of 1800 seconds. But the temp basal starting at 1:55 could not have gone on longer than until 2:04 when the next one began.

## Solution

This fix adjusts the calculation by changing the value initialDuration to represent the duration of the temp basal set on the pump.

``` 

val initialDuration = previousItem.itemOneTbr!!.durationMinutes*60

```

### DIFF
--- a/medtronic/src/main/java/info/nightscout/androidaps/plugins/pump/medtronic/data/MedtronicHistoryData.kt
+++ b/medtronic/src/main/java/info/nightscout/androidaps/plugins/pump/medtronic/data/MedtronicHistoryData.kt
@@ -217,8 +217,11 @@ class MedtronicHistoryData @Inject constructor(
             allHistory.removeAll(removeList.toSet())
             this.sort(allHistory)
             aapsLogger.debug(
-                LTag.PUMP, String.format(Locale.ENGLISH, "All History records [afterFilterCount=%d, removedItemsCount=%d, newItemsCount=%d]",
-                                         allHistory.size, removeList.size, newHistory.size))
+                LTag.PUMP, String.format(
+                    Locale.ENGLISH, "All History records [afterFilterCount=%d, removedItemsCount=%d, newItemsCount=%d]",
+                    allHistory.size, removeList.size, newHistory.size
+                )
+            )
         } else {
             aapsLogger.error("Since we couldn't determine date, we don't clean full history. This is just workaround.")
         }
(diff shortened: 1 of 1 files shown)


andyrozman (changes requested):
This is wrong fix. 

You need to always use durationAsSeconds, because it contains logic. If there is something wrong, with data returned, then it needs to be addressed there in durationAsSeconds in class TempBasalProcessDTO. This class collects all starting and ending TBRs together and calculates time between. So maybe your entry is not collected correctly or put in the structure... Error must be somewhere else... 

Which ticket is that concerning? Did you provide data? I can check this is next two weeks to see what is going wrong...

Issue #2055, opened 2022-08-29
SMBs often fail [3.1.0.3] [DASH]


rustymonkey:
Based on observation, it appears that after an initial SMB has been successfully given, subsequent SMBs will often fail for a few loop cycles (I think that basals are adjusting correctly though). This issue only started occurring after upgrading from 3.0.3 to 3.1.0.3 (3.0 experienced occasional disconnects from the DASH, but not multiple sequential SMB failures that often result in high BG).

I've replicated this using BOOST 3.1.0.3, with Boost disabled. In the example below, you can see a successful SMB at 5:51am, and then failed boluses at 5:06 and 6:01. My son was sleeping (no change in position).

Note: When the SMBs fail there's no notification in AAPS or a system notification.

![2022-08-29_AAPS_551](https://user-images.githubusercontent.com/34991028/187188313-3819b1ef-5344-45eb-8eee-a9390bf97983.jpg)

![2022-08-29_AAPS_556](https://user-images.githubusercontent.com/34991028/187188343-636be20f-de2a-4a81-a813-0c62d5551c16.jpg)

![2022-08-29_AAPS_601](https://user-images.githubusercontent.com/34991028/187188378-7bfae0e2-6dad-419f-90c3-ca4331652c96.jpg)

[AndroidAPS_LOG_1661768474395.zip](https://github.com/nightscout/AndroidAPS/files/9444031/AndroidAPS_LOG_1661768474395.zip)

Thomas-Wolff:
I can confirm the same behaviour with the Eros. The newly calculated basal rate is switched through to the pump, the SMB flashes briefly and the update of the main runs so that the SMB fails. In the Loop TAB under SMB there is then the information 'waiting for pump result'.

I have just switched to version 3.0.0.2 for testing, SMB runs cleanly.

Hardware:

Samsung Galaxy A40 Android 11
Dexcom G6 via Byoda

Greetings Thomas

Ashley112885:
I seem to be having a similar issue and often find the Loop tab says “SMB set by pump: waiting for result” and have to manually select “run” to trigger the SMB.
Omnipod Dash
AAPS 3.1.0.3
Pixel 3a
Android 12
Dexcom G6 via BYOD

rustymonkey:
2 other users on the Boost Discord channel also mentioned experiencing this problem on 3.1.x and that it might be caused by a timing issue described at #1395

rustymonkey:
[...]

PR #2070, opened 2022-09-11
TDD calculation cache optimization

justmara:
TDD calculation is based on 5-minute timeframes, but it uses 1-second granulated cache-keys when retrieving data. That makes all the TDD calculation calls to miss caches badly. Mostly it affects TDD-based Loop plugins, like Dynamic ISF.

Here I align time-range for TDD calculation on a 5-minute base. This significantly drops calculation times of subsequent calls.

First run
image

Second run (original)
image

Second run over 5-minute aligned interval
image

After some warmup it speeds up even more :)

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/configBuilder/ProfileFunctionImplementation.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/configBuilder/ProfileFunctionImplementation.kt
@@ -43,7 +43,7 @@ class ProfileFunctionImplementation @Inject constructor(
     private val deviceStatusData: DeviceStatusData
 ) : ProfileFunction {
 
-    val cache = LongSparseArray<Profile>()
+    val cache = HashMap<Long,Profile?>()
 
     private val disposable = CompositeDisposable()
 
@@ -54,10 +54,10 @@ class ProfileFunctionImplementation @Inject constructor(
             .subscribe(
                 {
                     synchronized(cache) {
-                        for (index in cache.size() - 1 downTo 0) {
-                            if (cache.keyAt(index) > it.startDate) {
-                                aapsLogger.debug(LTag.AUTOSENS, "Removing from profileCache: " + dateUtil.dateAndTimeAndSecondsString(cache.keyAt(index)))
-                                cache.removeAt(index)
+                        for (key in cache.keys) {
+                            if (key > it.startDate) {
+                                aapsLogger.debug(LTag.AUTOSENS, "Removing from profileCache: " + dateUtil.dateAndTimeAndSecondsString(key))
+                                cache.remove(key)
                             } else {
                                 break
                             }
@@ -98,14 +98,17 @@ class ProfileFunctionImplementation @Inject constructor(
         val rounded = time - time % 1000
         // Clear cache after longer use
         synchronized(cache) {
-            if (cache.size() > 30000) {
+            if (cache.keys.size > 30000) {
                 cache.clear()
                 aapsLogger.debug("Profile cache cleared")
             }
-            val cached = cache[rounded]
-            if (cached != null) return cached
+            if (cache.containsKey(rounded)) {
+                //aapsLogger.debug(LTag.PROFILE, "Profile cache HIT for $rounded")
+                return cache[rounded]
+            }
         }
 //        aapsLogger.debug("getProfile called for $time")
+        //aapsLogger.debug(LTag.PROFILE, "Profile cache MISS for $rounded")
         val ps = repository.getEffectiveProfileSwitchActiveAt(time).blockingGet()
         if (ps is ValueWrapper.Existing) {
             val sealed = ProfileSealed.EPS(ps.value)
@@ -130,6 +133,9 @@ class ProfileFunctionImplementation @Inject constructor(
             }
         }
 
+        synchronized(cache) {
+            cache.put(rounded, null)
+        }
         return null
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
```diff
@@ -121,14 +123,14 @@ class TddCalculator @Inject constructor(
             tdd.carbs += t.amount
         }
         val calculationStep = T.mins(5).msecs()
-        for (t in startTime until endTime step calculationStep) {
-            val tbr = iobCobCalculator.getTempBasalIncludingConvertedExtended(t)
```
justmara, line 125: this call is done inside getBasalData in case of cache miss, so it is fair replacement

app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
```diff
@@ -41,54 +41,25 @@ class TddCalculator @Inject constructor(
...
-                }
-            repository.getCarbsDataFromTimeToTimeExpanded(startTime, endTime, true).blockingGet().forEach { t ->
-                val midnight = MidnightTime.calc(t.timestamp)
-                val tdd = result[midnight] ?: TotalDailyDose(timestamp = midnight)
-                tdd.carbs += t.amount
-                result.put(midnight, tdd)
-            }
-
-            val calculationStep = T.mins(5).msecs()
-            val tempBasals = iobCobCalculator.getTempBasalIncludingConvertedExtendedForRange(startTime, endTime, calculationStep)
```
MilosKozak, line 71: this was more efficient because it reduced amount of SQL calls significantly

app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
```diff
@@ -111,7 +82,9 @@ class TddCalculator @Inject constructor(
     }
 
     fun calculate(startTime: Long, endTime: Long): TotalDailyDose {
-        val tdd = TotalDailyDose(timestamp = startTime)
+        val startTimeAligned = startTime - startTime % (5 * 60 * 1000)
```
MilosKozak, line 85: this should be already alligned because startTime is midnight (well not XXXXX000 in milliseconds but the same timestamp all the time)

MilosKozak:
the only benefit a see is reduce amount of Midnight calculations which is time expensive
But it should be cached as well

app/src/main/java/info/nightscout/androidaps/plugins/configBuilder/ProfileFunctionImplementation.kt
```diff
@@ -43,7 +43,7 @@ class ProfileFunctionImplementation @Inject constructor(
     private val deviceStatusData: DeviceStatusData
 ) : ProfileFunction {
 
-    val cache = LongSparseArray<Profile>()
+    val cache = HashMap<Long,Profile?>()
```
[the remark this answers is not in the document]
justmara: here we do care about speed much more. so few extra kb of memory are fair price imo :)
but surely it can be changed back if you think it is critical.

app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
justmara: not so clear, but still an improvement: because it were burning cpu in [nested loops](https://github.com/justmara/AndroidAPS/blob/fix/tdd-cache/app/src/main/java/info/nightscout/androidaps/plugins/iob/iobCobCalculator/IobCobCalculatorPlugin.kt#L561) while continuosly searching for matches for every step in interval. not it mostly hits or populates cache entries, used by other TDD calculating parts

app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt
justmara: it is aligned in case of daily tddCalc. but in [other cases ](https://github.com/justmara/AndroidAPS/blob/fix/tdd-cache/app/src/main/java/info/nightscout/androidaps/utils/stats/TddCalculator.kt#L78) `now()` is used as base. thats why 1 second alignment were here basically. but this 1 second mark misses badly, and we cant force callers to care about time alignment to match our internal caches. thats why interval is aligned here.

justmara:
> the only benefit a see is reduce amount of Midnight calculations which is time expensive

all of the TDD calculations benefit from this 5 min alignment. see screenshots and you can profile itself. I were logging hit/miss events and that was the point where I found the difference. there is enormous reduce in calculation times for subsequent calls.

app/src/main/java/info/nightscout/androidaps/plugins/configBuilder/ProfileFunctionImplementation.kt
```diff
@@ -98,14 +98,17 @@ class ProfileFunctionImplementation @Inject constructor(
...
         // Clear cache after longer use
         synchronized(cache) {
-            if (cache.size() > 30000) {
+            if (cache.keys.size > 30000) {
                 cache.clear()
                 aapsLogger.debug("Profile cache cleared")
             }
-            val cached = cache[rounded]
-            if (cached != null) return cached
+            if (cache.containsKey(rounded)) {
```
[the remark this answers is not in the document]
justmara: but old code acted like cache miss on missing data, forcing unneeded db queries. this one clearly stands for `we've searched already and there is nothing`. 
sure this would be better if some kind of sliding expiration were used for cached values... but it'll add complexity

rustymonkey:
ref. to the failed bolus issue: https://github.com/nightscout/AndroidAPS/issues/2055

MilosKozak:
I've got the point. OK

PR #2074, opened 2022-09-13
Intelligo support

christinadamianou:


app/src/main/java/info/nightscout/androidaps/di/PluginsModule.kt
```diff
@@ -368,30 +368,35 @@ abstract class PluginsModule {
...
-    abstract fun bindGlunovoPlugin(plugin: GlunovoPlugin): PluginBase
+    abstract fun bindGlunovoPlugin(plugin: GlunovoPlugin): PluginBase@Binds
 
-    @Binds
     @AllConfigs
     @IntoMap
     @IntKey(475)
+    abstract fun bindIntelligoPlugin(plugin: IntelligoPlugin): PluginBase
+
+    @Binds
```
MilosKozak, line 378: this file is changed too much
add new entry with 473 only

christinadamianou: fixed.

PR #2099, opened 2022-09-24
Fix: smb canceled because of expiration

### LINKED ISSUE #2055: SMBs often fail [3.1.0.3] [DASH]

justmara:
Fixes #2055

[explanation](https://github.com/nightscout/AndroidAPS/issues/2055#issuecomment-1256739170)

### DIFF
--- a/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.kt
+++ b/app/src/main/java/info/nightscout/androidaps/plugins/aps/loop/LoopPlugin.kt
@@ -370,32 +370,33 @@ class LoopPlugin @Inject constructor(
                         if (resultAfterConstraints.bolusRequested()) lastRun.smbSetByPump = waiting
                         rxBus.send(EventLoopUpdateGui())
                         fabricPrivacy.logCustom("APSRequest")
-                        applyTBRRequest(resultAfterConstraints, profile, object : Callback() {
+                        applySMBRequest(resultAfterConstraints, object : Callback() {
                             override fun run() {
+                                // Callback is only called if a bolus was actually requested
                                 if (result.enacted || result.success) {
-                                    lastRun.tbrSetByPump = result
-                                    lastRun.lastTBRRequest = lastRun.lastAPSRun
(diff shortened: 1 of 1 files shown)


MilosKozak:
changing order of SMB <-> TBR is wrong.
Given SMB and failed zero temp afterwards could lead to dangerous situation

justmara:
What exactly dangerous situation do you mean?
If SMB fails - it falls back to 'tempBasalFallback', so TBR part is not needed at all, because it will be overriden.
Whenever SMB command succeeds (even if there was no smb requested) - TBR is enqueued. Yes, TBR command can fail. Then we'll have another 5 minute round of TBR.
In original model it was "_another round of TBR, but without SMB on board_" or even "_one round of zero temp TBR without any SMB_". But look - when SMBs are delivered? On bg rise 99% of the time. So extra round of TBR is not so crucial. While skipped SMB because of failed/slow TBR will result in worse compensation.
When BG is going down and TBR must be canceled - there will be (mostly) no SMB so TBR will be triggered instantly.

MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/baab2dad4c50ff9e3ab7d93660f2204d3c2ffcf5

Issue #1741, opened 2022-05-17
[3.0.0.2-dev-k] "Error CGM data unchanged for the past ~45m..."


chrisbev:
[3.0.0.2-dev-k]  Got "Error CGM data unchanged for the past ~45m..." but data had changed over the 4.7 to 5.1mmol range, see attached.
NB using Tim2000 dISF latest.
Error cleared after about 30min.

![Screenshot_20220517-064101_AndroidAPS](https://user-images.githubusercontent.com/43484095/168740208-bc889efc-3613-4291-bec3-41f376926136.jpg)
![Screenshot_20220517-064123_AndroidAPS](https://user-images.githubusercontent.com/43484095/168740209-26a3cc2e-4eb6-497d-9038-4c346fbf9270.jpg)
![Screenshot_20220517-064239_AndroidAPS](https://user-images.githubusercontent.com/43484095/168740213-dfe7ddd1-46a3-4d12-81cd-51474ecea5b7.jpg)
![Screenshot_20220517-071342_AndroidAPS](https://user-images.githubusercontent.com/43484095/168741570-995a186f-8491-43dd-9c51-eb4416d2b62c.jpg)

Reporting bugs
--------------
- **Note the precise time the problem occurred** and describe the circumstances and steps that caused
  the problem
- Note the Build version (found in the About dialog in the app, when pressing the three dots in the
  upper-right corner).
- Obtain the app's log files, which can be found on the phone in
  _/storage/emulated/0/Android/data/info.nightscout.androidaps/_
  See https://androidaps.readthedocs.io/en/latest/EN/Usage/Accessing-logfiles.html
- Open an issue at https://github.com/nightscout/AndroidAPS/issues/new

MilosKozak:
what CGM?

chrisbev:
FSL2 with patched LibreLink.  (NB Continuos yellow triangle).

chrisbev:
But it was reporting correct values AND they were changing? Please see the xDrip tab values for the 45mins?

chrisbev:
[AndroidAPS._2022-05-17_00-00-05_.1.zip](https://github.com/nightscout/AndroidAPS/files/8712185/AndroidAPS._2022-05-17_00-00-05_.1.zip)

chrisbev:
Log attached

chrisbev:
OK thanks for answer and time....

justmara:
[...]

PR #2143, opened 2022-10-29
fix 'BG data unchanged' check

justmara:
fix for #1741

The stale bg data check moved to SafetyCheck constraint making in plugin-independent. Also [fixed logic](https://github.com/nightscout/AndroidAPS/issues/1741#issuecomment-1295771666) of determining flat (stale) BG data.

Original (erroneous) checks removed from determine-basal.

### DIFF
--- a/core/src/main/java/info/nightscout/androidaps/plugins/iob/iobCobCalculator/GlucoseStatusProvider.kt
+++ b/core/src/main/java/info/nightscout/androidaps/plugins/iob/iobCobCalculator/GlucoseStatusProvider.kt
@@ -5,8 +5,10 @@ import info.nightscout.androidaps.interfaces.IobCobCalculator
 import info.nightscout.shared.logging.AAPSLogger
 import info.nightscout.shared.logging.LTag
 import info.nightscout.androidaps.utils.DateUtil
+import info.nightscout.androidaps.utils.T
 import java.util.*
 import javax.inject.Inject
+import kotlin.math.abs
 import kotlin.math.roundToLong
 
 @Reusable
@@ -100,8 +102,35 @@ class GlucoseStatusProvider @Inject constructor(
         ).also { aapsLogger.debug(LTag.GLUCOSE, it.log()) }.asRounded()
     }
 
-    companion object {
+    fun isBgFlatForInterval(minutes: Long, maxDelta: Double) : Boolean? {
+        val data = iobCobCalculator.ads.getBgReadingsDataTableCopy()
+        val lastBg = iobCobCalculator.ads.lastBg()?.value
+        val now = dateUtil.now()
+        val offset = now - T.mins(minutes).msecs()
+        val sizeRecords = data.size
 
+        if (lastBg == null) {
+            aapsLogger.debug(LTag.GLUCOSE, "lastBg==null")
+            return null
+        }
+        if (sizeRecords == 0) {
+            aapsLogger.debug(LTag.GLUCOSE, "sizeRecords==0")
+            return null
+        }
+        if (data[0].timestamp < now - 7 * 60 * 1000L) {
+            aapsLogger.debug(LTag.GLUCOSE, "oldData")
+            return null
+        }
+
+        for (bg in data)
+        {
+            if (bg.timestamp < offset) break
+            if (abs(lastBg - bg.value) > maxDelta) return false
+        }
+        return true
+    }
+
+    companion object {
         fun average(array: ArrayList<Double>): Double {
             var sum = 0.0
             if (array.size == 0) return 0.0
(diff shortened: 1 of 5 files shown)


app/src/main/assets/AIMI/variant Pam/determine-basal.js
```diff
@@ -0,0 +1,1885 @@
+/*
```
jbr7rr, line 1: Oops, these files are not supposed to be here right?

core/src/main/java/info/nightscout/androidaps/plugins/iob/iobCobCalculator/GlucoseStatusProvider.kt
```diff
@@ -100,8 +102,35 @@ class GlucoseStatusProvider @Inject constructor(
         ).also { aapsLogger.debug(LTag.GLUCOSE, it.log()) }.asRounded()
     }
 
-    companion object {
+    fun isBgFlatForInterval(minutes: Long, maxDelta: Double) : Boolean? {
```
jbr7rr, line 105: Missing the original isSaveCgmSource check here which is set to check if the source is Dexcom (BYODA)

app/src/main/assets/AIMI/variant Pam/determine-basal.js
justmara: woah, sorry :) pushing without it. took commit from my branch

core/src/main/java/info/nightscout/androidaps/plugins/iob/iobCobCalculator/GlucoseStatusProvider.kt
justmara: I didn't changed that check at all. As I've said it _can_ be changed to respect this smoothing option too, but imho that must be an additional change, subject for further discussion.

jbr7rr: Well in the old behavoir (in dev) the Dexcom plugin is excluded from the 45min unchanged check, this is not included in your PR as far as I can see. 

Current implementation in dev:

variable saveCgmSource is set here:
https://github.com/nightscout/AndroidAPS/blob/b4edb71e279954a1aa93a9ac720898d57494221e/app/src/main/java/info/nightscout/androidaps/plugins/aps/openAPSSMB/OpenAPSSMBPlugin.kt#L210

And is used in determine_basal.js here
https://github.com/nightscout/AndroidAPS/blob/b4edb71e279954a1aa93a9ac720898d57494221e/app/src/main/assets/OpenAPSSMB/determine-basal.js#L113

justmara: ahh, this one. ok, i'll re-introduce it here too, ok

justmara: pushed update

justmara: the variable `isSaveCgmSource` is still used in determine-basal for some conditions so seems that it must be reviewed and replaced with something like `byoda || smoothed`, but there are still some points to discuss:

1. byoda with g7 does not have smoothing currently, so the check for byoda becomes not so trustful
2. this check must be moved somewhere else to be shared and not copy-pasted by plugins.

ga-zelle:
@justmara I read the arlier attempts to get this test improved. The final problem in the past was lack of buy-in from US/OpenAPS. Therefore I contacted Scott Lybrand and discussed the matter with him. He has no objection to using a better check for flatness than the one based on the 3 deltas. As a result I suggested a check similar to yours, but slightly different in 3 aspetcs.

Here is the code snippet I inserted in GlucodeStatusProvider.kt right after where the 3 deltas are determined:
```
// calculate length of cgm values staying within a narrow band
val bandWidth = 2                   // max allowed width; how about mmol/L conversions?
var minBG = now.value
var maxBG = minBG
var cgmFlatMinutes = 0.0
var oldDate = now.timestamp
for (i in 1 until sizeRecords) {
    val then = data[i]
    val thenDate = then.timestamp
    minBG = Math.min(minBG, then.value)
    maxBG = Math.max(maxBG, then.value)
    //          outside band            or pause > 11 minutes       or older than 1 hour
    if (maxBG-minBG > bandWidth || oldDate-thenDate > 1000*60*11 || nowDate-thenDate > 1000*60*60) {
        break
    } else {
        oldDate = thenDate
    }
cgmFlatMinutes = ( nowDate-oldDate ) / 60000.0
}
```
The first difference is a bandwidth of 2mg/dl for flatness because a CGM working in mmol/l is allowed a difference of 0.1mmol/l which is converted internally to 2mg/dl.

The second difference is using min and max values for the time window and ensuring that difference is not more than 2mg/dl. Your current check depends on whether the initial value happens to be in the middle if that band or on the edge.

The third diffference is that instead of true/false for fixed 45minutes I return the length of that time window . The orginal 45m came from a "quick fix" at the time because the 3 deltas were available. Now a more meaningful length can be picked like 30m inside determine-basal to switch to fail-safe mode and another length like 60m for raising any alarm like "check your CGM".

MilosKozak:
I did smoother integration in https://github.com/nightscout/AndroidAPS/commit/d5af299692efb0e05762dc0ccf1f38f49836cbdf
hopefully I fixed all bugs from your implementation
comments welcome

PR #2167, opened 2022-11-05
Combo v2 : new Combo driver, written from scratch in Kotlin, does not need or use Ruffy

dv1:
This driver is called "combov2" (to allow for it to coexist with the current combo driver) and is an entirely new Combo driver.

Major features:

- Fully written in Kotlin from scratch, using kotlin coroutines.
- Everything Combo related is done by this driver - Ruffy is not needed nor used. The Bluetooth, Combo protocol, pairing etc. logic that currently is handled by Ruffy is done by this driver. This solves problems with phones where Android aggressivelly kills off Ruffy, and significantly increases stability and usability.
- Pairing is done within AndroidAPS as well, with extra Android UI activities inside the AndroidAPI UI. Cancelling a pairing attempt is handled properly. During testing, pairing worked much faster and more reliable for most people (others did not see a difference). Pairing will always be a little problematic because the Combo operates Bluetooth during pairing in a way that isn't compatible with Android apps that aren't system apps (because the BLUETOOTH_PRIVILEGED permission is needed). But at least this implementation works more reliably.
- A previously paired Combo is automatically un-paired when the user chooses to start the pairing process in the pump's preferences.
- Stored states are also automatically synced with the paired Bluetooth devices, meaning that if there are states associated with pump A, and A was previously manually unpaired through the Android Bluetooth UI, the driver detects this, and automatically deletes the states in its store that refer to (now unpaired) pump A.
- Bolusing is done through the Combo's command mode. This is an alternate mode that is normally used by the Combo's remote control to deliver a bolus. Bolusing this way is much faster since no simulated user interaction in the remote terminal mode is necessary.
- This new driver keeps track of the current timezone offset and automatically adjusts the pump's date and time. The tricks that have been necessary with the current driver to deal with daylight savings changes are unnecessary; daylight savings are handled automatically now.
- The driver's UI fragment in AndroidAPS shows more information, such as the pump ID, progress during various command executions like setting/getting basal profile, details about the current connection attempts etc.
- When the pump runs in the remote terminal mode, the pump screen is shown in the driver's UI fragment.

Some screenshots:
combov2-screenshot-001
combov2-screenshot-002


dv1:
I force-pushed just now to remove some commits that were marked as "[DO NOT MERGE]" and thus weren't supposed to be part of the pull requestr

swissalpine:
Great!
Even though I am very affectionate towards Ruffy and will always be grateful to him, I have to admit that the new driver is superior: the speed of the direct bolus command is enormous, bolus abort alarms are unlikely to happen any more.
Pairing was also smooth and uncomplicated. I had to start a second pairing process because I was not sure whether I had to confirm the pairing on the combo or whether the new driver would do it. But then it went smoothly. Perhaps an additional line here would be helpful:
"4. confirm the pairing request on the pump."
Then the pairing would have worked for me right away.

Many thanks for the work that has been put into this! It is - I'm sorry to say - a big step forward!

dv1:
@swissalpine Thanks :) Good idea about the description, I'll add that line.

dv1:
One remaining issue here is that I was unable to integrate ComboCtl's unit tests due to kotlin.test requiring a newer JUnit than what AndroidAPS currently uses (JUnit 5 vs JUnit 4). Not sure how to proceed. Perhaps we can let them un-integrated until AndroidAPS moves to JUnit 5?

dv1:
As recommended by @jotomo , I ping @T-o-b-i-a-s @Heiner1 @gregorybel @Lexsus @boysim @MasterPlexus @philipgo @cluckj , perhaps they are interested in this as well.

swissalpine:
Looking at the text of the PR, I think we have misunderstood each other:
I meant the point when the name of the mobile phone is displayed on the pump and it has to be confirmed. After that, the actual pairing process starts with the number that has to be entered.

dv1:
> Looking at the text of the PR, I think we have misunderstood each other: I meant the point when the name of the mobile phone is displayed on the pump and it has to be confirmed. After that, the actual pairing process starts with the number that has to be entered.

So, you mean the text would go like this:

```
Steps to perform pairing with your Combo:\n \n 1.
        On your pump, navigate to the Bluetooth Settings\n 2. Check if a device is already shown
        as paired; if so, go to "Delete device" screen to delete/unpair that device\n 3. Go to the
        "Add device" screen, and initiate pairing\n 4. Click on the "Start pairing" button below\n
        5. After a while, the phone's name is shown on the pump's screen; press CHECK to confirm\n
        6. When pairing is successfully completed, confirm the finished pairing on your pump and
        return to the main screen by pressing the CHECK button twice
```

ga-zelle:
Sadly I have a spare phone but the spare Combo is dead (E7). So I wait a bit with testing the new driver while I am busy testing other things.
I very much like the display of the Combo screen in the pump tab, something I suggested aeons ago.

swissalpine:
@dv1 Exactly, this was the point.

dv1:
@swissalpine OK - see the latest commits. I updated the text accordingly.

jotomo:
> Sadly I have a spare phone but the spare Combo is dead (E7)

Was the E7 displayed just once or multiple times (if multiple times, the E7 is shown immediately after battery insertion)?

SandraK82:
If the pump is not yet ready configured for high tbr, the driver blocks while trying to set a higher tbr then 250%. The driver tries to set it higher, the pump goes at some point into timeout and I was not able to restart aaps. Restart was required.

swissalpine:
There is a small glitch in the Combo Tab: The minute display "Last Bolus" is sometimes not correct (see screenshots). In the example it should not say "Less than 1 minute". The bolus was already 23 minutes ago.

Screenshot_20221107-200033
Screenshot_20221107-200044

dv1:
Okay, some logic that checks whether the percentage stays the same for a while in spite of continuing to press the button is needed, followed by an alert to the user that the Combo isn't configured correctly.

As for the bolus; if you get out of the AAPS app and go back in, does it get fixed? Perhaps this is an issue with the refresh of the pump fragment itself.

dv1:
@jotomo Is there special code in the existing combo driver's scripter to catch the TBR percentage case @swissalpine mentioned? I can't find it, but it may be implicitly handled by the checks that verify that the percentage increases as expected.

Also, perhaps it would be wise to add the extra logic that waits for a near-finished TBR to end before programming the new TBR.

jotomo:
> @jotomo Is there special code in the existing combo driver's scripter to catch the TBR percentage case

@dv1 No, the users were expected to start with a proper setup : https://androidaps.readthedocs.io/en/latest/Configuration/Accu-Chek-Combo-Pump.html?highlight=combo#setup

> Also, perhaps it would be wise to add the extra logic that waits for a near-finished TBR to end before programming the new TBR.

There is some special logic in v1 to wait if the current TBR is within the last minute, see: https://github.com/nightscout/AndroidAPS/blob/master/combo/src/main/java/info/nightscout/androidaps/plugins/pump/combo/ruffyscripter/commands/SetTbrCommand.java#L109

Speaking about limits, there's also a hard limit for the size of a bolus (configurable via the PC software mentioned in the article above), have you tested exceeding that limit?

mordechi-david:
I used it today, and it works for me, looks really fast and good, and the initial connection was easy and fast
Thank you very much and all the best for the development

swissalpine:
@dv1 The time glitch happens from time to time and the screenshots show that 23 minutes have passed since the last bolus. Surely this means that AAPS and the combo tab have been exited since then? Or am I misunderstanding you?

dv1:
@swissalpine Does it ever go away? If so, how? Do you have to restart AAPS, or is the time sometimes suddenly OK?

swissalpine:
Unfortunately, I haven't seen when this changes, but occasionally the time is correct - but mostly it's wrong, even when I leave AAPS and go back to it. After a restart, no last bolus is displayed.
Here is another screenshot, this one is the standard view:
Screenshot_20221109-111341

dv1:
@swissalpine Uploaded a fix for the last bolus UI field. It should be displayed correctly now.

swissalpine:
> @swissalpine Uploaded a fix for the last bolus UI field. It should be displayed correctly now.

Yes, ist's fixed! Thanks for your work!

T-o-b-i-a-s:
Hello @dv1, thanks for this awesome driver. Never had such an easy pairing of my Combo before. I really like the setup with everything running from within AAPS, the display of the current pump screen and the faster bolus via the command mode. Will test it and provide feedback if I notice any bugs or issues.

T-o-b-i-a-s:
Hello @dv1, the driver works very well and is quite stable. I also like the feature that it automatically sets the pump's date/time when I insert a new battery and the actual time was lost.
I rarely experience situations where the driver cannot find the RT screen (Error in log is:
new: Error ("Command execution error"); throwable: info.nightscout.comboctl.main.CouldNotFindRTScreenException: Could not find RT screen class info.nightscout.comboctl.parser.ParsedScreen$TemporaryBasalRatePercentageScreen) and the pump vibrates three times in short succession when trying to set a new temporary basal rate. I can provide you with the relevant log files of three such events, if you need them, just tell me how I can send them to you for analysis.

dv1:
@T-o-b-i-a-s I have seen this error once, yeah. But - I was unable to reproduce it. I need a way for reproducing it so I can fix it. Can you come up with one it? As for logs, just export logs through AAPS (in the Maintenace section).

T-o-b-i-a-s:
Hello @dv1, please find attached three sections from the logfile containing the error described above (CouldNotFindRTScreenException): [Combo_v2.zip](https://github.com/nightscout/AndroidAPS/files/10050577/Combo_v2.zip). Tell me if you need more lines before or after the sections included in the file.

dv1:
@T-o-b-i-a-s I think I see now what is going on. In the logs, we can see that the remaining TBR duration changes from 21 to 20 minutes. The code incorrectly interprets this as if the previous mainscreen with remaining duration 21 and the current mainscreen with remaining duration 20 were two different screens. This causes the code to press the MENU button once too often, which in turn causes subsequent button press events to move past the TBR menu screen. I am adding a check for this, then we can test if this fixes things.

dv1:
@T-o-b-i-a-s Just posted a commit. Please check if this fixes the bug for you.

markvader:
Excellent work @dv1 So quick to pair and connect each time. I haven't seen any issues.

What do you think of having a holding image/bitmap in the ComboV2RTDisplayFrameView rather than the grey box.  Is this achievable?  Or a label to describe that it is a view of the pump screen

dv1:
@markvader I actually thought about enlarging the combo SVG at the bottom and placing the frame view somehow inside exactly where the display is in that SVG. Would maybe look nice :) But I don't know how to reliably do that without graphical artifacts.

gregorybel:
@dv1 what an incredible and well done job! How did you manage this alone?

dv1:
@gregorybel Thanks, this took a _long_ time to finish - that's the "secret" ;)

PR #2200, opened 2022-11-13
overview: temp basal as percentage

justmara:
An option to show TBR as percentage of current profile basal for any pump, not only ones that actually use basal percentage.

I (maybe temporary) switched to Dash and its TBR absolute values keep me frustrating: every time looking at it I must remember whats current 'normal' value so how this one is compared to it... The percentage value is much more informative. So I added an option to show TBR as percentage on overview screen (for aaps/nsclient/widget).

### DIFF
--- a/core/core-main/src/main/java/info/nightscout/androidaps/extensions/TemporaryBasalExtension.kt
+++ b/core/core-main/src/main/java/info/nightscout/androidaps/extensions/TemporaryBasalExtension.kt
@@ -60,6 +60,13 @@ fun TemporaryBasal.toStringShort(): String =
     if (isAbsolute || type == TemporaryBasal.Type.FAKE_EXTENDED) DecimalFormatter.to2Decimal(rate) + "U/h"
     else "${DecimalFormatter.to0Decimal(rate)}%"
 
+fun TemporaryBasal.toStringShort(forcePercentage: Boolean, profileBasal: Double): String =
+    if (isAbsolute || type == TemporaryBasal.Type.FAKE_EXTENDED) {
+        if (forcePercentage) to0Decimal(rate / profileBasal * 100) + "%"
+        else to2Decimal(rate) + "U/h"
+    }
+    else "${to0Decimal(rate)}%"
+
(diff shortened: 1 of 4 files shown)


core/core-main/src/main/java/info/nightscout/androidaps/extensions/TemporaryBasalExtension.kt
```diff
@@ -60,6 +60,13 @@ fun TemporaryBasal.toStringShort(): String =
     if (isAbsolute || type == TemporaryBasal.Type.FAKE_EXTENDED) DecimalFormatter.to2Decimal(rate) + "U/h"
     else "${DecimalFormatter.to0Decimal(rate)}%"
 
+fun TemporaryBasal.toStringShort(forcePercentage: Boolean, profileBasal: Double): String =
+    if (isAbsolute || type == TemporaryBasal.Type.FAKE_EXTENDED) {
+        if (forcePercentage) to0Decimal(rate / profileBasal * 100) + "%"
+        else to2Decimal(rate) + "U/h"
```
Andries-Smit, line 66: Units should be localized string

justmara: as you can see - I did not changed its representation. it already was 'U/h' there, I only added '%' variant and it does not need localization. so if this string must become localized - that would me made via additional commit, not this one.

justmara: image

Andries-Smit: No problem, it does not hurt to fix it now. If you want we can do it later.

MilosKozak:
@justmara I was thinking about features you are trying to add. Generaly I have problem with adding more items to preferences if it's usefull only for a few people.
I'd agree with configuration via adding file like "always_use_basal_percentage" to /AAPS/extra and access it in code through `interface Config`

Issue #2207, opened 2022-11-16
Possibility in cannula state, total insulin delivered to the site.


overfrenk:
Future request
--------------
Is it possible to add the amount of insulin injected into the site to the state of the cannula in addition to the time? Obviously resettable when the cannula is changed. To also get an idea of how much insulin was made at the site.

overfrenk:
@Andries-Smit My daily tdd is very variable and I never completely change the cannula and reservoir. So having 290U available with a full tank and consuming it in 2 or 4 days it would be very useful for me to know how much insulin I dispensed in the site, because beyond 120/150U or within 2/3 days the site starts to hurt and the action insulin is no longer good. The value would be convenient for me just below the cannula time or to the left of the cannula icon.

Andries-Smit:
@overfrenk The canula information is shown in the HOME and the ACTION page. The limited space on the home page statuslight bar will make it hard to place it there. Will be good enough to place it only ACTION page?
![image](https://user-images.githubusercontent.com/6724749/203147024-b8969d00-af19-4ec4-9c02-5c7b6d073535.png)

overfrenk:
@Andries-Smit PERFECT!

Andries-Smit:
@overfrenk The feature is implemented and is available in the dev branch or in the next release

overfrenk:
@Andries-Smit Thank you! I tried to generate the dev build, like almost every day, but crash...... Waiting for a fix

PR #2217, opened 2022-11-23
feat: cannula usage

Andries-Smit:
Show on the action-page the total amount of insulin that is injected into a cannula site. Some users have trouble with the effectiveness of the insulin when the cannula is used for more than X amount of insulin at the same site. Feature requested #2207
- When the "Pump site is change" action is performed, the counter will reset. 
- This counter is not used for patch pumps

image

- Refactor, underscore variables to camelCase


MilosKozak:
this doesn't count basals. number like thisn doesn't mean much

Andries-Smit:
@MilosKozak It should indeed count the basal as well; therefore, this PR leverages the TDD calculator.

yodax:
Why was this disabled for patch pumps? If you're using the TDD calculator it should be possible to use it right?

Issue #2223, opened 2022-11-27
[DEV] Crash loop when xDrip battery status of G6 is available in NS


jbr7rr:
Latest dev 305395758f-2022.11.27 (dev-c-nscv3)

Problem occurred at: 2022-11-27 09:49:59.450

Problem description:
AAPS comes in a crash loop, seems to be caused by battery status from xDrip in Nightscout.
See this line in the stacktrace:
" java.lang.NumberFormatException: For input string: "6 days (voltage: 308/289)" "

Context: 
Switched from Tsunami to latest dev, cleaned everything tried to pull data from NS to fill in latest day. after activating the NSClient plugin AAPS came into a crash loop. 
After a search on discord I came onto this post (user does not seem to report the issue here yet :( ): 
https://discord.com/channels/629952586895851530/629954570394533889/1041885298591797308

After lots of searching in MongoDB I got frustrated and just deleted the devicestatus collection. This seems to be a viable workaround for now.

See attachment for logs of the accident and a stacktrace.
[logs_and_stacktrace.zip](https://github.com/nightscout/AndroidAPS/files/10097987/logs_and_stacktrace.zip)

MilosKozak:
I need the json with data

jbr7rr:
> I need the json with data

The JSON of the DB entry I assume?

dv1:
@MilosKozak 

Example:

```
      {
         "_id" : "63828b92712ff2fa64f1ed6b",
         "created_at" : "2022-11-26T21:56:34.861Z",
         "device" : "G6 Transmitter",
         "mills" : 1669499794861,
         "uploader" : {
            "battery" : "38 days (voltage: 307/288)",
            "days" : 38,
            "daysEstimate" : "38 / 38.2",
            "lastQueried" : 1669499789395,
            "resistance" : 0,
            "resistance_status" : "GOOD",
            "status" : "OK",
            "temperature" : 27,
            "type" : "DEXCOM_TRANSMITTER",
            "voltagea" : 307,
            "voltageb" : 288
         },
         "utcOffset" : 60
      },
```

The bug is [here in xDrip](https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutBatteryDevice.java#L126), caused by [this commit](https://github.com/NightscoutFoundation/xDrip/commit/92956f73f49dc199e2539ba055ec0d647876b91e).

[...]

PR #2226, opened 2022-11-28
Various combov2 and nsclient fixes and additions to core XML strings

dv1:
* combov2: Fix DST adjustment and adjust datetime immediately when AAPS detects a DST / timezone change
* combov2: Send EventInitializationChanged
* combov2: Fix race condition when AAPS connects to the Combo while the user unpairs the same Combo
* core: Add strings to XML about datetime, timezone, DST changes (adding these to core since they are not combov2 specific and could be useful for other drivers as well)
* nsclient: Add workaround for non-int devicestatus uploader battery value (fixes https://github.com/nightscout/AndroidAPS/issues/2223 )

Also there are minor fixes to combov2 to solve TODOs in the code.

### DIFF
--- a/plugins/sync/src/main/java/info/nightscout/plugins/sync/nsclient/services/NSClientService.kt
+++ b/plugins/sync/src/main/java/info/nightscout/plugins/sync/nsclient/services/NSClientService.kt
@@ -509,10 +509,44 @@ class NSClientService : DaggerService(), NsClient.NSClientService {
                         val gson = GsonBuilder().also {
                             it.registerTypeAdapter(JSONObject::class.java, deserializer)
                         }.create()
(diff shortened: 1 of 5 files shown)


core/ui/src/main/res/values/strings.xml
```diff
@@ -97,6 +97,12 @@
     <!-- datetime layout -->
     <string name="event_time_label">Event time</string>
 
+    <!-- time/timezone/DST changes -->
```
MilosKozak, line 100: If string is not shared with other code, put it in values in module

dv1: I thought of that. But I also noticed that these type of strings are very generic. We already have some degree of string duplication among drivers. I can still move this into the module, just wanted to bring this up to think about it.

dv1:
Alright, rebased on top of latest dev and moved the strings to combov2/ .

Issue #2230, opened 2022-11-28
[DEV] Dynamic ISF autosens ratio does not respect maximum in settings


jbr7rr:
Occurs on the following builds:
dev 305395758f-2022.11.27 (dev-c-nscv3)
dev 5fa7fb34d4-2022.11.28 (dev-c-nscv3)

When enabling this function in the dynamicISF plugin:
![WhatsApp Image 2022-11-28 at 20 43 41](https://user-images.githubusercontent.com/60566713/204366893-36dde3ab-e43c-4ef6-ab57-44d59b82c3b5.jpeg)

There is a sensitivity multiplier enabled, which can multiply the basal that is used in calculations uncapped. this can result in unwanted results:
![WhatsApp Image 2022-11-28 at 20 43 45](https://user-images.githubusercontent.com/60566713/204367012-96244a2a-4f0a-4e70-8e2b-a941561e5c42.jpeg)

After some digging and discusions in discord, this is the code that is responsible for it
The ratio is calculated here:
https://github.com/nightscout/AndroidAPS/blob/a79053e5e2ebc7f70b5657ec8029439423c58a55/plugins/aps/src/main/java/info/nightscout/plugins/aps/openAPSSMBDynamicISF/DetermineBasalAdapterSMBDynamicISFJS.kt#L308
Is assigned in determine_basal.js here:
https://github.com/nightscout/AndroidAPS/blob/a79053e5e2ebc7f70b5657ec8029439423c58a55/plugins/aps/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js#L235
And is used to set basal here:
https://github.com/nightscout/AndroidAPS/blob/a79053e5e2ebc7f70b5657ec8029439423c58a55/plugins/aps/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js#L242

In master this calculation seems to be capped to the autosens max setting:
https://github.com/nightscout/AndroidAPS/blob/1de236a602d1bceae293e1e89f28d9e1f43432bd/app/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js#L331

MilosKozak:
limits mentioned are valid for autosens.
DynISF operates in much wider range and these limits are not applied

tim2000s:
[...]

PR #2231, opened 2022-11-28
Bring back autosens cap to all calculated autosens ratio's

jbr7rr:
Fix for Issue: https://github.com/nightscout/AndroidAPS/issues/2230

Restored autosens max limiter from master repo, cleaned up the code a bit by removing duplicate max limiter

### DIFF
--- a/plugins/aps/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
+++ b/plugins/aps/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
@@ -227,14 +227,24 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         //sensitivityRatio = 2/(2+(target_bg-normalTarget)/40);
         var c = halfBasalTarget - normalTarget;
         sensitivityRatio = c/(c+target_bg-normalTarget);
-        // limit sensitivityRatio to profile.autosens_max (1.2x by default)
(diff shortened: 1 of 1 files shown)


plugins/aps/src/main/assets/OpenAPSSMBDynamicISF/determine-basal.js
```diff
@@ -227,14 +227,24 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         //sensitivityRatio = 2/(2+(target_bg-normalTarget)/40);
         var c = halfBasalTarget - normalTarget;
         sensitivityRatio = c/(c+target_bg-normalTarget);
-        // limit sensitivityRatio to profile.autosens_max (1.2x by default)
-        sensitivityRatio = Math.min(sensitivityRatio, profile.autosens_max);
-        sensitivityRatio = round(sensitivityRatio,2);
```
swissalpine, line 232: This fix destroys the possibility of reactivating the exercise mode of high/low temptarget raises/lowers sensitivity. This is because it is "only" limited in one direction, "downwards" (autosens_max). As it was, it was correct.

jbr7rr: Yeah, seems I had some other assumptions wrong as well. Will remove this PR. and back to talk to some ppl about this.

PR #2237, opened 2022-11-29
Add new SMS command - LOOP LGS/CLOSED

pzadroga:
- Adding new SMS command in order to be able to switch between APS mode LGS/Closed via SMS.
- Adding also feedback about chosen mode when LOOP STATUS is queried.
- Refactor "magic strings" of APS mode and replace it with Enum.


pzadroga:
Will work on code coverage tomorrow.

MilosKozak:
are you sure that `ApsMode.OPEN.name == "open"` and not "OPEN"?

pzadroga:
> are you sure that `ApsMode.OPEN.name == "open"` and not "OPEN"?

That's a good point, thanks. I forgot that considering the patch gets merged, shared preferences value will still be lower case...
Will improve that.

MilosKozak:
generaly OK but I'd not use `.lowercase` everywhere. If potential source of future bugs.
Add migration to uppercase in MainApp and then use direct value

core/interfaces/src/main/java/info/nightscout/interfaces/ApsMode.kt
```diff
@@ -0,0 +1,19 @@
...
+
+enum class ApsMode() {
+    OPEN,
+    CLOSED,
+    LGS,
+    UNDEFINED;
+
+    companion object {
+
+        fun secureValueOf(stringValue: String): ApsMode {
```
MilosKozak, line 11: i'd prefer something like
`fun fromString(stringValue: String?) = values().firstOrNull { it.text == stringValue}  ?: UNDEFINED`

pzadroga: Done!

core/interfaces/src/main/java/info/nightscout/interfaces/ApsMode.kt
```diff
@@ -0,0 +1,13 @@
...
+enum class ApsMode() {
+    OPEN,
+    CLOSED,
+    LGS,
+    UNDEFINED;
+
+    companion object {
+
+        fun fromString(stringValue: String?) = values().firstOrNull { it.name == (stringValue?.uppercase() ?: UNDEFINED.name) } ?: UNDEFINED
+    }
```
MilosKozak, line 12: `(stringValue?.uppercase() ?: UNDEFINED.name)` is not necessary
simply `stringValue?.uppercase()`  is enough

pzadroga: Ooops, that was redundant. Simplified in next commit.

PR #2258, opened 2022-12-05
 pump status and receivers for SMS reply option

### LINKED ISSUE #2257: Too long SMS reply messages
Currently SMS communicator has 2 major problems with replies:
1. It adds `activePlugin.activePump.shortStatus(true)` to every successful reply
2. It is forced to send successful replies to every number in allowed numbers list.

For some pumps this is not an issue - we had no any pump status in reply messages while were using Combo pump. But for some it is. For example, this is reply message in Russian locale with Omnipod Dash pump:
```
Болюс 0,80ед. подан успешно
Предыдущее соединение: 0 мин. назад
Предыдущии болюс: 0,80 @ 12:04
Временн 0,00U/h @11:41 24/120'
Резервуар: 44ед
```
It contains 142 characters, so it is split onto 3 SMS messages (SMS can contain up to 70 non-ascii characters), and those 3 SMS are sent to each parent. So every successful reply takes 6 SMS messages to be served. Thats expensive :)

justmara:
fixes #2257 

1. made pump status optional in successful reply messages. by default it is added to keep current behaviour unchanhed
2. added an option to send successful reply messages only to comand sender. defaults to false to maintain current behaviour

### DIFF
--- a/plugins/main/src/main/java/info/nightscout/plugins/general/smsCommunicator/SmsCommunicatorPlugin.kt
+++ b/plugins/main/src/main/java/info/nightscout/plugins/general/smsCommunicator/SmsCommunicatorPlugin.kt
@@ -448,7 +448,8 @@ class SmsCommunicatorPlugin @Inject constructor(
                             override fun run() {
                                 if (!result.success) {
                                     var replyText = rh.gs(R.string.smscommunicator_tempbasal_failed)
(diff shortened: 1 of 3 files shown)


MilosKozak:
@justmara 
generaly i'm against adding more options for non important things. there are too much already and this makes AAPS hard to setup
But i'd accept configuration via file in extra directory

PR #2268, opened 2022-12-08
Smoothing plugin

MilosKozak:


MilosKozak:
not activated vs activated visualization
image
image

macarolobe:
With or without smoothing my bg is update 4 minutes after xdrip. So new bg comes when old bg is 9 minutes old and the new bg is already 4 minutes old.

justmara:
Smoothing's primary target is to leverage jumpy sensor values and give AAPS streamlined data, that can be used for predictions. With not only 'closer to real' BG values, but also cleaner deltas. Second screen shows still jumpy values with deltas that could easily lead to wrong predictions.

vanelsberg:
Wow! Love this @MilosKozak @justmara:  If this works as designed it definitely is a valuable new feature for AAPS
Can't wait to test this on dev! (even while on G6/BYODA)

justmara:
Second graph looks crappy, much worse than original Tsunami's smoothing. Didn't check why is this, but this graph would still give bad delta values for AAPS's predictions.
The main purpose of smoothing is to give AAPS better data to work with. And deltas are much more valuable than BG level itself. IMHO a bit slower but more reliable delta changes are in priority for advanced AAPS algorithms.

MilosKozak:
@justmara alg is the same :) just different integration
I generated really crappy data to see how it works

swissalpine:
Works for me as it should or as Tsunami Smoothing does:
G6 -> xDrip -> AAPS

macarolobe:
Don't work for me. Bg arrive 4 minutes later than xdrip (FSL2)

MilosKozak:
> Bg from xdrip is ok in xdrip module but on main screen it's not last bg but bg before.

Could you create an issue with screenshots and logs?

macarolobe:
> Could you create an issue with screenshots and logs?

I rebuild and bg is ok now

justmara:
History browser will be recalculating smoothing for every BG point loaded too? Will anyone survive the loading then? :)

Der-Schubi:
I'm running this for two days now and hat the yellow triangle showing "Recalculated data used" the whole time. Today its gone. Was this from switching on the filtering?

parapenT1sta:
@MilosKozak Is it possible to have SMB always active for sensors as Libre 2 when the plug-in is on or will keep as it is right now? No chance for L2 get SMB always on?

rICTx-T1D:
While testing I noticed something...

The clock always gets the unsmoothed values, while the start page shows the smoothed value. The bonus calculator also uses the unsmoothed value. Why is this?

MilosKozak:
> I was reading through the PR. And I find it hard to differentiate between what is smoothed data and what not. If I understand correctly now the in memory values are smoothed if a smoothing plugin is active. the bgReadings in the datastore are unsmoothed? Is there are more clear way to differentiatie between this? As for example the flat BG check is that reading unsmoothed value's? Based on my reading of the code I would say yes. But it's hard to read and get that information currently.

Buckedted data is smoothed (ie data after recalculation). db values and bgReadings are unchanged

MilosKozak:
> History browser will be recalculating smoothing for every BG point loaded too? Will anyone survive the loading then? :)

There is no visible delay because everything is done only in memory

MilosKozak:
> I'm running this for two days now and had the yellow triangle showing "Recalculated data used" the whole time. Today its gone. Was this from switching on the filtering?

Triangel dependends only on timestamps. No relation to smoothing

MilosKozak:
> @MilosKozak Is it possible to have SMB always active for sensors as Libre 2 when the plug-in is on or will keep as it is right now? No chance for L2 get SMB always on?

Primary reason why is SMS always disabled for Libre is not "non smoothed data"

Der-Schubi:
After switching the smoothing plugin on/off or switching the smoothing algorithm the insulin and calculator buttons on main screen are missing (as directly after start of aaps) and I have to force close aaps to get them back.
Anyone else? (Using Dana I)

vanelsberg:
Have been running the smoothing branch for 24 hours (DASH/G6).
Had just changed sensor. First thing that I noticed the BG-value gap of 2 hours without BG was smooted into a nice steady line in overview. From there the usual bumps in BG values while G6 was stabilizing.
Did not prevent 2 sudden G6 drops causing false hypo-alarms, but at least it damped the steep down/up trends.
Until now, did not noticed any problems.

@MilosKozak: looks good! Love it, even with G6. About to update to todays commits.

jbr7rr:
Some feedback on the action items:

- verify HistoryBrowser
    > Value's are smoothed, also from few days ago (see screenshot)
WhatsApp Image 2022-12-14 at 19 35 03


- verify what values is calculator using
    > For me it is using unsmoothed value's. what is desired? I think the smoothed value? 
    
WhatsApp Image 2022-12-14 at 20 00 07
WhatsApp Image 2022-12-14 at 20 00 06


- verify data sent to the watch
    > Unsmoothed value's are sent to the watch (delta's are the ones calculated over smoothed data), what is desired? Personally i dont care
    > I have pretty good input data so its hard to see on the graph, but you can see BG value is the unsmoothed one.

WhatsApp Image 2022-12-14 at 20 04 00
WhatsApp Image 2022-12-14 at 20 03 56

All tests and screenshots are taken with exponential smoothing, as that gives the most clear difference in BG value

vanelsberg:
Can confirm seeing @jbr7rr 's feedback.

@MilosKozak 
> Questions to answer:
> force/disable/hide for some BG plugins
Could you please elaborate on this: is this about different smoothing algo's here?

macarolobe:
Bg is not Last bg but the one before the Last. In xdrip module it's the Last but not in the home tab.

vanelsberg:
>Forcing the plugin to no Smoothing when BYODA G6 is active is a possible solution
For the current smoothing plugin I am not at all sure if I'd vote for that!
Have had enabled "Average smoothing" for the past week while using G6 with. Did not notice any issues with the smoothed data. Actually it seems to work pretty good....

-  Jumping (sudden BG value changes by +/- 1,5 mmol) while on the first 2 days of a new sensor were perfectly smoothed out, dampening the effect of these jumps.
- Same for sudden changes in BG value when calibrating the G6
In both cases the  BYODA/G6 does not smooth out these data jumps!

So for now: please keep the plugin as an option for G6

vanelsberg:
@MilosKozak Build on latest smoothing commits (#37510088) is failing for me:
```
:pump:omnipod-dash:dataBindingMergeDependencyArtifactsFullRelease
Could not resolve com.github.guepardoapps:kulid:2.0.0.0.
```

MilosKozak:
... processed open TODOs

vanelsberg:
Confirming: builds ok

jbr7rr:
Tested latest commits:
- Bolus wizard: Shows the smoothed values, nice
- Watch: Value shows smoothed value, graph seems to show the unsmoothed graph. (This is fine by me, but I can understand if it will be confusing for some users)

PR #2285, opened 2022-12-13
Dynisf: resolve missing TDD

MilosKozak:


tim2000s:
Makes sense. Is it worth attempting to pull 7 days of insulin data from Nightscout before reverting to missing days?

MilosKozak:
it will be possible with NSClient v3. In progress now ....

Issue #2302, opened 2022-12-26
time of sync on xDrip and AAPS 3.1.0.3-dev-f dont match


idalioreis:
last AAPS verson DEV  (3.1.0.3-dev-f), the time between sensor reading is not sync on xDrip and AAPS even with Smoothing disabled. AAPS reading from xdrip

Xdrip 09.12.2022

![aaps](https://user-images.githubusercontent.com/48919798/209567954-7a6674bf-9274-4b02-a91a-3eca91192ed8.jpg)
![xdrip](https://user-images.githubusercontent.com/48919798/209567963-bc547be1-e62d-476e-a541-7c5e944ec322.jpg)

ct2iqj:
I have de same problem.

MilosKozak:
not possible to resolve without logs
what cgm?

MilosKozak:
... and timestamps on xDrip tab match xdrip?

idalioreis:
> not possible to resolve without logs what cgm?

CGM is libre2

with smoothing on, the time of reading on aaps increase
![image](https://user-images.githubusercontent.com/48919798/209646373-f7ce39ac-c976-4bb5-a25e-ab793604ba55.png)
![image](https://user-images.githubusercontent.com/48919798/209646402-cfeaa612-58aa-4ea5-bc32-4553a2be0409.png)
[AndroidAPS_LOG_1672132132250.log.zip](https://github.com/nightscout/AndroidAPS/files/10307298/AndroidAPS_LOG_1672132132250.log.zip)


with smoothing off and when xdrip read sensor (0m) the AAPS change to 1m
![image](https://user-images.githubusercontent.com/48919798/209646458-cb26cc1a-5091-44e7-affe-01307387f4f9.png)
![image](https://user-images.githubusercontent.com/48919798/209646471-aa27a08b-0721-4b95-9c1f-6c085ab41cff.png)
[AndroidAPS_LOG_1672133977692.log.zip](https://github.com/nightscout/AndroidAPS/files/10307301/AndroidAPS_LOG_1672133977692.log.zip)


![image](https://user-images.githubusercontent.com/48919798/209647467-e0b14ad0-a9d1-445f-9935-6cdae57b4eaa.png)
![image](https://user-images.githubusercontent.com/48919798/209647513-cd8ce811-359f-4c60-a36f-09050f5cbe91.png)

MilosKozak:
there is BG every 6 min in AAPS

[...]

PR #2303, opened 2022-12-27
Fix recalculated data time sync

jbr7rr:
Fixes https://github.com/nightscout/AndroidAPS/issues/2302

The recalculation sets the time too far back in the past most of the times, which confuses users, this sets the start of the recalculated data less back in the past.

This PR is to minimize the times that the user encounters this added lag, it cannot be avoided completly

referenceTime is set by the first available readings by the given currentTime, however always doing this:
```
var currentTime = bgReadings[0].timestamp - bgReadings[0].timestamp % T.mins(5).msecs()
```
Causes that the recalculated readings will always lag behind as referenceTime is set to a time that is not the actual currentTime, even if further intervals are close to 5 minutes.

However removing ``- bgReadings[0].timestamp % T.mins(5).msecs()`` is relying on adjustToReferenceTime() to work as expected. Unit tests has been made to test verify behavoir. During unit testing it is found that adjustToReferenceTime() did not work as expected. This has been fixed in this PR as well.

previous behavoir of adjustToReferenceTime():

example 1:
```
someTime = 46
referenceTime = 5
diff = 46 - 5 = 41
diff = 41 % 5 = 1
diff = 1 
return 46 + 1 = 47 // Expected 45
```

example 2:
```
someTime = 48
referenceTime = 5
diff = 48 - 5 = 43
diff = 43 % 5 = 3
(if diff > 2.5 min) diff = diff - 5min = -2
return 48 + - 2 = 46 // Expected 50
```

Because of this ```bgReadings[0].timestamp - bgReadings[0].timestamp % T.mins(5).msecs()``` (which we changed), this didn't happen before with recalculated data, but could happen in other cases where this function was used.

New behavoir of adjustToReferenceTime():

example 1:
```
someTime = 46
referenceTime = 5
diff = 46 - 5 = 41
diff = 41 % 5 = 1
(if diff < 2.5 ) return 46 - 1 = 45 
```

example 2:
```
someTime = 48
referenceTime = 5
diff = 48 - 5 = 43
diff = 43 % 5 = 3
(if diff > 2.5 ) return 48 + abs(3 - 5) = 50
```


jbr7rr:
In the current state, there will still be a delay sometimes for the user.

Currently testing this to see if the modulo thing can be removed completly
https://github.com/jbr7rr/AndroidAPS/commit/4b05dcd5888cddb76cebfc7322a8ad6cf24409fd

PR #2309, opened 2022-12-28
chore: update docs base URL

Andries-Smit:
The new URL for documentation is https://wiki.aaps.app/


Philoul:
In most links you kept the `/EN/` after `/latest/`
I suggest to remove it (in crowdin, links doesn't work in other langages if you just replace `/en/latest/EN/...` by with `/xx/latest/EN/...`)

PR #2320, opened 2023-01-01
Trim leading and trailing whitespaces in incoming SMS messages

mushroom-dev:
It will allow users to use sms commands even when accidentally adding extra space at the end


MilosKozak:
resolve conflict pls

PR #2368, opened 2023-01-20
Fix DynISF TDD when 7Day or last Day is missing

Koeppchen:
DynISF uses risky TDD values for calculation of sensitivity

In DetermineBasalAdapterSMBDynamicISFJS.kt the TDD is calculated.
If the 7 day average TDD is not available the weighted 8 hours is used.
This will lead to a very low TDD calculation after hours if not eating,
for example after sleeping for 8 hours and as a result to high calculated ISF.
The opposite is true for two meals during the last 8 hours or even 4 hours,
In this case the calculated TDD will be high and the resulting ISF low.
A possible fallback is to use
(tddWeightedFromLast8H * 0.33) + (tddLast24H * 0.67)
this is possible because the tdd calculation and dynamic ISF claculation is only done
if 24h tdd is available. The weights are 1/3 for the Last 8 Hours and 2/3 for the last 24 hours. 
This is what is included in this pull request.


tim2000s:
I think 8 hours or a mixture of rolling and 8 hours are both a little too volatile and can result in very high TDD values that overstate the TDD dramatically and can result in very low ISF numbers. 

I think we should defer to profile basal in this context and use something like `profile basal / 0.35 ` as the 1D and/or 7 day when either is unavailable, as it's about the middle of the various accepted references.

Koeppchen:
@tim2000s Using Basal as a base as long as not 1d and 7d stats are not available sounds like a good alternative to me.
AAPS Code is new to me, and this is just a very small part i have looked at. But I'll look at the code to understand it better. But i think first i have to understand how the tests are coded because there was no code coverage.

Koeppchen:
So far i could not find out how the tests work. Also the basal profile is not sent to the invoke function. At this moment i am far away from having the knowledge to fix stuff like this.
The only idea i would be able to implement that came to my mind would be to use ISF from the profile as long as the 7 day stats are not available. This would mean during the first 7 days after a database reset DynISF will use the ISF from profile.

tim2000s:
I think something like this is probably safer:

```
val basal = this.profile.put("current_basal", basalRate)

        var tdd =
            if (tdd1D != null && tdd7D != null) {
                (tddWeightedFromLast8H * 0.33) + (tdd7D * 0.34) + (tdd1D * 0.33)
            }
            else {
                (tddWeightedFromLast8H * 0.33 ) + (((( basal * 12 ) * 100 ) /21) * 0.67 )
            }
            ```
            
           Although I'm not sure that's the correct way to call the current profile basal as the component to use in Kotlin, as I'm less familiar with the Kotliln than the Javascript.

Alamo04:
I lost my TDD and that was a chance for me to view how DynISF works if no 7 day TDD is avaible..
So I switched to DynISF after 12h and had an eye on it.

In my opinion DynISF is danger in that case. 
It changed my ISF from 60 to 13 whats way to much and never seen.
A bad hypo would be sure in that case. Changed back to smb realy fast.

I love DynISF, if 7 day TDD is avaible.
But for a normal user, or if you lost your TDD without to be aware of that, its a realy dangerous Algo in his current state.

Koeppchen:
i do not think it is good to use basal for tdd calculation. But i think the following might be a good solution to the problem:
I have removed the possibility to use DynISF when there is no tdd for the last complete day available.
If it is not available sensitivity from profile will be used.
It should be allowed to have days without stats for older days than the last day, if days for the last week are missing only those days available will be used for calculation.

Koeppchen:
I have done some tests on my backup phone with my latest changes. Just to have the situation where statistics are missing.  But the last day was automatically recovered from nightscout and statistics for one day was available. In this case the result with my changes are really good. Also i am using this as my daily driver for some weeks now.

PR #2387, opened 2023-02-03
Limit DynamicISF sensitivity detection to autosens min/max restrictions.

yodax:
I noticed when using DynamicISF that the sensitivity detection wasn't limited by the autosens limits. For the OpenAPS / AMA this is done in a different class. Since that logic wasn't used the limits weren't applied. 

This change applies the limits correctly. I've tested this for about a week and it seems to work.

### DIFF
--- a/plugins/aps/src/main/java/info/nightscout/plugins/aps/openAPSSMBDynamicISF/DetermineBasalAdapterSMBDynamicISFJS.kt
+++ b/plugins/aps/src/main/java/info/nightscout/plugins/aps/openAPSSMBDynamicISF/DetermineBasalAdapterSMBDynamicISFJS.kt
@@ -40,6 +40,8 @@ import java.lang.reflect.InvocationTargetException
 import java.nio.charset.StandardCharsets
 import javax.inject.Inject
 import kotlin.math.ln
+import kotlin.math.max
+import kotlin.math.min
 
 class DetermineBasalAdapterSMBDynamicISFJS internal constructor(private val scriptReader: ScriptReader, private val injector: HasAndroidInjector) : DetermineBasalAdapter {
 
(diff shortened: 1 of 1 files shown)


jbr7rr:
This has been suggested by me before, @MilosKozak stated it was not needed. But as you opened this discussion can you explain why this is needed?

Maybe @tim2000s can also shine some light on this?

See of my previous issue and PR for this: 
https://github.com/nightscout/AndroidAPS/pull/2231
https://github.com/nightscout/AndroidAPS/issues/2230

yodax:
What I have observed is that it usually works just fine. However, when my son gets sick or recovers from being sick it goes wrong. In the example below the multiplier almost reached 140% for the basal on that day.

TDD

 This is working in tandem with the higher glucose of being sick already being corrected by DynISF. The reason I started investigating it that a few weeks ago he was also sick and during that time a few day totals reached 48 units and after that while recovering it was 18 units. This is a big spread and caused a much higher percentage.

On both ends of the peak it was either correcting way to much or way to little. Having the algorithm limited by a safety setting can prevent unwanted high/lows and forcing him / me to make the conscious decision to increase/lower the profile temporarily.

Another area is where he eats loads of carbs, this would result in a high 24 hour total, but should not result in a 24h increased resistance. The dynamic ISF part does this very well imho with the sliding window and the last 4 hours contributing relatively more. The sensitivity calculation does not do this.

So in our case I don't think basal should be adjusted by more than 120% or less than 70% without him/me making that decision. 

In the case of my son, I've seen a big bump in TIR with dynamicISF and TBR is even better. It has given us way more rest 👍

MilosKozak:
see my answer here
https://github.com/nightscout/AndroidAPS/issues/2230#issuecomment-1329765101

yodax:
> DynISF operates in much wider range and these limits are not applied

I think the wider range for the ISF part makes sense. I have a problem with the basal part, having a day where you eat a lot of carbs doesn't mean the basal requirements go up for 24 hours.  Or that when you've been sick for a couple of days, and after recovery you don't suddenly need 30% of the normal basal. Now it is only restricted by the smb safety- and max basal settings. 

In the situations I described the only solution for us was to disable sensitivity detection until the "anomalous" days were out of the 7 day window. 

@MilosKozak are you in principle disagreeing with that, or would you be open to making this configurable?

MilosKozak:
My opinion is that limiting it would mean unsufficient reaction in most other situations. There are and always will be situations where radical change of sensitivity or insulin needs will need manual intervence because we detect it from BG which will have always some delay

yodax:
But this change affects the sensitivity part of dynamicISF. It’s not dependent on BG at all, only TDD.

Alamo04:
@yodax 
I understand your point, have the same problem if I get sick over a few days, or in other cases as 14 days in Italy. :D
But you can't handle that with DynISF, because DynISF is using your TDD. Also Sensetivity works in a spezial DynISF mode.

**I recommend:**
If you have a problem with your TDD over a few days, you have to switch back to normal SMB/AMA until your situation is in normal range. Same for Database reset, or complete new installation of AAPS.

**I recommend everyone to use a profile, that has also your own tested ISF, so you can allways switch to one of the other algorithm (SMB/AMA) in such a case.**

PR #2433, opened 2023-02-22
Pump common v1.2

andyrozman:


MilosKozak:
* this diff is really huge.  I'm not able to evaluate impact just by reading it. thus drivers would have to be retested. Do you really need this to be merged in 3.2?
* walk through diff and eliminate use of !! operator (using should be rare). it goes against kotlin philosophy to eliminate NPE

andyrozman:
Nothing has really changed... I just moved files a little and add few new ones (in pump-common)... 

Every dependency stayed the same (no code changes on existing code). 

Only big changes is that I moved one key from pump-common to core, and so removing "insight" dependency on pump-common... 

The new classes are not the last version yet, I just wanted to add them there, so that they could be used by new pump. 

We could actually wait for this to be merged, after 3.2 release is out, if that is what you prefer. Few weeks won't make much difference.

PR #2441, opened 2023-03-01
[fix] change string to make it clearer what it does

grahamjenson:
When doing Objective 1, this string was presented without context and it took me more than an hour (and a search through the source code) to figure out what it meant. At first I was searching though the NSClient config, but it is in the pump config. Adding the work "pump" would be a good clue for this.


MilosKozak:
you were targeing master. replaced by https://github.com/nightscout/AndroidAPS/commit/d77c87a56de58768ff4848395c6f7c9f0a8b41d1

grahamjenson:
Cheers. Will target dev next time.

Issue #2382, opened 2023-01-31
[3.1.0.3-dev-g] [ComboV2] User Entry Log


markvader:
ComboV2 driver automatically recognises insulin cartridge change && (if newLevel > currentLevel)

Entry automatically made in the Careportal log and User entry log recording the change.

**The Careportal log entry shows** 
1. the time "hh:mm"
2. the title "Insulin Cartridge Change"
3. a note that the "Insulin reservoir change inserted automatically by combov2 driver".

All okay here

**The User Entry log shows** 
1. the time "hh:mm:ss"
2. the title "CAREPORTAL"
3. The combo icon
4. the date and time MM-DD-YY HH:MI:SS and the title "Insulin Cartridge Change"
5. the note "Insulin reservoir change inserted automatically by combov2 driver" is right aligned to the same width as the combo icon

**Issues:** 
1. the date time duplication.  In the entry I have their is a 1 second difference between the two entries, is this why it appears twice.  Does their need to be an increase in tolerance for timedelta between phone and pump?
2. The title on 3.1.0.3 for combov1 is 'Reservoir Change' rather than CAREPORTAL
3. Will removing the date time duplication give enough space to the note.

I can see that pumps like the DanaRS to not use the note parameter in its pumpSync.insertTherapyEventIfNewWithTimestamp record, however I see it's value in careportal, particularly as users switch over from old combo to new.

![careportal   user entry](https://user-images.githubusercontent.com/217953/215629826-34c2a872-6b53-4852-ac5e-6e8b55bfbce0.jpg)

@dv1

dv1:
> The title on 3.1.0.3 for combov1 is 'Reservoir Change' rather than CAREPORTAL

I am confused. Where does the combov1 (I assume you mean the old Ruffy based driver) use this? It does not enter any treatment.

markvader:
agreed, the ruffy driver is all user entry. 
![combo-careportal](https://user-images.githubusercontent.com/217953/215717980-45b78e88-72a4-464a-874a-ff373c266363.jpg)

I got someone to send me the danaRS screens, which has automatic cartridge change recognition (time: 22.21).  This user has also manually recorded the change (time: 22.24) (hence two different entries)
![danaRS](https://user-images.githubusercontent.com/217953/215725620-7d1b3c39-eb39-4582-921f-5839d0a02517.jpg)

dv1:
[...]

PR #2457, opened 2023-03-12
combov2: Fixes for the refresh button, unsafe casts, crashes due to Bluetooth, improved warning screen handling, and additional language support

dv1:
* Added UserEntryLogger.log overload that accepts a timestamp to avoid having to rely on multiple dateUtil.now() calls
* Removed the note from the automatic battery and reservoir replacement insertions into the DB (fixes https://github.com/nightscout/AndroidAPS/issues/2382)
* The Refresh button in the combov2 fragment wasn't being enabled when the driver was in the error state, preventing the user from informing the driver that the pump error got dealt with
* Improved warning screen handling by interpreting the text to the right of the check mark to distinguish in which state the warning is (the snooze vs. the confirm state)
* Added support for pumps whose language is set to Slovenian or Lithuanian
* Fixed wait in onStop() and removed a number of unsafe casts to address issues like trying to access a null bluetoothInterface instance
* Added exceptions and exception handling for when the Bluetooth adapter is missing or disabled; amongst other changes, UI notifications are shown now (both in the overview and in the pairing activity)


MilosKozak:
@dv1 tests are failing

dv1:
@MilosKozak Yeah, there was a typo in one test. Fixed.

PR #2464, opened 2023-03-19
Fixes to combov2 pairing PIN entry UI, the pairing activity, and TDD history retrieval and activity background

dv1:
* The pairing UI had a PIN entry field with a hardcoded width. Change this to instead use a weight based layout to make it adapt to the screen width better.
* On Android 12+, the pairing activity needs to ask for `BLUETOOTH_CONNECT` permissions. Currently, it doesn't.
* Allow for retrieving TDD history in the suspended state
* Fix activity background to no longer show the AAPS logo (which was making it difficult to read the white text)


MilosKozak:
@dv1 maybe would be a worth to use `interface BlePreCheck`

dv1:
@MilosKozak Ah, good point. I'll try that instead.

dv1:
@MilosKozak I use a custom counterpart to `BlePreCheck` that is designed for Kotlin coroutines. It now asks for permissions when the pairing activity is up and running.

Issue #2524, opened 2023-04-30
[DEV] DASH: Temp basal cancelled/stopped by pod for unkown reason, AAPS History not updated


jbr7rr:
Build: 34e7a9ad52-dev-i (personal branch) merged at 6cfd5b0c1673a79d9d33498ee2a7031b54ccd6c4
Pump: Dash
Time: Temp basal start at 11:52:37.814. Reconnect where we find the changed deliveryStatus from deliveryStatus=TEMP_BASAL_ACTIVE to deliveryStatus=BASAL_ACTIVE at 17:04:08.902
Logs: [AndroidAPS_LOG_1682783403444.log.zip](https://github.com/nightscout/AndroidAPS/files/11361767/AndroidAPS_LOG_1682783403444.log.zip)

Context: On this day I went to the pool, set a 10 hour long basal at 11:52, did a few bolus after that, but did not change the basal or profile. The phone was stored in the locker so had no contact with the pod for some time, at 17:00 the pool closed so I went home. In the car I noticed my dash ticking while AAPS reported still a 0 temp basal. so I checked the status screen. found this:

<img src="https://user-images.githubusercontent.com/60566713/235364857-dc7449a9-e093-4f7e-8c4b-4ac0dd0ff386.jpeg" width="200"> <img src="https://user-images.githubusercontent.com/60566713/235364858-e473df41-9dc3-4dbb-b21e-44601be7ee84.jpeg" width="200">  <img src="https://user-images.githubusercontent.com/60566713/235420568-12ae2511-798d-42f6-80da-6f8005143fa7.jpeg" width="200">

As you can see according to AAPS there was still a 0 temp active. But on the dash tab the deliveryStatus of the pod was BASAL_ACTIVE. I have not received any alarm other then no BG and no pump connection. And I know for sure that the DASH pod was delivering Insulin (looking back at events it makes sense I had to correct with food more often then I used to after 16:00)

I see 2 problems here:

1. Why did the pod stop the TBR? I have no real answer to this other then some guesses: 
    a. maybe the pod got damaged during the wild water slide, the pod was pretty banged up and had some faint yellow stains inside. 
    b. Or maybe the pod cancels the TBR when the controlling device is out of range for x amount of time? I intent on testing this hypotheses

[...]

PR #2549, opened 2023-05-22
DASH: handle tbr out of sync

jbr7rr:
Fix for AAPS not recognizing TBR state change of dash #2524 : 

Handling pump TBR cancellation and unknown TBRs in AAPS. Added total delivered insulin to History view for user information.

Note: While it would be ideal to estimate the timestamp of TBR cancellation by DASH, doing so would introduce significant complexity. This situation should not occur under normal circumstances. If there are cases where it does happen and cannot be avoided, we can consider adding this feature in a future pull request (PR).

Screenshot of changed history view:
WhatsApp Image 2023-05-22 at 09 59 58


MilosKozak:
tested by someone?

jbr7rr:
Tested the following scenario's myself:

- Move time forward, TBR will be active on POD, but is stopped for AAPS > On next connection: TBR is cancelled on POD and ërror message is shown
- Force faulty behavior by setting all TBR commands to 30 minutes in the code, > After TBR is expired for POD and AAPS still thinks TBR is running: On next connection: TBR is updated and error message is shown
- Normal usage, no unwanted warnings or TBR cancellations etc on 2 different devices and spanning several pods

Let me know if you need more, I can ask a few ppl to help testing

Issue #1724, opened 2022-05-11
Overlapping basals lead to incorrect IOB [Medtronic]


ArthurusDent:
Continuation of #1597 for Medtronic.

### I'll lead with the instance for which I actually have logs:

What can be seen here is a jump of IOB by ~ -0.4 units. I suspect that between 17:05 and 17:10 pump history had been read and records were updated. AndroidAPS did not know it had already set a zero temp earlier, which explains the jump in IOB. 
![Screenshot 2022-05-09 174153-markiert-zugeschnitten](https://user-images.githubusercontent.com/6848629/167875218-2a700826-39a8-4b29-b42d-3f4088012bca.png)

Pump history shows that AndroidAPS indeed set a zero temp twice.
![Screenshot_20220509-171000_AndroidAPS](https://user-images.githubusercontent.com/6848629/167876316-29a0dc60-4a49-4332-af0b-7a3df1451ed4.png)

Logs: 
[AndroidAPS._2022-05-09_16-33-16_.zip](https://github.com/nightscout/AndroidAPS/files/8670841/AndroidAPS._2022-05-09_16-33-16_.zip)

### The obligatory negative absolute insulin (no logs):

If I remember correctly, none of the 120 min zero temps actually ran the whole 120 min, I did stop them manually. Take a look at the two records I marked as A and B. The screenshots are only two minutes apart but the A-record was altered after the fact.

![Screenshot_20220424-165010_AndroidAPS](https://user-images.githubusercontent.com/6848629/167877319-42daa192-d33c-4045-a117-3204b602ef3e.png)

![Screenshot_20220424-165230_AndroidAPS](https://user-images.githubusercontent.com/6848629/167877354-18bd8e68-7480-473b-b94e-531d10882a52.png)

![Screenshot_20220424-165716_AndroidAPS](https://user-images.githubusercontent.com/6848629/167877384-c33b3f0b-0d80-4ca0-8cb4-01a890fcb45a.png)

![Screenshot_20220424-165726_AndroidAPS](https://user-images.githubusercontent.com/6848629/167877404-81d478cc-a766-4fa1-a130-ba550fd35144.png)

### System:
AndroidAPS 3.0.0.1 479bfc410
Oneplus 3T
Android 11 (LineageOS 18.1)
Medtronic 523
Dexcom G6 with BYODA

MilosKozak:
can you reset db and try dev?

ArthurusDent:
I'm on it.

[...]

PR #2550, opened 2023-05-22
Skipped creation of "dummy cancel temp basal" records for temp basals which are already paired with real cancels

yp1ripe:
[at least partial] fix for  #1724 

The "for loop" commented as   // fix for Zero TBRs in the 

`fun createTBRProcessList(entryList: MutableList<PumpHistoryEntry>) : MutableList<TempBasalProcessDTO>
`

should skip zero temps that already have their corresponding CancelTBRs paired in the previous section of the same function.


MilosKozak:
did someone test it?

yp1ripe:
> did someone test it?

I am.
It is running on my daughter's 3.1.0.3 for about 2 weeks now and it is much better than without the patch.   Without the patch IOB is off by 0.1-0.3 units quite often.

andyrozman (approved):
If it was tested and works then it should be merged. 

I hope none of other models will work worse with this change.

Issue #2551, opened 2023-05-22
[DEV] DASH: Driver hangs at handshaking for a long time (5mins +)


jbr7rr:
Build: personal branch based on https://github.com/nightscout/AndroidAPS/pull/2549
Pump: Dash
Time: 01:23:23.260
Logs: [AndroidAPS_LOG_1684713297602.log.zip](https://github.com/nightscout/AndroidAPS/files/11534685/AndroidAPS_LOG_1684713297602.log.zip)

The driver kept on handshaking for a long time. 

I believe this is caused to a race condition because the driver already states it's disconnected while the actual disconnecting still has to happen. See logs at given timestamp

Fix is being tested and can be found here: https://github.com/jbr7rr/AndroidAPS/tree/dash-fix-race-condition-on-disconnect
Will make PR after testing

PR #2554, opened 2023-05-24
DASH: Fix race condition on dash disconnect

jbr7rr:
Fix for: https://github.com/nightscout/AndroidAPS/issues/2551

Fix race conditions when disconnects happen at different times


MilosKozak:
did someone test it?

jgslade:
> did someone test it?

I have tested it. Everything running smooth.

vanelsberg:
Have been running this for 24 hours. Code looks clean & OK. 
So I think it is safe/ok to merge with dev.

Issue #2567, opened 2023-06-04
[DEV] Pump connection loss alarm is not working


jbr7rr:
AAPS will not alarm after 30 minutes of no conneciton to the pump as expected

This check in KeepAliveWorker:
```
runningProfile.value.originalEnd < dateUtil.now())
```
Introduced in 01bb554651f2dbfafdec7f069dd99f2ed16cc578
Always returns true when the current profile has no duration, always triggering the parent condition and preventing a statusUpdate to the pump
The condtition will return false when the profile has a duration, and triggering the statusUpdate for the pump

This results in the pump timeout not triggering for some pumps when a profile with no duration is set (currently facing this with a pump under development, but this condition also triggers when using virtual pump)

I think its due to runningProfile.value.originalEnd only being set when there is a duration present.

I hope this makes sense, otherwise I'm happy to explain futher in DM

PR #2569, opened 2023-06-04
Set EPS.originalEnd to Max to fix KeepAliveWorker conditions

jbr7rr:
Fix for: https://github.com/nightscout/AndroidAPS/issues/2567


jbr7rr:
It's not this simple unfortunately :(

PR #2639, opened 2023-07-27
Redesign logo 

thiagomsoares:
Redesign the aaps logo.
Changes: Improvements in design and change from Androidaps to aaps. Blue logo was the winner in the voting in the Facebook group.


ArthurusDent:
The SVG has a small problem. It's not that visible, if you don't zoom in, though.

logo-vector-2023.svg
grafik

PR #2653, opened 2023-08-09
combov2: Additional checks for edge cases

dv1:
* Check that Bluetooth is actually supported by the hardware
* Check that the pairing PIN is 10 digits long


MilosKozak:
tests are failing

dv1:
Do you mean `DetailedBolusInfoStorageTest` ? Because I don't see how these changes broke it.

MilosKozak:
then only merge latest dev

PR #2660, opened 2023-08-11
New ic_blueowl icon

Philoul:
I finally put a round background to have the same size than AAPS icon...


Philoul:
Yellow Round icons created, but I saw at the end that within build.gradle we could set 2 different icons (one round and one square)
=> I put the round one everywhere (yellow for NSClient and Blue for NSClient2...) Is it ok or do you prefer keep also the square one ?

MilosKozak:
Round are perfect

MilosKozak:
improved by https://github.com/nightscout/AndroidAPS/commit/b2d9f1c39bd1d352c2b590dd1d77311f74263d3d

PR #2687, opened 2023-08-22
Wear CWF : little bug fix + Code improvement

Philoul:
Now all json keys are included into one single file within enum keys.
- I will first finish user interface of Wear Plugin (I want to add an "Info" button to show all informations linked to loaded Watchface), then I will see if I move these "HardCoded keys" into one String file (translatable false)
I also renamed all the CustomWatchfaceXxx classes to CwfXxx to shorten the name and improve code readability
Minor bug fix in second and BGI visibility in cwf


Philoul:
Wrong target, sorry

PR #2717, opened 2023-08-29
Fix for project's AndroidTests

buessow:
With these fixes, I can run gradlew connectedFullDebugTest to execute all AndroidTests in the project. This is the first step to get common coverage numbers and to enable AndroidTest in CI. 
I'm not sure why the fixes to missing colors are need only now.


MilosKozak:
Do you how studio can run androidTest so fast?

buessow:
I think it's fast once everything has warmed up. It probably only needs to copy some binaries onto the virtual device. A fast build machine helps tremendously :-)

MilosKozak:
code is in dev
thanks @buessow

PR #2719, opened 2023-08-29
Wear CSW Add AllCaps and dayName Month format customization

Philoul:
3 optional keys added in json file to allow Upper case for text fields :

within TextViews
"allCaps": true

day formating or month formating (global key outside views)
"dayNameFormat": "E" to "EEEE" (short or long day name)

"monthFormat": "M" to "MMMM" (month number, short or long month name)

No impact on communication between phone and watch


jbr7rr:
It seems code lint is targeting the Dev branch of the source repo. @MilosKozak is this something you can fix?

Otherwise we have to keep our Devs up to date when making PR. Works for me though

MilosKozak:
> It seems sonar is targeting the Dev branch of the source repo. @MilosKozak is this something you can fix?
> 
> Otherwise we have to keep our Devs up to date when making PR. Works for me though

I don't know how

PR #2743, opened 2023-09-05
Wear Move latest remaining Display setting from Phone to Watch (prediction lines)

Philoul:
Now all display settings concerning watchfaces are on Watch side


Philoul:
@MilosKozak Do you know if we have an easy way to organize Watch settings ?
=> I tried to use `<PreferenceCategory` within `preference.xml` file to organize preferences within Watch but it crashed immediatly...

The idea was to have an intermediate layer with for example: (currently we have close to 30 items now within watch...)
- Display settings
- Graph settings
- User interface settings
- Complication settings

MilosKozak:
because wear app works only with main app i'd agree with moving settings to phone

jbr7rr:
+1 for moving all/most wear settings to phone app instead of in wear. It's fiddly to change settings on watch. Especially when there are 30+ items in the menu ;)

Philoul:
@jbr7rr see #2753 to start discussion on improvements...

Issue #2745, opened 2023-09-07
Assert.assertTrue(expression) tests do not explain failures


ryanhaining:
`assertTrue` with an expression inside often contains little information in a test failure. For example, this `DateUtilTest`

```
Assertions.assertTrue(DateUtil(context).dateString(1513902750000L).contains("22"))
```

Fails opaquely with

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at info.nightscout.androidaps.utils.DateUtilTest.dateStringTest(DateUtilTest.kt:43)
```

Rewritten to use a [truth](https://truth.dev) matcher, the output is much clearer
```
assertThat(DateUtil(context).dateString(1513902750000L)).contains("22")
```

fails with

```
value of           : dateString(...)
expected to contain: 22
but was            : 12/21/17
	at app//info.nightscout.androidaps.utils.DateUtilTest.dateStringTest(DateUtilTest.kt:44)
```

ryanhaining:
There's a lot of other tests that use `assertTrue` which should be using matchers. Can we keep this open for a bit? I am working my way through them and will tag to this issue.

PR #2746, opened 2023-09-07
Fixes timezone-dependent failing DateUtilTest

ryanhaining:
Issue #2745, #2747

### DIFF
--- a/core/main/src/test/java/info/nightscout/core/utils/DateUtilTest.kt
+++ b/core/main/src/test/java/info/nightscout/core/utils/DateUtilTest.kt
@@ -8,14 +8,36 @@ import info.nightscout.shared.utils.T
 import info.nightscout.sharedtests.TestBase
 import org.junit.jupiter.api.Assertions
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.AfterAll
 import org.mockito.Mock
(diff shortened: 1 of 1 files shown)


MilosKozak:
thanks
this should go directly to `TestBase` as it can happen anywhere
why assertThat?

ryanhaining:
>  this should go directly to TestBase as it can happen anywhere

I recommend against adding that setup logic to every test since it isn't relevant for most tests. The `DateUtilTest` is the only one that's failing for me without it.

>  why assertThat? Assertions.assertTrue is junit5 thus not depending on another lib

I show the test output  in #2745 with `assertTrue` (the failure I was originally getting) and `assertThat`. Using specific matchers like this is common, [the project already depends on truth](https://github.com/nightscout/AndroidAPS/blob/dev/core/main/test_dependencies.gradle#L9). I wasn't able to understand the problem with `assertTrue`

Related, I am interested in contributing more to this project, and would like to start with improving tests and test coverage  since I do so much of it in my real job.

ryanhaining:
My email address is in my profile, contact me if you'd like to discuss any of that more directly, I have the usual apps.

MilosKozak:
@ryanhaining awesome if you can contribute on tests. I'm working on it wery slowly. Please PM me on discord

ryanhaining:
Sent FR

PR #2761, opened 2023-09-10
Drops hamcrest depedency

ryanhaining:
Replaces hamcrest matchers with truth.


ryanhaining:
This isn't actually a codecoverage reduction afaict.

Issue #2763, opened 2023-09-11
NSClientV3 Full sync fails when encounters non-sgv entries


justmara:
[Here](https://github.com/nightscout/cgm-remote-monitor/issues/8090) is the discussion started in NS issue.
tl;dr
/entries/history api can return non-sgv records

This [line of code](https://github.com/nightscout/AndroidAPS/blob/dev/core/ns-sdk/src/main/java/info/nightscout/sdk/mapper/SvgMapper.kt#L17) results in null record when there is no sgv field in entry.

And [here](https://github.com/nightscout/AndroidAPS/blob/dev/core/ns-sdk/src/main/java/info/nightscout/sdk/NSAndroidClientImpl.kt#L125) all the nulls are filtered out 🙂

So when Full sync tries to iterate over entries it stops prematurely because it receives lower amount of records than it requested. But these records were filtered-out on NSClient side.

MilosKozak:
resolved in https://github.com/nightscout/AndroidAPS/commit/d379b769d3739b75bb02d07d300f41871cb18c6d

PR #2767, opened 2023-09-11
DynISF unified (wip)

justmara:
Idea:

Current Dynamic ISF plugin does not bring anything new to oref1 itself. It only changes the way ISF is calculated. Commonly used ISF modes:
1. static profile-based - simple to use, for average users
2. TDD-based - proof of concept, seems to be working fine.
3. BG based (like in AutoISF) - many people simply want AAPS to act stronger and adjust CSF with BG rise, because they noted that higher BG levels require more insulin for every BG point to deal with.

In details.

TDD-based Dynamic ISF in reality consists of two steps:
1. calculate `base` (normalTarget) ISF from TDD
2. calculate `current` ISF as shift from base one to current BG

Step two seems totally like `BG based ISF` itself.
So I've decided to incorporate all the ISF calculation in one place with following options available:

1. Use Dynamic ISF (on/off)
2. Use TDD for ISF (on/off)
3.    TDD ISF adjustment factor (% scale of dynisf)
4. Scaling velocity (how much BG level affects ISF)
5. Adjust sens (ISF change affects ratio too)
6. Follow the profile changes for ISF calculation (currently DynISF ignores profile changes and you don't have any power over it, being forced to wait till TDD stabilizes).
7. BG cap (level at which ISF scaling applies diminishing returns)

image

At the same time all the JS ISF calculation code is moved to single JS-string injected from `DetermineBasalAdapterJS` so it can be reused in any other oref1-based modification like EN/Boost/etc, so that mods no more require to copy-paste ISF calculations and can focus of their own logic.


koelewij:
I always was quite happy with the behavior to higher BG’s that DynISF has. I have used it in dev for about 7 months. I quit it because my TDD started fluctuating due to illness and after that I ended up with a roller coaster effect. Tho option to use profile ISF and having a stronger ISF at higher BG’s sounds like a very good option. I see the ISF calculated based on TDD as a nice reference, but constantly updating ISF using TDD is a bit too much. 
I vote positive for this change.

plugins/sync/src/main/java/info/nightscout/plugins/sync/nsclientV3/workers/LoadBgWorker.kt
```diff
@@ -64,7 +64,7 @@ class LoadBgWorker(
                         // Objective0
                         sp.putBoolean(info.nightscout.core.utils.R.string.key_objectives_bg_is_available_in_ns, true)
                         // Schedule processing of fetched data and continue of loading
-                        continueLoading = !(sgvs.size != NSClientV3Plugin.RECORDS_TO_LOAD || response.code == 304)
+                        continueLoading = !(sgvs.size <= NSClientV3Plugin.RECORDS_TO_LOAD*0.95 || response.code == 304)
```
justmara, line 67: sorry, commited this by an accident. its a crude temporal fix for [this](https://github.com/nightscout/AndroidAPS/issues/2763)

jbr7rr:
Do I read this PR correctly and have you moved the settings to the Sensitivity (autosens) settings? 

While I like the idea of moving it to a separate place so it can be used by other plugins in the future as well, not sure if from a UX perspective the sensitivity plugin is the right place to put these settings it can further complicate the understanding of the system for the user.

Also can you outline which dynISF functionalitiy is currently implemented and what modes are you adding ?

justmara:
Yes, I moved it to sensitivity section, right near autosens settings. Because currently `sensitivity` is taken by autosens in AAPS and it would add a lot of confusion if another `sensitivity` plugin is added. But surely its only UX thing and can be moved to any better place.

Current (dev-i state) DynISF is TDD+BG scaling with TDD adjustment factor and `adjust sens` (when recent tdd change affects `autosens.ratio`).

I've added:

1. Non-TDD (profile one) with BG level scaling mode.
2. BG cap (taken from Boost) to control BG/TDD scaling on higher BG levels
3. Velocity - to control whole impact of BG on ISF scaling (both for TDD and Non-TDD modes)
4. `Obey profile` - so profile % affects both TDD and Non-TDD calculated ISF

Extracted ISF calculation function from `determine-basal.js` so it can be altered from kotlin code.

faldor20:
This sounds really good to me and addresses many of the issues we have switching to dynisf. 
I think the current implementation makes a lot of assumptions about your situation and this brings the customisation needed to make it work much better for people outside the "norm", which is an admirable goal.

szantos:
+1 for it to get merged!

More and more demand (issues) rise because of various controversies regarding DynISF.

Every option can be useful, but let me point out some aspects:

There were more users in the (early) DynISF Discord community, who could not profit of the TDD dependency of DynISF. Problems like site leakage could also need to disconnect TDD from ISF.

Obey/follow profile change % introduces a new way for the user to affect the ISF, which is currently only manually through the DynISF adjustment factor possible. This opens the way towards automation, too.

justmara:
updated to reflect latest dev changes.
moved DyniSF settings back to Dynamic ISF plugin settings but still within its own section.
unified a bit openapssmsplugin/determinebasal inheritance to avoid copy-paste

Issue #2323, opened 2023-01-02
[3.1.0.3] Wear entered treatments are not received by my phone (no confirmation)


old-square-eyes:
Wear entered treatments are not received by my phone (no confirmation)

Keys should be the same for both mobile app and wear app, as was working fine for a month after an initial issue.

Reporting bugs
--------------
- **Note the precise time the problem occurred** and describe the circumstances and steps that caused
  the problem. 11:05am NZDT
- Logs Sent
- FB Post: https://www.facebook.com/groups/AndroidAPSUsers/posts/3438350919719634

vanelsberg:
Important information is missing on this issue:

* No logs included with this issue
* What AAPS release/version/build are you running (add a screen print from AAPS's about?
* What phone brand/type/OS
* What watch brand/type/OS
* Are Phone and Watch running latest updates
* How did you install the wear-full-release.apk?
* What watch face are you using?

Did you try uninstalling the wear app and select a stock watch face, then switch off the watch, then switch it on again and reinstall the AAPS wear app?

old-square-eyes:
@vanelsberg

- logs were shared via email from the app. I also referenced this Github issue number in the email. See "Logs sent" above. Sorry if I did it wrong.
- Version number is in the title.
- Phone S22 Ultra, Galaxy Watch 4 (both on latest updates)
- Wear app installed via "Wear Installer 2". Was working for treatments then unexpectedly stopped.
- Digital watch face and standard AAPS watch face have the same issue

Will try the reinstall steps suggested and report back.

maheini:
Does your watch have the latest built of AAPS? Have you tried to rebuild & generate the latest phone .apk

I'm using build 2022.12.15 on watch 4 & Huawei phone - everything is working - but just after I installed the latest build onto the watch.

old-square-eyes:
I'm using 3.1.0.3

vanelsberg:
Will testing dev I have been updating my Watch4 with new .apk versions multiply times without anu problems.
Are you absolutely sure you did not by accident install an new version .APK that was build with a different signing key?
This think this could maybe cause a duplicate installation, hence problems?

old-square-eyes:
100% sure on that. I built both the apks at the same time with the same key and stored them in my Google Drive where they are on thier own. Also it worked for a month. It's working again after a reinstall.

wootmasterslick:
Same problem here with Galaxy watch 5. Works for short time after reboot of watch then it stops

[...]

PR #2801, opened 2023-09-19
[WIP] Refactor

MilosKozak:


Philoul:
Do you think this refactoring could help to fix #2323 ?

I started to analyze this issue, but it seems that rxbus sent by tiles activities (or heartrate and probably complications too) are not received correctly by DataLayerListenerServiceWear when an external watchface is selected...
I don't know how I can help more on thîs issue...

PR #2857, opened 2023-10-03
Wear CWFNew customization capabilities with Dyn Datas

Philoul:
With this PR, we can now customize images, rotation, text color according to values:
=> See as example Steampunk watchface reproduced as a cutom watchface:
- Background image rotate according to BG value
- Avg Delta background change according to range (5mgdl, 10mgdl or 20mgdl, or 0.3, 0.5 and 1mmol) (background image of freetext 1
- Avg Delta Pointer rotation vary according to Avg Delta range (managed by freetext 2 freetext 3 and freetext4)
- Rig_Battery color switch to dark red below 20% (color variation according to Rig Battery level)
- I replaced % value for Uploader battery by 5 icons to illustrate

See original Steampunk Watchface, and the 2 corresponding Custom Watchfaces... (note json files has a lot of comments to explain how to manage customization, my idea was to illustrate with an example possibilities...)

### DIFF
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/rx/weardata/CustomWatchfaceFormat.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/rx/weardata/CustomWatchfaceFormat.kt
@@ -238,7 +238,20 @@ enum class JsonKeys(val key: String) {
     ALLCAPS("allCaps"),
     DAYNAMEFORMAT("dayNameFormat"),
     MONTHFORMAT("monthFormat"),
-    BACKGROUND("background")
+    BACKGROUND("background"),       // Background image for textView
+    LEFTOFFSET("leftOffset"),
+    TOPOFFSET("topOffset"),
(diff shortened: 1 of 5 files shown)


MilosKozak:
so the Steampunk can be replaced by custom json?

Philoul:
I think yes (instead of one single dedicated watchface that manage both mgdl, mmol, and some custom settings like auto range or one specific manual range), we have 2 custom watchfaces (one mgdl and one mmol) and auto range...
I think most of current watchfaces can be replaced... (only circle is very very specific).
But some custom settings (feature not available in cwf plugin) are easier to use on dedicated wf (when you select digital, and you can tune colors easily with preferences, to manage this we will have to create as many zip than we have available colors in digital wf...).
I think I will work on wf preferences (but it's a full wf preference refactoring...).

MilosKozak:
then remove the code at all and push watchface json
I'm fan of cleanig the code

Philoul:
Yes I will do
It was my next question I had in mind when I worked on steampunk custom but you already answered me 🤣...

Issue #2853, opened 2023-10-02
BUG / DEV-M / DASH History button restarts AAPS


robertrub:
When I look at the DASH history, AAPS restarts (I see the AAPS icon and then the main page with the progress bar etc). First found by @ohjay

![Screenshot_20231001_100512_AAPS.jpg](https://github.com/nightscout/AndroidAPS/assets/76626474/a6c748a6-afa6-4cda-9a78-36b44363cc61)

![AndroidAPS_LOG_1696214180512.log.zip](https://github.com/nightscout/AndroidAPS/files/12779212/AndroidAPS_LOG_1696214180512.log.zip)

TheNeskik:
Same issue here
[AndroidAPS_LOG_1696248292700.log.zip](https://github.com/nightscout/AndroidAPS/files/12782465/AndroidAPS_LOG_1696248292700.log.zip)

TheNeskik:
Same here : log below
[AndroidAPS_LOG_1696250330973.log.zip](https://github.com/nightscout/AndroidAPS/files/12782791/AndroidAPS_LOG_1696250330973.log.zip)

robertrub:
To all that have commented and added logs.

Please try to use pod history again. It started working for me but I did nothing new... I didn't start a new pod. I just rebooted the phone but no idea if this is related. Now, pod history works.

The bug got solved all by itself... Not normal.

parapenT1sta:
> To all that have commented and added logs.
> 
> Please try to use pod history again. It started working for me but I did nothing new... I didn't start a new pod. I just rebooted the phone but no idea if this is related. Now, pod history works.
> 
> The bug got solved all by itself... Not normal.

A fix is waiting to be merged: https://github.com/nightscout/AndroidAPS/pull/2858

TheNeskik:
> To all that have commented and added logs.
> 
> Please try to use pod history again. It started working for me but I did nothing new... I didn't start a new pod. I just rebooted the phone but no idea if this is related. Now, pod history works.
> 
> The bug got solved all by itself... Not normal.

Not on my phone :)

Tried the fix and everything seems to be back in place.

ohjay73:
powercycled and still have the same crash experience.

jbr7rr:
> A fix is waiting to be merged: #2858
Fix needs testing. Have not tested it on a active pod. So if any of you can test the PR and see if everything is there in the pod history that would be really helpful

robertrub:
[...]

PR #2858, opened 2023-10-03
Dash: Fix crash in history

jbr7rr:
https://github.com/nightscout/AndroidAPS/issues/2853

Remove usage of Stream().toList() as this requires api 34. Will cause crash on some devices.

Testing:
See comments below and : https://github.com/nightscout/AndroidAPS/issues/2853#issuecomment-1746647261


TheNeskik:
I'm testing it on active pod.

As far as I tested it, History is back, every events are here :
- Basal
- Alert/Alam
- Bolus
- Event (changing pod) are listed (I didn't change my pod, But i can see the event for the current pod and the old one)

Didn't try to change the pod yet, All the test were made with the same pod which was already activated

jbr7rr:
@MilosKozak Ready for merge imo

PR #2872, opened 2023-10-05
WIP Refactor

MilosKozak:


Philoul:
- do you planned to also include triple up and down management within AAPS ? (included in enum but never used in UI)
- it would be fine if in wear we receive TrendArrow enum object instead of "arrow string"...

MilosKozak:
> * do you planned to also include triple up and down management within AAPS ? (included in enum but never used in UI)
> * it would be fine if in wear we receive TrendArrow enum object instead of "arrow string"...

I don't know if it's even used anywhere

MilosKozak:
> * do you planned to also include triple up and down management within AAPS ? (included in enum but never used in UI)
> * it would be fine if in wear we receive TrendArrow enum object instead of "arrow string"...

TrendArrow yes, but after I finish this

MilosKozak:
The goal is to split database and code completely

PR #2876, opened 2023-10-06
Wear CWF Include Cockpit Watchface into assets and simplify code

Philoul:
Reduction of code Complexity
Include TwinView Feature (this feature is less powerfull than full LinearLayout management when several views are hidden but it's a little improvement in a full FrameLayout)

Include a proposal of replacement for Cockpit Watchface.
The limitations compare to original Cockpit Watchface are:
- No LowBit mode management (with dark clouds)
- No tuning of Font size according to settings (for ex. a bit smaller if detailed IOB selected)
- No translated letter for H and L in the 2 lights
- Management of "Hidden View" a bit better in Cockpit WF (if TrendArrows are hidden, Delta values are vertical centered, it's not the cas in Custom Cockpit)

But also some little improvement
- centering of BG Value is better (and font size a bit bigger)
- I splitted lights (only upper light in yellow in Hyper, only bottum light in Red in Hypo)

If Ok with this proposal I will also remove Cockpit Watchface

Cockpit Watchface:
Cockpit Default

Cockpit Custom with zip file (also included in Assets):
CustomWatchface
[Cockpit.zip](https://github.com/nightscout/AndroidAPS/files/12828100/Cockpit.zip)


MilosKozak:
> If Ok with this proposal I will also remove Cockpit Watchface
👍

PR #2909, opened 2023-10-12
Local http server communication to integrate Garmin devices.

buessow:
This sends glucose values to the watch and receives heart rate.


buessow:
just realized that I added this commit to prev PR as well ....

PR #2922, opened 2023-10-14
Enable androidTest in CircleCI bulid

buessow:
For now only for database:impl package. The other androidTests needs some fixes that I'll provide later. This makes the entire run much slower (~ 37min). Seems to be mostly because of compilation. I'm not sure why it's so much slower and so slow in general.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,21 +17,22 @@ jobs:
       tag: 2023.07.1
 
     steps:
-      # Checkout the code as the first step.
       - checkout
+
       - android/change-java-version:
           java-version: 17
-      # The next step will run the unit tests
+
+      - android/start-emulator-and-run-tests:
+          system-image: system-images;android-29;google_apis;x86
+	  # Compilee while the emulator starts to use the time.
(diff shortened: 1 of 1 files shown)


MilosKozak:
It seems coverage is not added

MilosKozak:
is it possible to run more executors and do unit tests and android test in parallel?

MilosKozak:
taking back executors. it uses all cpus anyway

buessow:
I believe running normal unit tests and packaging for android tests takes
long. I believe it would be possible to run this in parallel with CircleCI.
That would save wall time but not cpu cycles/cost. I don't understand why
it's so slow on circleCI, the entire build + test takes only 4.5 minutes on
my laptop.

Robert Buessow . Mobile: +41 78 899 8296

MilosKozak:
there are only 4CPUs

buessow:
You can run the pipeline on multiple machines but I believe that would cost
more CPU cycles.

Robert Buessow . Mobile: +41 78 899 8296

MilosKozak:
are you somewhere on chat?

PR #2931, opened 2023-10-17
UI: Increased touch target size of action buttons

kenzo44:
Improved look and accessibility of action buttons

Screenshot 2023-10-16 at 10 55 02 PM


MilosKozak:
I reverted this change. New layout takes too much space and text which fit before now leads to word wrap

PR #2941, opened 2023-10-18
add 5 missing translations to wear

olorinmaia:
The most noticeable is the one when you try to use calculator on wear, and zero insulin is required.


olorinmaia:
Built APK and testet locally, strings appear in right place.

plugins/sync/src/main/kotlin/app/aaps/plugins/sync/wear/wearintegration/DataHandlerMobile.kt
```diff
@@ -175,7 +175,7 @@ class DataHandlerMobile @Inject constructor(
                                EventMobileToWear(
                                    EventData.ConfirmAction(
                                        rh.gs(R.string.loop_status).uppercase(),
-                                       "TARGETS:\n$targetsStatus\n\n$loopStatus\n\nOAPS RESULT:\n$oAPSResultStatus",
+                                       "$targetsStatus\n\n$loopStatus\n\n$oAPSResultStatus",
```
MilosKozak, line 178: simply removed text?

PR #2986, opened 2023-10-26
Insight -> kt migration

MilosKozak:


Philoul:
I tried to build Insight branch but it failed (even with invalidate caches and restart)
image

Should I merge within dev33 to be able to build and test ?

PR #3005, opened 2023-10-29
Fix typo of DynISF setting "Adjust sensitivity and BG" and add summary

olorinmaia:
Tim has approved text in summary, taken from docs and fix'd typo.

image


plugins/aps/src/main/res/values/strings.xml
```diff
@@ -23,7 +23,8 @@
     <string name="key_DynISFAdjust" translatable="false">DynISFAdjust</string>
     <string name="key_adjust_sensitivity" translatable="false">dynisf_adjust_sensitivity</string>
 
-    <string name="dynisf_adjust_sensitivity">Adjust sensitivity and BG</string>
+    <string name="dynisf_adjust_sensitivity">Adjust sensitivity and basal</string>
```
MilosKozak, line 26: This is not true. Event activating DynISF deactivates autosens

PR #3008, opened 2023-10-30
Garmin: Fix glucose unites and widget support

buessow:
Get glucose unit from settings rather than profile.
Add post carbs and connect/disconnect functionality for Garmin Widget.


MilosKozak:
and the watchface is? :)

PR #3016, opened 2023-11-01
Add missing permission to the notificationManager

AllWorkNoPlay-95:
I found a missing permission needed in the wear/src/main/kotlin/app/aaps/wear/comm/DataHandlerWear/handleBolusProgress function on line 250.

I added it.


MilosKozak:
* all changes must target dev branch
* this permission targets API level 33+ (we have 28)
* adding this is not enough anyway. app still must ask or permission

PR #3017, opened 2023-11-01
Fix crash when bolus by using DanaR Korean

diyaps:
Compare to AAPS 3.1, it should be require(detailedbolusinfo.carbs <=0)


MilosKozak:
actually it must be == 0
replaced by https://github.com/nightscout/AndroidAPS/commit/113f51126972a132ef505fb4d175ecf8210de080 and fixed everywhere
thanks for letting me know

PR #3034, opened 2023-11-05
Added NSCv3 handling for when remote NS durations are not stored as Long

E-Allie:
If a Nightscout instance stores Treatment Durations as non-Longs, such as Doubles, an error will be yielded as soon as the duration is parsed/serialized into the RemoteTreatment duration, a Long.

This changes the RT duration to a Number, and adds a function within the class to handle the conversion of it to a Long for further use.

".toLong" or "as Long" is not alone sufficient, as there are edge cases to be considered with decimal values <1.

### DIFF
--- a/core/nssdk/src/main/kotlin/app/aaps/core/nssdk/mapper/TreatmentMapper.kt
+++ b/core/nssdk/src/main/kotlin/app/aaps/core/nssdk/mapper/TreatmentMapper.kt
@@ -32,6 +32,7 @@ fun String.toNSTreatment(): NSTreatment? =
 
 internal fun RemoteTreatment.toTreatment(): NSTreatment? {
     val treatmentTimestamp = timestamp()
+    val duration = durationAsLong()
     when {
         insulin != null && insulin > 0                                     ->
             return NSBolus(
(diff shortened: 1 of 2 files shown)


E-Allie:
> How can happen that durations are not Long?

In my case, I am syncing a Tandem tslim x2 pump with Nightscout. The temp basals that ControlIQ [the pump's built in loop functionality] initiates stores durations as Doubles, not as integers/longs, and thus my NS instance also stores doubles

Since NS only expects Numbers/is type tolerant, this works fine on the NS end, but not when syncing to AAPS

E-Allie:
While I understand your concerns, this is a full blocker for syncing NS durations that aren't longs

An example response from my NS database is:

`{"status":200,"result":[{"eventType":"Temp Basal","reason":"profileDelivery","duration":242.18333333333334,"absolute":1.15,"rate":1.15,"created_at":"2023-07-30T04:00:00.000Z","enteredBy":"Pump`

where duration is in fact in minutes, but not an int/long

This would cause the error:

`com.google.gson.JsonSyntaxException: java.lang.NumberFormatException: Expected a long but was 242.18333333333334 at line 1 column 107 path $.result[0].duration`

on existing release/dev branch

The code provided wouldn't lead to inconsistencies between duration in minutes or milliseconds either, since the duration to long rounding is done before any further functions are called, so the internal minute and millisecond durations are equal.

Do you have any suggestions on how to alternatively handle this discrepancy?

MilosKozak:
put `duration=242`, `durationInMilliseconds=14531000` and AAPS will accept it

E-Allie:
That does make sense for one off values,

When values are being automatically synced from this pump/there are thousands of durations in NS as double

MilosKozak:
What SW does the synchronization?
And why are this data needed in AAPS?

PR #3036, opened 2023-11-06
Change for xdrip

rosscortb:


MilosKozak:
@rosscortb thanks for this PR but it will not be accepted

PR #3053, opened 2023-11-10
Add missing dash notification string when there is no active pod

olorinmaia:
When there is no active pod we now in master/dev get a non-translated notification. This adds a string that's already translated to properly notify in local language setting.

image


olorinmaia:
I think this is what shows in AAPSClient in Pump status overview, add strings for that also or leave be?
image

PR #3069, opened 2023-11-18
Only Upload TBR that are not in progress to Tidepool

emmatovar27:
This would avoid the overlapping of Temp Basal Rates


MilosKozak:
* there are typos in code
* I don't see `inProgress` field in documentation. Where is it from?

PR #3092, opened 2023-12-01
Add support for SensorStartedAt extra in xDrip+ BG source

Jackenmen:
This PR updates the xDrip+ BG source to auto-add sensor change time from the bundle (see addition from https://github.com/NightscoutFoundation/xDrip/pull/3218), similarly to how Dexcom BG source does it.

Works as expected when I tested it with xDrip+ using the "Libre (patched App)" source with Libre2 sensors that I'm on.


Jackenmen:
I added a setting for this to the xDrip+ BG source plugin though have not tested that functionality yet.

I see that the coverage report check is failing though it doesn't seem like the existing code was covered either so I'm not sure whether there are tests I need to add.

Jackenmen:
The PR I made to xDrip+ has been accepted and merged and I have now added unit tests for the part that I added to the xDrip+ BG plugin so this is ready for review now :)

MilosKozak:
do not create new prefs file as content is the same
👍 for tests

Jackenmen:
> do not create new prefs file as content is the same

That would mean that I would have to use the Dexcom's prefs file/settings, is that what you want? I pushed that but I found it a bit weird to refer to Dexcom-specific stuff in another plugin.

parapenT1sta:
@Jackenmen Is it possible to make calibrations be shown in AAPS as Dexcom app?

Jackenmen:
I don't know, I neither use calibrations in xDrip+ for my Libre 2 sensors, nor did I ever use Dexcom sensors or their app. If you want this kind of feature, just make a feature request in the issue tracker, maybe someone else will be interested in it.

Der-Schubi:
@MilosKozak @Jackenmen 
Switched to new xDrip and AAPS builds today, both now with this change.
My sensor statuslight is shown red with sensor age of 7 days, but critical threshold is set to 240h / 10 days.
I hope this will work after the next sensor change?!

Jackenmen:
I'm sorry but you haven't given enough information for me to know. This change should work as long as xDrip+ is actually aware of sensor change time (shows expiration date at the bottom when you're on its main screen with the graph) *and* you **enabled it** in xDrip+ BG plugin settings. If those conditions are met, there should afaik be no need to wait for the next sensor change either since AAPS will record the sensor change treatment if there isn't one already for the time given by xDrip+.

If you think there's a bug in this feature, I'd suggest creating a new issue and attaching to it the AndroidAPS logs which will, among other things, contain info about what xDrip+ is sending in the broadcast. Feel free to mention me on it in that case.

> My sensor statuslight is shown red with sensor age of 7 days, but critical threshold is set to 240h / 10 days.

This makes me wonder if your problem is even related to this feature or just the sensor status light settings which have not been touched by this PR? Again, you didn't include enough information for me to know what your issue is.

Der-Schubi:
> This makes me wonder if your problem is even related to this feature or just the sensor status light settings which have not been touched by this PR? Again, you didn't include enough information for me to know what your issue is.

Thanks for the reply. Yes, xDrip is aware of the change time and yes, the option is enabled.
I just wanted to find out if the problem possibly could be caused by this change. But I think its not related.
Will see tomorrow after the next sensor change...

PR #3093, opened 2023-12-01
Auto-add "BG Check" treatment for Contour Link readings

Jackenmen:
Adds Contour Link readings received by Medtronic pump as BG check treatments with FINGER as meter type.

**WARNING: I do not have a mmol/l Contour Link meter to test this with, I was only able to validate that it works with mg/dl.** The PR is written to work with mmol/l profiles and the data in pump history *should* be in mg/dl regardless of the meter used but I just want to make it clear that I haven't *actually* tested such a setup.

---

To implement this, I had to add `insertFingerBgIfNewWithTimestamp` to the `PumpSync` interface. The implementation is similar to the existing `insertTherapyEventIfNewWithTimestamp` but accepts additional data about the glucose value and its unit.

---

Originally, I thought I could just use the already implemented `decodeBgReceived()` but it turns out that it was unused for a reason - I found two main problems with it:
- byte indexes used for the glucose were off-by-one (`.rawData[0]` and `.rawData[3]` instead of `.rawData[1]` and `.rawData[4]`) - I decided to replace the usage of them with equivalent `.head[0]` and `.datetime[2]`
- meter variable contained a three-element byte array instead of something useful, such as a string representation of the meter's serial number

While rewriting the decode logic, I figured I'd improve the names to `GlucoseMgdl` and `MeterSerial` *and* set a `displayableValue` as well.

---

Processing BGs is mostly identical to processing primes, rewinds, and battery changes - it uses a slightly different `PumpSync` method and converts mg/dl glucose to the profile's unit before passing it to the treatment. The conversion needs to use the profile's unit as the pump event always stores this data in mg/dl units, no matter what unit the glucometer (or pump's bolus calculator) uses.

Note that the pump's bolus calculator does have a unit setting that determines how the pump's UI shows it but it's irrelevant to the unit that the meter shows it in so I figured it's best to just use the profile's unit here (especially since using pump's unit would require implementing a command for getting that info from bolus calculator settings since iirc it's not part of the output of regular settings command; that would also mean that a full refresh would need to run *another* command as well and generally just complicate this simple change a lot).

### DIFF
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/pump/PumpSync.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/pump/PumpSync.kt
@@ -1,5 +1,6 @@
 package app.aaps.core.interfaces.pump
 
+import app.aaps.core.interfaces.db.GlucoseUnit
 import app.aaps.core.interfaces.profile.Profile
 import app.aaps.core.interfaces.pump.defs.PumpType
 import app.aaps.core.interfaces.utils.DateUtil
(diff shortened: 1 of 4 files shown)


Jackenmen:
I'm not sure I have to add any tests here:
- `PumpSyncImplementation` has 0% coverage so I assumed I don't need to cover that
- `MedtronicHistoryData` has only `createTBRProcessList` and `getFilteredItems` methods covered which also suggests that I don't need to cover it
- `PumpSync` is partially covered, including the (similar to `insertFingerBgIfNewWithTimestamp`) `insertTherapyEventIfNewWithTimestamp` method but I'm not sure what that means exactly since it's just an interface

Let me know if there is a need for some tests though (and what for) and I will try figuring something out.

MilosKozak:
👍

PR #3096, opened 2023-12-03
Dev33 glucose status

ga-zelle:
This is the first element of integrating autoISF into mainstream AAPS. It extends the GlucoseStatus by more information, mainly bg acceleration as a very important input to autoISF. The remaining integration actions will mostly be contained in its own, new plugin alongside DynamicISF and the others.

I want to thank @T-o-b-i-a-s for his patience in introducing me into the miracles of git. A big thank you goes to @Philoul for his support and guidance in how to structure and organise this integration effort.

### DIFF
--- a/core/data/src/main/java/app/aaps/core/data/iob/GlucoseStatus.kt
+++ b/core/data/src/main/java/app/aaps/core/data/iob/GlucoseStatus.kt
@@ -1,5 +1,9 @@
 package app.aaps.core.data.iob
 
+/**
+ * added information about recent glucose pattern properties
+ */
+
 data class GlucoseStatus(
     val glucose: Double,
     val noise: Double = 0.0,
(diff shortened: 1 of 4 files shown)


core/data/src/main/java/app/aaps/core/data/iob/GlucoseStatus.kt
```diff
@@ -6,5 +6,15 @@ data class GlucoseStatus(
     val delta: Double = 0.0,
     val shortAvgDelta: Double = 0.0,
     val longAvgDelta: Double = 0.0,
-    val date: Long = 0L
+    val date: Long = 0L,
```
MilosKozak, line 9: please add javadoc here to explain meaning of new values

ga-zelle: Sure; what is javadoc?

PR #3103, opened 2023-12-10
Fix for a misleading string (Garmin -> Samsung)

swissalpine:
This explanation has puzzled me for years, because the broadcaster has nothing to do with Garmin. Rather, it refers to this Samsung related app:
https://github.com/trupici/G-Watch-Wear


MilosKozak:
ahhh right. But I'd add too it's TizenOS related as Samsung has WearOS watches now

swissalpine:
No problem.

PR #3123, opened 2023-12-18
Equil: Fix bugs

EquilHack:
Fix bugs
1.Add auto close Bluetooth connection
2.Fix pair, unbind pair code first exception bug


MilosKozak:
@EquilHack it would be better to utilize `connect` `disconnect` interface functions which are called automatically from `CommandQueue` as needed

MilosKozak:
@EquilHack do not call `disconnect` explicitly. It's call automatically when command queue for the pump is empty

PR #3150, opened 2023-12-30
Garmin: Add basal IOB to IOB and support low/high glucose mark to show range on the watch.

buessow:


plugins/sync/src/main/kotlin/app/aaps/plugins/sync/garmin/LoopHubImpl.kt
```diff
@@ -90,6 +91,12 @@ class LoopHubImpl @Inject constructor(
             return if (apsResult == null) Double.NaN else apsResult.percent / 100.0
         }
 
+    private fun toMgDl(glucose: Double) =
+        if (glucoseUnit == GlucoseUnit.MGDL) glucose else glucose * 18.0
+
+    override val lowGlucoseMark get() = toMgDl(preferences.get(UnitDoubleKey.OverviewLowMark))
+    override val highGlucoseMark get() = toMgDl(preferences.get(UnitDoubleKey.OverviewHighMark))
```
MilosKozak, line 98: value can be in both units: MGDL and MMOL. Use `ProfileFunctions`

buessow: I guess you ProfileUtil. Changed that and removed the local toMgDl function. I wasn't so sure if another dependencies is worth the reuse, since now the test mocks out the functionality so you only learn in in production if I use the function correctly.

Issue #3148, opened 2023-12-30
Can't get past objective 1 in fresh AAPS install with NSClientV3


olorinmaia:
AAPS 3.2.0.3
Build: bb1a276078-2023.12.17
Nightscout version: 15.0.2

I've seen some new users having issues with this, so I have recreated the problem. Steps:
1) Do a fresh install of AAPS
2) configure NSclientv3
3) enable all uploads etc to NS

Alltho BG is uploaded to NS, the objective regarding this doesnt get completed.

Workaround is to select NSClientV1, then objective gets completed.

Logs:
[AndroidAPS_LOG_1703899762554.log.zip](https://github.com/nightscout/AndroidAPS/files/13796768/AndroidAPS_LOG_1703899762554.log.zip)

Screens:
![image](https://github.com/nightscout/AndroidAPS/assets/114103483/c7908fbe-96a8-43a3-befe-69b6fe9848ac)
![image](https://github.com/nightscout/AndroidAPS/assets/114103483/46aa9973-48f4-440b-8dd5-67db44f7c436)
![image](https://github.com/nightscout/AndroidAPS/assets/114103483/57d5a195-7c0d-45ae-8ffc-86318faef6a7)

emna2305:
i got the same problem how i can fix it

olorinmaia:
Switch to NSClientv1 and see that objective pass then switch back

olorinmaia:
Fixed by https://github.com/nightscout/AndroidAPS/commit/172075ebf80a929004dfd8be5d8863beda42712b

PR #3165, opened 2024-01-09
Bugfix objective 1 not completing "BG available in NS or Tidepool" when NSClientV3 is used

olorinmaia:
Tested locally and it solves the issue below. Let me know if it's more complicated than so. Could this also be merged to dev32 for 3.2.0.4, since I've seen quite a few new users having problems with this.

https://github.com/nightscout/AndroidAPS/issues/3148


MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/172075ebf80a929004dfd8be5d8863beda42712b

PR #3176, opened 2024-01-19
Fixed boken regexp in allCommitted()

vaterlangen:
The Regexp used to allow changes in various files was broken as regexp syntax has to be casted explicitly when using string.replace() function. 

Steps to reproduce:
* Change file within `.idea/codeStyles/`
* Try to build
* FAIL due to uncommited changes


Philoul:
Target branch should never be master but dev branch when you make a PR...

vaterlangen:
Sorry.
Updated branch and base to dev

PR #3177, opened 2024-01-21
Medtrum improvements

jbr7rr:
- Remove "patch fault 2" string to avoid confusion
- Show Medtrum ref no instead of raw int in overview and error message
- Fix stuck on error when activation failed due to connection issue (disconnect)

Changes have been tested by myself

Let me know if correct target branch.  Otherwise I can rebase

### DIFF
--- a/pump/medtrum/src/main/java/info/nightscout/pump/medtrum/util/MedtrumSnUtil.kt
+++ b/pump/medtrum/src/main/java/info/nightscout/pump/medtrum/util/MedtrumSnUtil.kt
@@ -1,33 +1,20 @@
 package info.nightscout.pump.medtrum.util
 
+import info.nightscout.pump.medtrum.comm.enums.ModelType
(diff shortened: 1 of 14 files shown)


pump/medtrum/src/main/java/info/nightscout/pump/medtrum/util/MedtrumSnUtil.kt
```diff
@@ -1,33 +1,20 @@
...
-        const val MD_5201 = 81
-        const val MD_0202 = 82
-        const val MD_5202 = 83
-        const val MD_8201 = 88
-        const val MD_8301 = 98
-    }
-
-    fun getDeviceTypeFromSerial(serial: Long): Int {
-        if (serial in 106000000..106999999) {
-            return INVALID
```
jbr7rr, line 20: This still needs to be in

MilosKozak:
if you want to have it in prodaction sooner target dev32

PR #3206, opened 2024-01-31
3-in-1 determine basal.kt including AutoISF

ga-zelle:
This PR is focussed on the kt variant of DetermineBasal and a replacement of the earlier js-focussed PR #3161.

It proposes to unify AutoISF also into the dev version of SMB+DynISF unified to avoid extensive code duplication for a separate DetermineBasal containing AutoISF. The principle of inclusion was based on DynISF and AutoISF being mutually exclusive in their influence because I cannot see an option to combine any of their elements.

The existing kt versus js verification tests were re-run successfully for AMA, SMB and DynISF.

Again, I want to say thank you to @Philoul  for his continued interest and consultancy


MilosKozak:
The amount of changes to determinebasal and profile is huge and makes the code unreadable. Even integration with DynISF is wrong. I'd create new plugin with new OAPSProfile

ga-zelle:
superseeded by #3327

Issue #3210, opened 2024-02-02
Easy to miss toggles on the calculator


Mendanbar:
![Screenshot_20240202-083736](https://github.com/nightscout/AndroidAPS/assets/10580609/12d6d36c-4db5-4aa4-af23-8fc448c8deae)

Can we brainstorm some way to make the buttons a little more obvious in the hotspot shown?  Due to the proximity to the OK and Cancel buttons, the toggles are very easy to accidentally press, and when they are toggled only a dimming is applied on/off.  This makes it really hard for young or new users to the system to see, and it can often result in wildly incorrect bolus calculations.

Some ideas:
- Borders around the buttons (toggles and/or OK and Cancel)
- A more obvious on/off visualization
- Placement of the toggles moved to be harder to hit accidentally
- Confirmation dialog or some other mechanism to warn that the toggles have been pressed
- A lock icon that requires pin entry to allow changes to toggles or other advanced features

Mendanbar:
An optional lock would be my preference I think. It could be something enabled in the settings and opt in, so that power users don't have new and cumbersome UI mechanics foisted upon them.

kenzo44:
1) This is one with more padding between the buttons and the cancel button
<img width="321" alt="more_padding" src="https://github.com/nightscout/AndroidAPS/assets/4112129/155292fa-7a0a-4f19-b755-98d99d133eb2">

2) This is with padding and the icons are now placed inside buttons
<img width="321" alt="button_padding" src="https://github.com/nightscout/AndroidAPS/assets/4112129/7790c0ea-b5c9-48a8-af13-5c512e867d5b">

Thoughts

Mendanbar:
@kenzo44 I really like the second one.  I wonder if it's possible to have the "toggled off" state appear without a button surround, and the "toggled on" appear as a raised button?  Visually that would make it very clear which are toggled on/off at a glance.

I think the extra padding will help to reduce accidental toggles.

I also still think that the safest approach would be to implement an optional lock to prevent any toggles without a pin entry or other "unlock" mechanism, but understand that it would represent quite a bit more work.

kenzo44:
I am on board with increasing the padding for now. 

We will need to discuss the button states - example, true state would mean pressed in, and false would be raised like a tape recorder.

I don't know about the optional lock mechanism you described. It might complicate the UX. At the moment, we have a 2-step confirmation before delivering insulin.

kenzo44:
[...]

PR #3216, opened 2024-02-07
UI: Improved button accessibility in bolus wizard

kenzo44:
Solution https://github.com/nightscout/AndroidAPS/issues/3210

- Enabled/disabled state is obvious
- Added spacing between buttons and cancel button to prevent accidental taps


MilosKozak:
I don't like this predefined material design. In anyone able to create custom toggle button?

kenzo44:
@MilosKozak I can look into it. Initially, I was using custom buttons but switched to the default material look since it was faster.

What kind of design do you have in mind?

MilosKozak:
This looks like a tab we use elsewhere
I could imagine border around and change backgroung on toggle or something similiar

kenzo44:
@MilosKozak please review. I have updated the look of the buttons for both light and dark themes.

MilosKozak:
I like it now!

PR #3227, opened 2024-02-19
Update broken URL in objective 3

olorinmaia:
Fixes broken URL in obj3 reported in Discord:
image


Philoul:
Strings in other languages than english are managed within crowdin, not with PR.

So only main String file with english (../res/values/exam.xml) should be updated, not the others...

And within this link, only the filename is broken ("-in-xDrip" removed), not the anchor, so it's better to keep anchor, even if this one is close to the top:
https://wiki.aaps.app/en/latest/Usage/Smoothing-Blood-Glucose-Data.html#smoothing-blood-glucose-data

olorinmaia:
> Strings in other languages than english are managed within crowdin, not with PR.
> 
> So only main String file with english (../res/values/exam.xml) should be updated, not the others...
> 
> And within this link, only the filename is broken ("-in-xDrip" removed), not the anchor, so it's better to keep anchor, even if this one is close to the top: https://wiki.aaps.app/en/latest/Usage/Smoothing-Blood-Glucose-Data.html#smoothing-blood-glucose-data

Thanks, fixed. I thought so initially but thought maybe I could spare some crowdin updates. However they may get reset anyway when main res value gets updated? :)

PR #3233, opened 2024-02-22
AAPSClient xDrip+ statusline fix

olorinmaia:
Now when using AAPSClient string disabled_loop is showing 100% of the time in xDrip+ when using statusline.

This PR changes so it only sends string disabled_loop for AAPS. It's not really that useful for AAPSClient statusline imo.

If there is a way for AAPSClient to know that loop actually is disabled, and properly update xDrip+ statusline, please advise. :)


MilosKozak:
improved by https://github.com/nightscout/AndroidAPS/commit/acc1d17dd4457dd77e77829d14ba4a1868d433b9

PR #3241, opened 2024-02-25
Disable future_sens block when dynISF is not active

jbr7rr:
Having this block active when dynISF is not active add confusing text "Dosing sensitivity:" to the log and reason. This is not used when dynISF is not active. This PR disables this block when dynISF is not active


MilosKozak:
I didn't notice 👍

PR #3254, opened 2024-03-03
Allow for recordOnly bolus when pump is disconnected or suspended

jbr7rr:
This PR allows to record a bolus using the Insulin button when loop is disconnected or pump is suspended/not active

Have seen this request on discord some time ago, and experienced it myself.

Consider following scenarios:
- You are in a sauna with disconnected/suspended pump, you want to do pen bolus
- You canula is blocked or leaking, you do not trust AAPS to run, so you suspend pump and do manual pen bolus

There are other scenarios as well when you want to do a pen bolus while disconnected.

Tested:
- loop disconnected (medtrum)
- Medtrum Patch not activated
- Dash pod not active
- Insight not initialized

Update:
Screenshot_20240304-194743_AAPS

### DIFF
--- a/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
+++ b/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
@@ -583,8 +583,7 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
                 && preferences.get(BooleanKey.OverviewShowTreatmentButton)).toVisibility()
             binding.buttonsLayout.wizardButton.visibility = (!loop.isDisconnected && pump.isInitialized() && !pump.isSuspended() && profile != null
                 && preferences.get(BooleanKey.OverviewShowWizardButton)).toVisibility()
-            binding.buttonsLayout.insulinButton.visibility = (!loop.isDisconnected && pump.isInitialized() && !pump.isSuspended() && profile != null
-                && preferences.get(BooleanKey.OverviewShowInsulinButton)).toVisibility()
+            binding.buttonsLayout.insulinButton.visibility = (profile != null && preferences.get(BooleanKey.OverviewShowInsulinButton)).toVisibility()
 
             // **** Calibration & CGM buttons ****
             val xDripIsBgSource = xDripSource.isEnabled()
--- a/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
+++ b/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
@@ -594,7 +594,7 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
             } else {
                 setRibbon(
                     binding.buttonsLayout.insulinButton,
-                    app.aaps.core.ui.R.attr.ribbonTextDefaultColor,
+                    app.aaps.core.ui.R.attr.icBolusColor,
                     app.aaps.core.ui.R.attr.ribbonDefaultColor,
                     rh.gs(app.aaps.core.ui.R.string.overview_insulin_label)
                 )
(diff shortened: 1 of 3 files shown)


Der-Schubi:
Tested with ComboV2, works well. Thanks!
I really missed that, as you wrote: Sauna, disconnected pump and pen.

MilosKozak:
I generally have nothing against but it must visible on first sight that is record only. Maybe even on icon. Otherwise it can lead to mistakes.
Not sure how ... waiting for ideas

Der-Schubi:
> I generally have nothing against but it must visible on first sight that is record only. Maybe even on icon. Otherwise it can lead to mistakes. Not sure how ... waiting for ideas

For example the text / checkbox "do not bolus ..." could be red?!

jbr7rr:
> I generally have nothing against but it must visible on first sight that is record only. Maybe even on icon. Otherwise it can lead to mistakes.
Not sure how ... waiting for ideas

Currently upon entering the bolus, it is clearly stated in the confirmation screen.
I could make the checkbox/text "do not bolus, record only" red as Schubi suggested.

WhatsApp Image 2024-03-04 at 10 19 05

MilosKozak:
Based what I see on my daughter, nobody reads
I could imagine different color of icon + different color of window header for example

jbr7rr:
> Based what I see on my daughter, nobody reads I could imagine different color of icon + different color of window header for example

Something like this?
Screenshot_20240304-193343_AAPS
Screenshot_20240304-194743_AAPS
Screenshot_20240304-194827_AAPS

plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
```diff
@@ -583,8 +583,22 @@ class OverviewFragment : DaggerFragment(), View.OnClickListener, OnLongClickList
                 && preferences.get(BooleanKey.OverviewShowTreatmentButton)).toVisibility()
             binding.buttonsLayout.wizardButton.visibility = (!loop.isDisconnected && pump.isInitialized() && !pump.isSuspended() && profile != null
                 && preferences.get(BooleanKey.OverviewShowWizardButton)).toVisibility()
-            binding.buttonsLayout.insulinButton.visibility = (!loop.isDisconnected && pump.isInitialized() && !pump.isSuspended() && profile != null
-                && preferences.get(BooleanKey.OverviewShowInsulinButton)).toVisibility()
+            binding.buttonsLayout.insulinButton.visibility = (profile != null && preferences.get(BooleanKey.OverviewShowInsulinButton)).toVisibility()
```
MilosKozak, line 586: `.toVisibility()` should not be there

MilosKozak: I see it's on more places here

olorinmaia:
Thumbs up from me :) I actually thought about this myself one day as we disconnect pump when in the swimming pool!

Small "bug", when pump is connected (tested with virtual pump) text color of "Insulin" is white instead of blue like in dev/master.
image

plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
jbr7rr: Its needed here because:
Type mismatch.
Required:
Int
Found:
Boolean

jbr7rr:
> Small "bug", when pump is connected (tested with virtual pump) text color of "Insulin" is white instead of blue like in dev/master. 
Thx, fixed

vanelsberg:
This feature was high on my list:
Thanks!!!

mgit5987:
Disabling bolus when pump disconnected is ok.
Disabling bolus when Suspended is a wrong direction for me.
It i also inconsistent logically because basal is still deliveted and TBR set. 
In SMA bolusing and TBR when loop suspended was an very useful feature. We used that in 3.3.2.1 and before when we have rapid BG increase not from carbs so loop is not able to cover that by profile settings. That case we do bolus and maximize TBR for 30 minutes to stop rising BG. That is best quick possibility to react an I know that when loop will return it secure possible BG fall.  It is really usefull at night too.
Please allow bolusing when loop suspended or provide an option to choose behaviour in settings.

jimk9:
Hi, I also think bolusing should be allowed when loop is disabled or suspended. 
If my sensor is not giving stable/accurate BG readings (happens occasionally when new sensor is inserted) I will disable the loop and manually bolus using aaps since there is no issue with the pod. If I'm understanding the above correctly I will need to switch loop back on to either closed or lgs, bolus and then disable it. Appreciate I could use lgs but there is still the possibility of it giving a bolus based on bad BG values.
As @mgit5987 suggests a setting to allow choice of behavior would be good or only Disconnect pump stops the ability to bolus.
Thanks

PR #3273, opened 2024-03-19
Wear Plugin Add button to access to Exchange site for cwf in doc

Philoul:
Button "Set Default" within Wear plugin replaced by button "More watchfaces" to have direct access to exchange site within documentation.

image

image


plugins/sync/src/main/res/values/strings.xml
```diff
@@ -203,6 +203,7 @@
     <string name="wear_infos_watchface">Infos Watchface</string>
     <string name="wear_export_watchface">Export template</string>
     <string name="wear_new_custom_watchface_exported">Custom watchface template exported</string>
+    <string name="wear_link_to_more_cwf_doc">https://androidaps.readthedocs.io/en/latest/ExchangeSiteCustomWatchfaces/index.html</string>
```
jbr7rr, line 206: set translatable to false, so we don't have to copy this in crowdin :D

Philoul: I kept the link translatable to be able to replace the "/en/" by language code within crowdin and then access to the "translated" exchange page within doc.

PR #3306, opened 2024-04-19
Graph menu proposal

Philoul:
First commit to include a direct access to graph scale with long press
Second commit to keep menu opened during graph settings (to close menu just click outside menu

### DIFF
--- a/app/src/main/kotlin/app/aaps/activities/HistoryBrowseActivity.kt
+++ b/app/src/main/kotlin/app/aaps/activities/HistoryBrowseActivity.kt
@@ -4,10 +4,13 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Bundle
 import android.util.DisplayMetrics
+import android.view.Menu
+import android.view.View
 import android.view.ViewGroup
 import android.widget.LinearLayout
 import android.widget.RelativeLayout
 import android.widget.TextView
+import androidx.appcompat.widget.PopupMenu
 import app.aaps.core.data.time.T
 import app.aaps.core.interfaces.configuration.Config
 import app.aaps.core.interfaces.logging.AAPSLogger
@@ -126,6 +129,25 @@ class HistoryBrowseActivity : TranslatedDaggerAppCompatActivity() {
                     }
                 }
                 .show(supportFragmentManager, "history_date_picker")
+        }
+
+        binding.chartMenuButton.setOnLongClickListener { v: View ->
+            val popup = PopupMenu(v.context, v)
+            popup.menu.add(Menu.NONE, 6, Menu.NONE, rh.gq(app.aaps.core.ui.R.plurals.hours, 6, 6))
+            popup.menu.add(Menu.NONE, 12, Menu.NONE, rh.gq(app.aaps.core.ui.R.plurals.hours, 12, 12))
+            popup.menu.add(Menu.NONE, 18, Menu.NONE, rh.gq(app.aaps.core.ui.R.plurals.hours, 18, 18))
+            popup.menu.add(Menu.NONE, 24, Menu.NONE, rh.gq(app.aaps.core.ui.R.plurals.hours, 24, 24))
+            popup.setOnMenuItemClickListener {
+                // id == Range to display ...
+                rangeToDisplay = it.itemId
+                setTime(historyBrowserData.overviewData.fromTime)
+                loadAll("rangeChange")
+                return@setOnMenuItemClickListener true
+            }
+            binding.chartMenuButton.setImageResource(app.aaps.plugins.main.R.drawable.ic_arrow_drop_up_white_24dp)
+            popup.setOnDismissListener { binding.chartMenuButton.setImageResource(app.aaps.plugins.main.R.drawable.ic_arrow_drop_down_white_24dp) }
+            popup.show()
+            false
         }
 
         val dm = DisplayMetrics()
--- a/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
+++ b/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
@@ -16,12 +16,14 @@ import android.os.HandlerThread
 import android.util.DisplayMetrics
 import android.util.TypedValue
 import android.view.LayoutInflater
+import android.view.Menu
 import android.view.View
 import android.view.View.OnLongClickListener
 import android.view.ViewGroup
 import android.widget.LinearLayout
 import android.widget.RelativeLayout
 import android.widget.TextView
+import androidx.appcompat.widget.PopupMenu
 import androidx.core.text.toSpanned
 import androidx.recyclerview.widget.LinearLayoutManager
 import app.aaps.core.data.configuration.Constants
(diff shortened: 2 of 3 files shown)


olorinmaia:
Built and tested ok :) Improves the functionality compared as it is today.

Is it possible, let's say you have 4 graphs, that when you make a change on graph 4 that position in menu is locked to where you did the change, now it jumps to top of menu and you need to scroll down again.

Philoul:
I noticed that when you are tuning third or fourth subgraph, but unfortunatly the menu is rebuilt on each click. Here I use existing features and I just show again menu to have it opened again (I don't think it's possible to save and restore shift position, and if yes I don't know how to do that easily...)

Philoul:
PR converted to draft because several variants of rescale graph are included at the same time for tests
On latest commit I added Text to rescale on simple click graph (selected scale in bigger size):
image

Philoul:
@vanelsberg if we finally keep the4 textviews above graph to rescale, then we don't need anymore long click for rescale menu...

vanelsberg:
> @vanelsberg if we finally keep the4 textviews above graph to rescale, then we don't need anymore long click for rescale menu...

Yes, Was following the discussion on Discord on this. My commets:
Actually, this is about AAPS overview which should be limited to showing ongoing status. So important to keep away form "fancy views", limiting only to what is absolutely relevant?

olorinmaia:
> > @vanelsberg if we finally keep the4 textviews above graph to rescale, then we don't need anymore long click for rescale menu...
> 
> Yes, Was following the discussion on Discord on this. My comment: Actually, this is about AAPS overview which should be limited to showing ongoing status only? So imho important to keep away form "fancy views", limiting only to what is absolutely relevant...

I have to agree, the idea is good to simplify changing X hourly span of overview, but i think its sufficient to be able to change this by long pressing graph or the arrow button.

Philoul:
@olorinmaia @vanelsberg The idea was to reduce clicks for frequent actions and also to simplify graph menu very long.

So the 2 compromise can be compare to current solution (if no other ideas):
1) with TextViews (or Buttons)
++ direct single click to rescale graph
++ graph menu simplification with no submenu for rescale
++ no long click action for graphs
-- 4 additional items on graph area

2) Long Press menu only
++ long click for graph rescale menu (better than 3 click but less than one single click action)
-- most users will not use the long click feature on graph button (hidden so not intuitive) so will need 3 clicks (even the long click on graph not known by several users...)
-- no simplification of the graph menu that should keep the rescale submenu

Philoul:
@olorinmaia @vanelsberg The idea was to reduce clicks for frequent actions and also to simplify graph menu very long.

So the 2 compromise can be compare to current solution (if no other ideas):
1) with TextViews (or Buttons)
++ direct single click to rescale graph
++ graph menu simplification with no submenu for rescale
++ no long click action for graphs
-- 4 additional items on graph area

2) Long Press menu only
++ long click for graph rescale menu (better than 3 click but less than one single click action)
-- most users will not use the long click feature on graph button (hidden so not intuitive) so will need 3 clicks (even the long click on graph not known by several users...)
-- no simplification of the graph menu that should keep the rescale submenu

Philoul:
See below a proposal to simplify graph menu.
In this proposal I deleted rescale submenu (available with Textview on the top left of main graph)
image

vanelsberg:
> See below a proposal to simplify graph menu. 

@Philoul 
Much better, love this!

Request: could we merge this with [PR#3311](https://github.com/nightscout/AndroidAPS/pull/3311) (which has your "text menu" replace by those nice scaling button?
(PS - Tried merging PR#3311 with yours but that resulted in a merge conflict...)

Philoul:
> Request: could we merge this with PR#3311

This PR is only for testing, I asked @kenzo44 to open a dedicated PR to be able to test and compare both solutions for graph rescale. (that's why we have some little conflicts between both branches, but very easy to fix).

I think I will re-open a dedicated PR for final solution (I understood some users prefer my version with Textboxes overlapping main graph, and some other users prefer iceman's proposal with buttons above graph, both have ++ and --).

On final PR I will remove long press menu for rescale (I think now everyone prefer dedicated "buttons/textviews" solution, I got good comments on new menu for graph settings but would like to test it on low res screen to be sure it works fine (or tune it if necessary).

@MilosKozak any comments on this proposal (and variant with #3311)?

vanelsberg:
> .. I think I will re-open a dedicated PR for final solution

@Philoul Sounds great. Would love that! Will test your new PR a.s.a.p.

olorinmaia:
> I noticed that when you are tuning third or fourth subgraph, but unfortunatly the menu is rebuilt on each click. Here I use existing features and I just show again menu to have it opened again (I don't think it's possible to save and restore shift position, and if yes I don't know how to do that easily...)

I think you solved this very nicely @Philoul ! Love the new layout. Everything in one place without scrolling. I would also like to see how this looks with the buttons for 6, 12,18,24 hrs @kenzo44 have in his PR. Great job both of you as always =)

Philoul:
On latest commit pushed this evening I included several improvements, but I would like to have testers with very low res screen to test on physical device how this menu works...

As explain in a previous comment, this draft PR is for testing. I will probably close it to re do a new clean PR with final solution...

ga-zelle:
I prefer the text view alternative to save space. If it obscures anything just shortly change the scale and it won't be hidden any longer.

robertrub:
I prefer the small texts without backgrounds and on the graph. Much less invasive that buttons. It'd be better for small screens too.

robertrub:
Except the starts of the profile changes (that most people use rarely), there is no interesting information there.
Buttons would either use more height or cover more the graph :(

vanelsberg:
On using screen resources: graph scaling still is not optimal?
Most of the time I have 2 or 3 rows black graph space / empty rows at the top of the graph? It contains no graph data, only profile change icons?

vanelsberg:
**Suggestion**: as the graph scaling feature is also available by long-pressing the graph, how about making the buttons optional through overview config? This way user can choose?

ga-zelle:
Alternatively the text views could be placed below the time axis labels. There is space left on top of graph 1 and they reside close to what they impact.

robertrub:
> Edit: Also for me with buttons the UI looks a lot more cleaner and professional or may it is just that I hate using "text" areas as clickable item which UI-wise should be buttons? ;-)
> 
I agree with you but buttons use one more line and that is not great with small screens

olorinmaia:
I think both buttons and text is good solutions, there are + and - with both. It would be interesting to see how it looks if both of those are place d between main graph and graph 1.

Philoul:
3 variants of layout was tested:
On the left current version, in the middle, scale buttons moved below main graph (graph menu same place than now), on the left all buttons included above main graph.
To illustrate the overlay, I selected a scale (18h) with a lot of information...
image

vanelsberg:
Nice!

I'd go for the 3th option. There is enough "unused" space at the top of the graph.

Buttons do shield some of the profile indicators. Maybe (in another issue) move them slightly lower then?

If noth the 3th, then 2nd option (buttons below the graph) would also do for me.

robertrub:
Here is my vote ;)



It think by aligning the buttons to the right they will cover less information and will stay "near" the area where they were before.

vanelsberg:
How about optionally hiding the buttons through the pulldown/config menu? (me, running for cover ;-) )

Philoul:
> How about optionally hiding the buttons through the pulldown/config menu?
I during my several trials I put the 4 buttons vertically below menu icon. but it doesn't work... (and/or if you hide predictions, or if you select long range like 18h or 24h), you have too many recent information hidden by scale buttons...

Philoul:
> How about optionally hiding the buttons through the pulldown/config menu?

And if your idea is to include these buttons within dropdown menu (in a first row of menu), keep in mind that this menu is hidden in simple mode (so only remaining way to resize scale in simple mode is long press on main graph...

ga-zelle:
Leaning back and thinking about where we are now my first priority is to get rid io the long press changing time scale, second priority is not to waste srceen space. Anything else gets used very infrequently and therefore of minor importance with some pros and cons.

vanelsberg:
> I during my several trials I put the 4 buttons vertically below menu icon. but it doesn't work...

Actually I was thinking more about only "showing" the scaling buttons from your screenshots above when selected, in a similar way the graphs are selected. So when not selected, buttons are not visible. But in that case we would need to keep the long-press to scale the graphs. Hmmm...

Maybe just ignore this idea. It is adding another complexity to a solution that is good enough...

Philoul:
On last commit I replaced the 4 buttons by a popup rescale menu
image
 
Size and positionning of each graph /sub graph remains identical to current master
Overlay  for rescale menu is less important than with 4 dedicated buttons
Rescale is done with 2 clicks (instead of 3 in current master, or 1 in previous proposal with 4 dedicated buttons)
Seems to be a good compromise for me, don't hesitate to test and comment here if you agree !

vanelsberg:
New PR with rescale menu tested ok. Think 2 clicks only is better than current dev.
(+ Grown accustomed 😢 to the 1-click rescaling buttons @robertrub voted for - can we please get them back😉) 

@Philoul You showed many ways to do this. How about a poll?

Philoul:
Replaced by #3347

PR #3311, opened 2024-04-28
UI: Graph Scale Buttons instead of Dropdown

kenzo44:
An alternative solution to https://github.com/nightscout/AndroidAPS/pull/3306 - instead of having text overlayed on the graph, added buttons to switch between scales.

Users can tap on buttons to quickly switch between graph scales instead of doing it with 3 actions. This will also make the dropdown exclusively for graph data management.

Todo:
- Remove graph scale from the dropdown menu
- Remove long tap on graph to switch between graph scales


vanelsberg:
Nice...

Q: Looks like this PR was based on dev + PR#3306?
Merging PR#3311 on top of PR#3306 results in a merge conflict... 🤔

vanelsberg:
Sidenote: The buttons and profile change icons are overlapping. So I think layout needs some additional work on this...

vanelsberg:
Merge with current dev. Works without issues. Nice!

Philoul:
Final proposal (clean PR) done here #3347

PR #3327, opened 2024-05-06
AutoISF plugin prototype

ga-zelle:
According to the hint given in the earlier submission (#3206) this version has AutoISF completely contained in its own plugin.

It represents AutoSF version 3.0 as described in https://github.com/ga-zelle/autoISF/blob/A3.2.0.2_ai3.0/autoISF3.0_Quick_Guide.pdf. However, some functionality was not yet included, namely
- no dedicated automations
- no activity monitor based on step counting
- no home screen adaptations

AutoISF version 3.0.1 is in its final testing. Once verfied it will be added and offers a reduced set of preferences and flatter preference menus.

The code was tested
- by unit tests of its basic functions. The AutoISF function itself needs more work.
- test json's were added to ReplayApsResuiltsTest. One test failed due to minor deviation of eventualBG 103 vs. 104. Most likely this was due to unfortunate rounding in determine basal code (js and kt versions).
- observing test runs and verifying results against my AAPS emulator

A big thank you goes to @Philoul . Not only did he create the plugin structure but he also put in considerable work to align the architecture as much as possible with DynamicISF. This resulted in some code duplication but AutoISF is independent of the remaining code and can be tested and developed independantly.


core/interfaces/src/main/kotlin/app/aaps/core/interfaces/aps/OapsProfile.kt
```diff
@@ -47,5 +47,29 @@ data class OapsProfile(
     //DynISF only
     var variable_sens: Double,
     var insulinDivisor: Int,
-    var TDD: Double
+    var TDD: Double,
```
MilosKozak, line 50: Create `data class OapsProfileAutoIsf : OapsProfile`
this makes the profile value confusing for other algs

MilosKozak:
make it available in engineering mode only for now (see Autotune plugin)

core/interfaces/src/main/kotlin/app/aaps/core/interfaces/aps/OapsProfile.kt
Philoul: I tried, but difficulty was within APSResult interface and DetermineBasalResult class that also include OapsProfile for input information (so your request was not limited to AutoIsf plugin and associated unit tests, it also has side impacts on database with probably an additional field for profileJsonAutoIsf and OpenAPSFragment...)
=> I finally prefered to keep OapsProfile class for that reason (looks to me globally much more simple to keep OapsProfile class with default values for dedicated AutoISF values...)

image
image

Do you confirm this request ?

Philoul: I added the dedicated OapsProfileAutoIsf file (without modifying database).
😉

ga-zelle:
5 preferences and the related flatter menus from AutoISF 3.0.1 standalone are now added to this PR.

MilosKozak:
add to plugin
```
override fun specialEnableCondition(): Boolean = config.isEngineeringMode() && config.isDev()
```
and I'll merge it to simplify your life

ga-zelle:
Done and tested. Thanks and I am looking forward to an easier life ...

med06053:
Does this includes the new 1 minute libre capability?
And the dedicated automations?
This is a great addition.

ga-zelle:
In this phase NO to both your questions. Let's take it step by step.

Issue #3284, opened 2024-03-27
MEDTRUM Nano disconnecting


Korostensky:
After updating with https://github.com/nightscout/AndroidAPS/pull/3270, 
is the pump unreachable after an accidental disconnection (long phone distance) and no longer available.
--------------
Before the update, the phone (watch with full android) worked normally with the pump. 
I have been using it in this combination since 20.09.2023.
- 3.3.0-dev-b-app-full-release
- pump model - MD8201, FW 1.80.83
- phone model - KOSPET Optimus2, Android 10.7- non-standard, but fully functional until the last update.

To reconnect the pump, you must exit and restart the application. Then the pump will connect correctly. 

![image](https://github.com/nightscout/AndroidAPS/assets/54218454/97571057-c904-447d-be75-52606c4f158a)
[Log Export (1).txt](https://github.com/nightscout/AndroidAPS/files/14771939/Log.Export.1.txt)

jbr7rr:
Interesting device..
How are other Bluetooth devices working? Other pumps, dash etc?

Before I make any modifications to the code, can you double check the following:
- Location permission set to "always allow"
- ANY powersaving measures for app, Bluetooth app. and general ?

Also can you send me logs which also contains successful connection ?

Korostensky:
I've been using different types of watches, customized with Nova Launcher, since 2020, the problem still remains battery life. So far I have only used it in combination with the DANA R and MEDTRUM Nano and the connection works quite correctly. Sometimes it happens that the signal from the Dexcom drops out, but it comes back on after a few minutes. The devices connected via Bluetooth are only the Dexcom G6 transmitter and pump, nothing else.  

All settings are correct, I have enabled constant location access, Bluetooth enabled in the app. I don't have the battery saver activated. Before the modification it worked correctly and when I install the older version, the pump disconnection does not happen.

Log attached. 
[Log Export-succesful.txt](https://github.com/nightscout/AndroidAPS/files/14785475/Log.Export-succesful.txt)

jbr7rr:
It looks your device needs to scan to properly wake up or reset the Bleutooth when a connection attempt fails.
However other devices don't handle scanning on every connection error too well. (see: https://github.com/nightscout/AndroidAPS/issues/3268)

I see few options

1. Add a retry mechanism and after 3 retries we start scanning anyway
2. Add a setting, for those phones who need it can enable scanning.
[...]

PR #3331, opened 2024-05-10
Medtrum connection issues

jbr7rr:
https://github.com/nightscout/AndroidAPS/issues/3284

- Check better for permissions in driver and request when attaching new patch.
- Use deviceAddress instead of deviceObject to save the Bluetooth address for reconnects which should prevent inability to reconnect


jbr7rr:
Open for merge, has been tested for over 2 weeks on at least 2-3 phones, no regressions

PR #3337, opened 2024-05-14
Rearrange buttons in actions tab to save screen space 

olorinmaia:
When using pump Omnipod Dash (without battery/priming need) the action tab look like this now:
image

This PR will change it so that there will be less scrolling in action tab and look like this:
image

For pumps with battery/priming it will look like this.
Before:
image

After:
image

### DIFF
--- a/plugins/main/src/main/res/layout/actions_fragment.xml
+++ b/plugins/main/src/main/res/layout/actions_fragment.xml
@@ -204,21 +204,6 @@
                     app:layout_row="2" />
 
                 <app.aaps.core.ui.elements.SingleClickButton
-                    android:id="@+id/fill"
-                    style="@style/GrayButton"
-                    android:layout_width="0dp"
-                    android:layout_height="wrap_content"
-                    android:drawableTop="@drawable/ic_cp_pump_cannula"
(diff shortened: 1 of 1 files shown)


kenzo44:
Maybe move the sensor info section to the top so the profile switch buttons are reachable

olorinmaia:
> Maybe move the sensor info section to the top so the profile switch buttons are reachable

I'm not sure I follow, they are reachable or do I misunderstand what you mean? :) For my part I think it's nice that sensor info section stays where it is today to give some "space" between all the various buttons.

vanelsberg:
Hmmm.... Personally, I can live with it but I'd favor current UI layout over this PR.
Also agree with @MilosKozak remarking the info section was designed to be near the Careportal section.

(btw: does the solution actually match the PR's objective as suggested by its title? For saving screen space I's more expect changing button size?)

olorinmaia:
This PR simply rearranged the order of buttons to save screen space. It doesn't move the sensor/pump info section.

Without this pr you need to scroll to reach all buttons on my phone, so imo it does what title suggest 😀

PR #3346, opened 2024-05-23
UI: Moved actions below status section

kenzo44:
Adding to https://github.com/nightscout/AndroidAPS/pull/3337

Moved status to the top so buttons are more reachable

330615811-d0553aca-e4a6-415a-af45-ddeb571057d8


vanelsberg:
PR merges ok and is functional 👍

MilosKozak:
the reason why it was  below is that data belongs to Careportal actions

vanelsberg:
> the reason why it was below is that data belongs to Careportal actions

Agree. 
FWIW: I like UI as is better. Also, solution does not really match PR's objective.

PR #3347, opened 2024-05-24
New Graph menu + Scale menu

Philoul:
Final proposal for Graph and Scale menu (replace #3306 and #3311)
See below screenshots of UI proposal for Graph settings and new menu for Scale
image


vanelsberg:
Thanks, well done!
Merges build fine with current dev. First impression: all as expected!

MilosKozak:
is it still disabled to select the same graph multiple times? @Philoul

Philoul:
Yes,
I used checkboxes but they works close to radiobuttons on each row...
If you want to move a curve from one subgraph to a new one, you just click on the new corresponding column and previous one will be disabled.
(I selected a checkbox because you can disabled a curve with nothing checked on one row, it's not possible with radiobuttons)

PR #3354, opened 2024-05-30
Fix strings

MaiadeOlive:


MilosKozak:
* If you want to push changes you must ALWAYS target dev branch
* Strings must be changed only in EN language, the rest will be done in crowdin once changes are merged to dev

PR #3374, opened 2024-06-23
Change visibility of Boluswizard and Insulin button in AAPSClient

olorinmaia:
You can't bolus via pump from AAPSClient, thus it makes no sense to hide Insulin and Boluswizard button in AAPSClient when pump is disconnected from AAPS. This change will make visibility to these buttons to be similar to that of Carbs button.

We disconnect pump in AAPS when our son goes swimming, and it would be useful to check Bolus wizard for missing carbs for example in AAPSClient.

| AAPS (Pump connected) | AAPSClient (Pump connected) |
|--------|--------|
| image | image |
| AAPS (Pump disconnected) | AAPSClient (Pump disconnected) |
| image | image |


Philoul:
Not sure to understand your proposal (why you keep Insulin button in AAPSClient version 🤔).
Keep also in mind that you also have a PumpControl version of AAPS.

olorinmaia:
AAPSClient is for followers. It is not possible to adimister insulin via pump in any scenario. Thus there is no reason to hide the buttons.

When we disconnect pod it is useful some times to register insulin given by pen.

Philoul:
I can be wrong (never used AAPSClient myself), but I thought it was possible to send a bolus from AAPSClient to AAPS followed remotly...

olorinmaia:
You can add a bolus from AAPSClient that count towards iob, but will not be delivered by pump. This warning is always visible and can't be turned off.😉

Screenshot_20240721_220637_AAPSClient

MilosKozak:
I prefer notify parent about disconnected pump by having the same bahaviour

PR #3383, opened 2024-07-14
add simple stop SMB button functionality

### LINKED ISSUE #2572: Feature Request: Ability to stop SMB
The desire to be able to cancel running SMB's pops up often in Discord.

A few questions for discussion:

a) Why do we want this?
b) How do we want this?
c) Why don't we want this?
d) Other consolidations

My personal answers:

a) Mistakes happen, I once forgot to disable loop during exercise, a fairly large SMB was delivering and I could not cancel other then ripping off the pod
b) A solution where we can able to open the bolus progress screen (what we see when doing a normal bolus) where we can watch the progress and cancel the bolus
c) It adds complexity,
d) We need a very clear UI for this, I would prefer some good input on this before this gets implemented

lolokraus:
Added a Stop button when a SMB is administering like requested in the picture from the issue. 
The LinearLayout was changed to RelativeLayout so that the button doesn't move the text to the left.
When clicking the button the same logic gets executed as when stopping a normal bolus.

Fixes #2572

### DIFF
--- a/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
+++ b/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/overview/OverviewFragment.kt
@@ -51,6 +51,7 @@ import app.aaps.core.interfaces.profile.ProfileFunction
 import app.aaps.core.interfaces.profile.ProfileUtil
 import app.aaps.core.interfaces.protection.ProtectionCheck
 import app.aaps.core.interfaces.pump.defs.determineCorrectBolusStepSize
+import app.aaps.core.interfaces.queue.Command
 import app.aaps.core.interfaces.queue.CommandQueue
 import app.aaps.core.interfaces.resources.ResourceHelper
 import app.aaps.core.interfaces.rx.AapsSchedulers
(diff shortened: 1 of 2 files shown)


vanelsberg:
Confusing feature.
The "stop" SMB button suggests it is stopping SMB - it does not (just only one single instance) ?

MilosKozak:
this goes against the app logic. automatic events should not be affected by user.
If you know you don't want SMB you should tell it upfront

Der-Schubi:
@MilosKozak related to this PR:
For some pumps bigger boluses (not SMB!) can take multiple minutes.
There were multiple discussions in the past about getting rid of the bolus progress dialog, as the rest of AAPS is inaccessible while the dialog is visible.
My suggestion: We could make the bolus progress dialog closeable or disable it completely and show the stop button from this PR for normal boluses instead of SMBs. So we would still be able to stop a bolus and also the progress is still visible on the bottom.
What would you prefer, a setting to use this button instead of the full progress dialog or maybe a minimize button on the progress dialog and then show this button?

See here for a example: https://github.com/Der-Schubi/AndroidAPS/commits/bolusStopButton/
I could work this into a PR.

jbr7rr:
Another though on this issue, If you enable the setting "wear_notifysmb" you can stop the SMB from the watch.. So behavior is not consistent

I do like the option to minimize and maximize the bolus progress dialog as schubi proposed (for both SMB and normal bolus)

vanelsberg:
For safety reasons I really think it is not a good idea to do large manaul boluses in the background.

When doing a manual bolus (especially the ones taken longer) I do want the option to monitor the bolus progress. Such a large manual bolus I need to know it is executed as expected and I want to know when the bolus is finished. 
In case I have no need to monitor progress I usually switch app, lock the screen or use the watch to initiate the bolus.

That said, I would not have a problem with giving the user the option to minimize the current bolus progress pop-up as an option (with maximized as initial default?)

PR #3441, opened 2024-09-26
Removed check that incorrectly filters existing pumpbases

saFariW:
removed SN check that incorrectly filters existing pumpbases;

After removal tested with ''non'' confirming SN Pumpbase -> no issues found

status
issue
fixed


MilosKozak:
model is reported correctly after this patch?

saFariW:
Yes, The upper screenshot shows the pumpbase type, which matches the pumpbase, see "REF MD8201" on the last image

Issue #3445, opened 2024-09-29
Dev: Medtrum pump description is wrong


jbr7rr:
After f583a6cb1454dae8453cc91bb777cc1561720874

The reservoir reading in status bar is incorrect:
![WhatsApp Image 2024-09-29 at 20 28 32(1)](https://github.com/user-attachments/assets/1800c74f-6eca-41e6-8806-f386f83094ce)
![WhatsApp Image 2024-09-29 at 20 28 32](https://github.com/user-attachments/assets/e9c3856d-0e32-4b7d-8ba8-1736b0bda5b2)

This is due to the pump type being dynamicly updated by the driver but this is not represented to AAPS because the get() was removed.

I'm not sure if this is a problem for other pump drivers, hence the issue.

PR #3446, opened 2024-09-29
Fix medtrum pump description

jbr7rr:
https://github.com/nightscout/AndroidAPS/issues/3445

This fix is only for medtrum, not sure if the causing commit will give issues for other pump drivers if they update pump type dynamically


MilosKozak:
i better reverted the wrong commit
https://github.com/nightscout/AndroidAPS/commit/a8747687af4b64d07d5aed682c1fa6b138caa62b
I did't realize the way you are handling it

PR #3458, opened 2024-10-07
Restore access to logfiles in Android 11 to 15,...? (API level 30+)

Gohtraw:
Several new features in Android 11 made it nigh impossible to access and use the AAPS logfiles.
The changes were introduced with Android 11 but so far seem to (only) affect devices running Android 14 since the end of August 2024.
This fix works on my Pixel 8 and 5 (both running Android 14) and I tested it with success on an emulated Pixel 3 running Android 9.
I regained access by moving the AAPS user files in the Documents folder.
Further I renamed the logs folder in aapsLogs - a bit redundant but searching for "logs" kept spamming me with "dialogs".

Background: [Storage updates in Android 11](https://developer.android.com/about/versions/11/privacy/storage#file-directory-restrictions)
https://developer.android.com/about/versions/11/privacy/storage#file-directory-restrictions

The overview here: https://developer.android.com/training/data-storage lists what can be accessed and what can't.


MilosKozak:
Changing this is counterproductive. It would mean update wiki, explain to users etc. and it would be for maybe only year only
What we need is solution for Android 14

vanelsberg:
Agree,
And can't see the problem anyway?

* Users can get access to recent logfiles using "SEND-LOGS BY EMAIL" (or by connecting an USB cable and go to `\Internal storage\Android\data\info.nightscout.androidaps\files`). In general, only recent logfiles are useful and dev's will have access this way anyway)

* Imho logfiles (=application system files) should **not** be in the phones "Documents" folder.

This PR only adds unnecessary code complexity for an OS version that will be non-supported soon.

Gohtraw:
> Changing this is counterproductive. It would mean update wiki, explain to users etc. and it would be for maybe only year only What we need is solution for Android 14

It works in Android 9 to 14

MilosKozak:
it will not work once we compile for API level 34

Gohtraw:
> Agree, And can't see the problem anyway?
> 
> * Users can get access to recent logfiles using "SEND-LOGS BY EMAIL" (or by connecting an USB cable and go to `\Internal storage\Android\data\info.nightscout.androidaps\files`). In general, only recent logfiles are useful and dev's will have access this way anyway)
> * Imho logfiles (=application system files) should **not** be in the phones "Documents" folder.
> 
> This PR only adds unnecessary code complexity.

The directory \Internal storage\Android\data\info.nightscout.androidaps\ can not be accessed by anything else but the application "info.nightscout.androidaps".
Logs in that folder cannot be read.
Exported settings in that folder cannot be backed up or moved.
Exported user entry CSV files in that folder stay there until the app is deinstalled.

vanelsberg:
> The directory \Internal storage\Android\data\info.nightscout.androidaps\ can not be accessed by anything else but the application "info.nightscout.androidaps". Logs in that folder cannot be read. Exported settings in that folder cannot be backed up or moved. Exported user entry CSV files in that folder stay there until the app is deinstalled.

Nope. Check again?

1. Exported settings go to **/AAPS/preferences**
2. Exported .csv's go to **/AAPS/exports**
3. All logfiles at info.nightscout.androidaps are accessible by **connecting the phone over USB** (send-by-email)

Gohtraw:
> > The directory \Internal storage\Android\data\info.nightscout.androidaps\ can not be accessed by anything else but the application "info.nightscout.androidaps". Logs in that folder cannot be read. Exported settings in that folder cannot be backed up or moved. Exported user entry CSV files in that folder stay there until the app is deinstalled.
> 
> Nope. Check again?
> 
> 1. Exported settings go to **/AAPS/preferences**
> 2. Exported .csv's go to **/AAPS/exports**
> 3. All logfiles at `\Internal storage\Android\data\info.nightscout.androidaps\files`  are accessible by connecting the phone over **USB** (or send-by-email as an alternative)

which android version are you looking at?

MilosKozak:
If I read the docs well, for Android 14 we need to ask user where to store and user must explicitly allow it. Basically /AAPS/ folder location
Not sure about logs. Either the library allows such location or we must move the files on user's request to make them readable

MilosKozak:
does this work when you set target sdk to A11 or later?

Gohtraw:
> does this work when you set target sdk to A11 or later?

minSdk = 31 
I can export and import settings on the phone but I can't seem to read json files created on another phone.
Screenshot 2024-10-22 210217


with SDK lvl 32, trying to export settings I get a java.lang.SecurityException when I'm exporting settings because 
"Settings key: <bluetooth_name> is only readable to apps with targetSdkVersion lower than or equal to: 31"
But I think that's another new Google "security feature"

vanelsberg:
@Gohtraw @MilosKozak This PR contains other changes that do not belong c.q. change AAPS behavior?
- PumpCapability.kt
- XdripSourcePlugin.kt

vanelsberg:
Really do not like AAPS dir (=app system files) being in my phones /Documents folder. This folder should only contain users **documents**. For some apps (like Office365) my documents view would get cluttered with .log and .zip files! 😕

If, then at least make this change active when Android version = 11 only, or maybe optional through user setting.

Gohtraw:
> Really do not like AAPS dir (=app system files) being in my phones /Documents folder. This folder should only contain users **documents**. For some apps (like Office365) my documents view would get cluttered with .log and .zip files! 😕
> 
> If, then at least make this change active when Android version = 11 only, or maybe optional through user setting.

read this: https://developer.android.com/training/data-storage

MilosKozak:
> Really do not like AAPS dir (=app system files) being in my phones /Documents folder. This folder should only contain users **documents**. For some apps (like Office365) my documents view would get cluttered with .log and .zip files! 😕
> 
> If, then at least make this change active when Android version = 11 only, or maybe optional through user setting.
It's exactly like it has been written. Once we target sdk 30 (that's something I want to do - and even more), we lose access to filesystem except:
/Documents - almost free for use
android data directory (original AAPS dir location) - but there are some disadvantages:
 * user cannot access files there from another apps ie. file managers to copy files from
 * directory is removed on uninstallation - not much usefull for storing backups etc.

I believe we end up with Documents directory

PR #3460, opened 2024-10-11
Migrate kapt to ksp

anbraten:
Migrates `kapt` to `ksp` to be able to use `compose` in the future.

# Changes

- [x] replace `kapt` with `ksp` 
- [x] bump kotlin to `2.0.21`
- [x] bump dagger to `2.52`
- [ ] migrate code to use new annotations / fix types

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -6,6 +6,7 @@ plugins {
     id("com.android.application")
     id("kotlin-android")
     id("kotlin-kapt")
+    id("com.google.devtools.ksp")
     id("com.google.gms.google-services")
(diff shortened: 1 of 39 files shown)


MilosKozak:
Interresting. I thought we cannot switch to ksp because of xml layouts ...

anbraten:
AFAIK they can be both used simultaneously. Kapt is required as long as data-bindings are used. XML in general should not be a problem.

MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/485e27781f02f43e06ae2d2cdda8378cf44f1fd8
thanks for hint!

PR #3469, opened 2024-10-15
Handle unknown DB GlucoseValue SourceSensors

### LINKED ISSUE #3466: Regression with BG type ??
I have problems with updating from a 2 weeks old dev to the last dev version. Seems like something went bad in a commit....

Build is ok, update is ok. AAPS craches systematically when starting. Clearing cache doesn't help

This is the logcat when the upgraded AAPS tries to start.

plummer86:
Fixes #3466

Updated toSourceSensor to return 'UNKNOWN' SourceSensor for GlucoseValues that have undefined sensor enum values. This change prevents toSourceSensor from raising an exception when the source sensor enum is not found.

### DIFF
--- a/database/impl/src/main/kotlin/app/aaps/database/Converters.kt
+++ b/database/impl/src/main/kotlin/app/aaps/database/Converters.kt
@@ -60,7 +60,12 @@ class Converters {
     fun fromSourceSensor(sourceSensor: GlucoseValue.SourceSensor?) = sourceSensor?.name
 
     @TypeConverter
-    fun toSourceSensor(sourceSensor: String?) = sourceSensor?.let { GlucoseValue.SourceSensor.valueOf(it) }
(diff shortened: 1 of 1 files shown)


database/impl/src/main/kotlin/app/aaps/database/Converters.kt
```diff
@@ -60,7 +60,12 @@ class Converters {
     fun fromSourceSensor(sourceSensor: GlucoseValue.SourceSensor?) = sourceSensor?.name
 
     @TypeConverter
-    fun toSourceSensor(sourceSensor: String?) = sourceSensor?.let { GlucoseValue.SourceSensor.valueOf(it) }
+    fun toSourceSensor(sourceSensor: String?): GlucoseValue.SourceSensor? {
+        return sourceSensor?.let {
+            GlucoseValue.SourceSensor.entries.find {
```
MilosKozak, line 65: firstOrNull is the corect kotlin way

Issue #3296, opened 2024-04-05
[Suggestion] export settings from maintenance go to dowloads folder or straight to google drive


Not-AriStienfeld:
Without a proper file sorting system, it can be annoying (and potentially confusing) to get the settings export onto the cloud. The option to send it straight to google drive would probably result in more people backing up more frequently, and would be more helpful for people who aren't as good at working with technology.

MilosKozak:
Storing config to some cloud is a way .... But changing diricetory will be impossible with upcomming Android versions anyway

vanelsberg:
Suggestion: I'm using FolderSsync to automagically sync AAPS settings the (cloud) storage of my choice.
(https://play.google.com/store/apps/details?id=dk.tacit.android.foldersync.lite&pcampaignid=web_share)

olorinmaia:
xDrip has automatic backup to google drive which is very useful

andyrozman:
@MilosKozak Could we add some sort of automatic backup like that? And we could see how Xdrip has this handled... My proposal would be to create export in such way that we just add date in readable format to the end (for example settings_20240702_1202.zip) and then copy it to specific google drive directory...
Is it even possible to create export now that we need password for creating export?

MilosKozak:
the problem is encryption of backup. AAPS doesn't know master password

KelvinKramp:
@MilosKozak  => maybe we can ask the user to input the password when it's time for a backup? 

I recently ran into the same issue. It has also been discussed in 2022 here: https://github.com/nightscout/AndroidAPS/issues/1208#issuecomment-1018023093

The options provided there were:
1. Automatically save settings on a regular basis (every day, keep last 5 days, last 4 nondays, last 12 months)
2. Automatically save settings after enabling a new omnipod (or other event requiring data from settings)
3. Provide an option to keep a copy of the settings in an alternate location
4. Provide an option to keep a copy of the settings in a predefined cloud location (note IMHO, since we are on android, google drive might be better suited than dropbox)

I would vote for a combination of 1 and 4 and for 4 (although more cumbersome) we could use mongodb via Nightscout for backupstorage
https://www.mongodb.com/developer/products/mongodb/storing-large-objects-and-files/

vanelsberg:
> the problem is encryption of backup. AAPS doesn't know master password

[...]

PR #3487, opened 2024-10-22
PR3487 unattended settings export (final)

vanelsberg:
This PR enables unattended exports
(see also issue #https://github.com/nightscout/AndroidAPS/issues/3296)

**Functional short description:**

The master password should be entered as usual when manually exporting settings from the AAPS maintenance menu.
When the "unattended exports" option is enabled in maintenance preferences the password entered is encrypted and securely stored on the user's phone (not in AAPS).

Subsequent exports will no longer require the user to enter the master password until it expires or is reset.
- A new Pod Activation automation is added for automating settings export.
- A new Trigger is added for triggering an action on Pod Activation

When active, the new "Export settings" automation will alert in AAPS overview on exporting and (when relevant) not being enabled or password expiry. The user can (grace period) or is required (expired) to reenter the password executing a manual export.

**Choice:**

User has a choice to enable/disable unattended exports through the maintenance settings.
When disabled, no password is stored and unattended exports are unavailable.

**When enabled:**

Password will be securely stored encrypted on the local phone's Android DataStore (1).
Encryption key needed for decrypting is generated and protected by the local phone's Android KeyStore (2).
To ensure user needs to "maintain" the master password it will expire after 4 weeks with a grace periode of 1 week (non-configurable)
Stored password will be removed on changing the master password or AAPS password reset.
Importing setting or other functionality that needs entering the master password/PIN/Biometrics are not affected.

_Ad1) Android DataStore:_
_This system provides a robust and flexible way of storing key-value pairs on the phones storage._

_Ad2) Android Keystore:_
_This system allows to store cryptographic keys in a secure container, making them difficult to extract from the device._

**Testing:**

- Password expiry/grace period (it is set fixed at 4 weeks+1 week grace):
For testing purposes, placing a semaphore file named `DebugUnattendedExport` (no extension!) will set expiry to 1 day with a grace periode of 1 day.

### DIFF
--- a/app/src/main/kotlin/app/aaps/MainActivity.kt
+++ b/app/src/main/kotlin/app/aaps/MainActivity.kt
@@ -502,10 +502,10 @@ class MainActivity : DaggerAppCompatActivityWithResult() {
      * clear password stored in datastore if file exists
      */
         private fun exportPasswordResetCheck(context: Context) {
-        val exportPasswordReset = File(fileListProvider.ensureExtraDirExists(), "ExportPasswordReset")
-        if (exportPasswordReset.exists()) {
+        val fh = File(fileListProvider.ensureExtraDirExists(), "ExportPasswordReset")
+        if (fh.exists()) {
             exportPasswordDataStore.clearPasswordDataStore(context)
-            exportPasswordReset.delete()
+            fh.delete()
             ToastUtils.okToast(context, context.getString(app.aaps.core.ui.R.string.datastore_password_cleared))
         }
     }
--- a/app/src/main/kotlin/app/aaps/MainActivity.kt
+++ b/app/src/main/kotlin/app/aaps/MainActivity.kt
@@ -43,6 +43,7 @@ import app.aaps.core.interfaces.notifications.Notification
 import app.aaps.core.interfaces.plugin.ActivePlugin
 import app.aaps.core.interfaces.plugin.PluginDescription
 import app.aaps.core.interfaces.profile.ProfileFunction
+import app.aaps.core.interfaces.protection.ExportPasswordDataStore
 import app.aaps.core.interfaces.protection.ProtectionCheck
 import app.aaps.core.interfaces.rx.AapsSchedulers
 import app.aaps.core.interfaces.rx.events.EventAppExit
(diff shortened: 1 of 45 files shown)


vanelsberg:
@MilosKozak 
This PR3487 **replaces PR3463**
 
To dev's that would like to test, please do (thanks!):
I think this could be the "final" PR, ready for "field testing" finding possible bugs?

MilosKozak:
I found only one thing to consider. Assume automatic export on child's phone. The child will ignore the warning regarding re-entering the password (own experience). Backup will stop work without parent's knowledge.
In this case i'd create "Announcement" (rather than "Settings export") event because it pop-ups in NS as notification.

vanelsberg:
> child's phone.

Yes. Will look into that. There is already alerting on "about" to expire and "expired" on overview and on Care portal.
Will check/make sure this also goes to NS.

(B.t.w: main reason for the Settings export event is because Pod Chance triggers by when latest Cannula change event is new then latest Settingsexport event)

vanelsberg:
Password expiry/grace period (it is set fixed at 4 weeks+1 week grace):

For testing purposes, placing a semaphore file named `DebugUnattendedExport` (no extension!) will set expiry to 1 day with a grace periode of 1 day.

vanelsberg:
Fixed build: rebase on current dev commit#c12639b7 + fix .toml dependencies)

WesSec:
@vanelsberg 

Error found on in logcat, i recently switched back from AIMI to dev. the key was likely still stored somewhere and not deleted upon deletion of the AIMI branch. If i understand the error correctly: I suggest checking if the key already exists before storing it. 

`2024-11-26 08:48:29.681 28698-28698 PreferenceGroup         info.nightscout.androidaps           E  Found duplicated key: "enable_unattended_export". This can cause unintended behaviour, please use unique keys for every preference.`

vanelsberg:
> ... duplicated key: "enable_unattended_export".

Thanks! Will check on this.

MilosKozak:
fixed in https://github.com/nightscout/AndroidAPS/commit/2d0944ac8735ed10be966df74374ffc959fbbf3b

Issue #3433, opened 2024-09-12
[Feature Request]: Bolus Calculator in AAPS Client, Record Carbs and Calculation only


gerison77:
Case: 
Several Caretakers (incl. parents) with AAPS Client in a  Master/Client setup (parental setup) for kids in school. 
Bolus is administered via SMS.

Current situation (AAPS 3.2.0.4): 
Meal Bolus is calculated via Bolus Calculator on client. Bolus is administered via SMS. Carbs are recorded manually. Calculation is not recorded (and therefore can't be checked later on).

Problem to solve: 
Record Result of claculation and carbs without recording Bolus. Being able to check calculation in order to improve settings. 

Solution and feature request:
New feature in Bolus Calculator: Record Carbs and Calc only, skip Bolus. Which would be administered via SMS.

Ps: This feature would come in handy as well, if Pump is disconnected an Insulin is administered via Pen. I as I learned, the feature to record boli administered by pen with disconnected pump is added already in the current dev.

szantos:
IMHO basically a good idea, but only the calculation should be recorded, without Carbs for safety reasons to avoid overdosing.

If Carbs get recorded, it immediately triggers the algo to recalculate and it'll find there is COB, but not enough IOB - therefore it'll dose insulin. Then the SMS command arrives with the originally calculated insulin which gets dosed, too. In total there will be more IOB as should be.

gerison77:
Good Point, but to keep things simple I would encounter this issue with a fixed delayed entry of carbs, say 2 to 5 min.

Sanitariuszek:
I would add suggestion not to request bolus password each time when calc window is opened but only on confirming, when bolus/carbs are really going to be added. Many times I open calc just to see some numbers,entering password every time is annoying.

gerison77:
can be closed

olorinmaia:
@gerison77 @MilosKozak can be closed due to new master/client signed commands in dev

PR #3507, opened 2024-10-31
sms bolus from AAPSClient

jgr-lab:
Implement bolus sms feature in AAPS client, if feature is enabled in menu and target phone number is provided it allows to send SMS directly from insulin/calc wizard dialog instead of "record only" feature. It still requires authorization code to be typed manually in the second message. If calc wizard was used, results will be saved in "treatment" tab history.


mushroom-dev:
Implementes #3507 and #3433

app/src/main/res/values/strings.xml
```diff
@@ -47,6 +47,9 @@
     <string name="remove_bg_readings">Remove BG readings</string>
     <string name="identification_not_set">Identification not set in dev mode</string>
     <string name="a11y_dialog">dialog</string>
+    <!-- AAPS client sms -->
+    <string name="sms_receiver_number_dialog">If number is provided, insulin and calculator buttons will be able to send SMS command to inject bolus. Note that you still need to confirm it and reply with security token/code. Note that you need to enabled SMS control of AAPS phone to use this function.</string>
+    <string name="client_allow_sms">Allow AAPS client to send SMS with bolus command</string>
```
olorinmaia, line 52: Typo here, AAPSClient instead of 'AAPS client'

core/ui/src/main/res/values/strings.xml
```diff
@@ -446,9 +446,11 @@
     <string name="cobvsiob">COB vs IOB</string>
     <string name="slowabsorptiondetected"><![CDATA[<font color=\'%1$s\'>!!!!! Slow carbs absorption detected: %2$d%% of time. Double check your calculation. COB can be overestimated thus more insulin could be given !!!!!</font>]]></string>
     <string name="partialboluswizard">Deliver this part of bolus wizard result [%]</string>
+    <string name="sms_receiver_number">Receiver phone number for sms commands send from AAPS client phone</string>
```
olorinmaia, line 449: Some typos here.
Sent instead of send
AAPSClient instead of AAPS client

olorinmaia:
One concern is that AAPSClient doesn't always sync perfectly with AAPS. Both IOB and COB can be wrong due to many different reasons. We should consider adding a warning about this in the confirmation box of the bolus calculator in AAPSClient if enabling this feature since bolusing based on IOB/COB from AAPSClient's bolus calculator will then be more tempting to use when it's more accessible. 

I don't think there is a way with SMS commands to verify what IOB and COB AAPS really got. It's rare for my part that AAPSClient and AAPS got different IOB/COB, because if sync settings and NS is setup correctly, it's very unlikely to happen. But when looking at threads for support in FB/Discord/Github regarding wrong sync between AAPS and AAPSClient this is a concern I have. 

@MilosKozak In the OpenAPS log you can click to view in AAPSClient you can see the exact COB from AAPS; regardless if COB in NS / AAPSClient is wrong. Would it be possible to add IOB amount to the openAPS-log also? That way there is one place one could verify that COB/IOB in AAPSClient is correctly synced with AAPS without visiting master phone.

core/objects/src/main/kotlin/app/aaps/core/objects/wizard/BolusWizard.kt
```diff
@@ -536,13 +545,18 @@ class BolusWizard @Inject constructor(
                                 ValueWithUnit.Gram(this@BolusWizard.carbs).takeIf { this@BolusWizard.carbs != 0 },
                                 ValueWithUnit.Minute(carbTime).takeIf { carbTime != 0 })
                         )
-                        commandQueue.bolus(this, object : Callback() {
-                            override fun run() {
-                                if (!result.success) {
-                                    uiInteraction.runAlarm(result.comment, rh.gs(app.aaps.core.ui.R.string.treatmentdeliveryerror), app.aaps.core.ui.R.raw.boluserror)
+                        var phoneNumber = preferences.get(StringKey.SmsReceiverNumber)
+                        if (preferences.get(BooleanKey.SmsAllowRemoteCommands) && !phoneNumber.isNullOrBlank()) {
```
mushroom-dev, line 549: I will try to encapsulate it somehow

mushroom-dev: I give up, I don't have time to resolve all the issues with gradle and dependencies during build

mushroom-dev:
@olorinmaia  @MilosKozak what are next steps?

olorinmaia:
I'm not able to test this as my test-rig don't have sim-cards. One way is to inform about PR on Discord and see if there is any testers there and get more input regarding this PR generally.

mushroom-dev:
@MilosKozak can your review it? I assume groups of people building dev branches and parents needing features I implemented are disjoint. I asked about feedback on parents channel, didn't get any feedback. I have been testing it myself for some time.

emilisev:
I am testing this on top and beta1 and first use went all right. I will continue to test in the next few days.

emilisev:
When using the bolus assistant or quickwizard, the carbs are lost !
Screenshot_20241208-124246
Screenshot_20241208-124437

mushroom-dev:
@emilisev thanks for catching that! See my latest update for fixes, I found few other corner cases. I also moved sms related settings under button menu as you suggested.

@olorinmaia  I added current IOB to sms request response.

I've tested AAPSClient, haven't testes AAPS with updated SMS response yet, but I don't expect any issues there

emilisev:
I can't build it, I get : e: file:AndroidAPS-dev/plugins/main/src/main/kotlin/app/aaps/plugins/main/general/smsCommunicator/SmsCommunicatorPlugin.kt:936:65 Unresolved reference 'plus' for operator '+'.

mushroom-dev:
I've pushed compilation fix. Note I didn't test it functionally yet. You are free to do it or you can wait for me. I will send update

mushroom-dev:
@emilisev I did some testing, haven't found any issues

emilisev:
Builds all right and carbs are not lost anymore, but I have encountered another case which seems problematic. If I use the bolus wizard from aapsclient but the bolus fails (previous bolus too close or wrong pin....), the carbs are still entered. I think we should be careful with that, it would be an issue is the parent misses the carbs entry, enters again the bolus wizard and the carbs are entered twice... I think this is quite risky...

mushroom-dev:
Personally I am fine with it. It works the same way with bolus alerts. Even if bolus fail timer is not stopped, if you provide another bolus with alarm then you have 2 countdowns. Or maybe it could be improved too? I don't think there's easy way to record carbs only after insulin was successfully administered. 

Maybe I could create some job which awakes every 20s to check if there's new carb entry in treatments tab, but I am not sure if such mechanism fits AAPS workflow. Also in that case some other person could enter carbs on primary phone, such scenario is also technically possible.

Personally as caregiver I am fine with current solution, if bolus fails I can always edit treatments tab, it's still big improvement for me. I will need to find time to fix that countdown issue though :)

gerison77:
I tested it on 3.3.0-rc1 and it works as expected. I wouldn't miss that feature. It is a big improvement. Especially when you have to kids to remote bolus sometimes. Furthermore it helps to check the calculations afterwards.

I want to express my urgent wish this to become a feature in the standard release!

mushroom-dev:
I will try to postpone carbs entry after receiving confirmation about bolus delivery

emilisev:
I've been using the functionality almost on a daily basis for the last 2 weeks and it's growing on me :-) I never used the sms commands before, and just started using it to help test your functionality, but I have to admit it is quite useful!
For the "carb when bolus failed" behavior, I guess you get used to how it works. You should probably get more feedback on the matter, not just my own opinion :-)

mushroom-dev:
Sorry for long delay. I've made quite a lot of changes. I introduced new sms command "boluscarbs", as it's name indicates you can now send single sms with information about insulin and carbs. It's processed together so there should be no issue with carbs entry when insulin delivery fails. It allows also to set remote alarm on AAPS phone. I've made also few other minor changes

core/ui/src/main/res/values/strings.xml
```diff
@@ -454,13 +454,16 @@
...
+    <string name="bolus_advisor_not_supported">Bolus advisor doesn\'t work with SMS commands. Wait till BG drops or provide correction bolus using Insulin or Treatment buttons.</string>
 
     <string name="cobvsiob">COB vs IOB</string>
     <string name="slowabsorptiondetected"><![CDATA[<font color=\'%1$s\'>!!!!! Slow carbs absorption detected: %2$d%% of time. Double check your calculation. COB can be overestimated thus more insulin could be given !!!!!</font>]]></string>
     <string name="partialboluswizard">Deliver this part of bolus wizard result [%]</string>
+    <string name="sms_receiver_number">Receiver phone number for sms commands sent from AAPSClient phone</string>
     <string name="partialboluswizard_reset_time">Old glycemia time threshold [min]</string>
     <string name="bolus_constraint_applied_warn">Bolus constraint applied: %1$.2f U to %2$.2f U</string>
     <string name="bolus_recorded_only">Bolus will be recorded only (not delivered by pump)</string>
+    <string name="sms_request_notification">SMS command will be send</string>
```
olorinmaia, line 466: "SMS command will be send" seem to be bad grammar.

"SMS command will be sent" or "The SMS command will be sent" is probably better if I understand the context. :)

gerison77:
I have been testing it now for one week on two rigs. Since the most recent update, I experienced no problems. I it a great improvement. Saves time and accurate entries of carbs.

mushroom-dev:
@MilosKozak it's ready for your review.

core/keys/src/main/kotlin/app/aaps/core/keys/BooleanKey.kt
```diff
@@ -70,7 +70,7 @@ enum class BooleanKey(
     AutotuneCircadianIcIsf("autotune_circadian_ic_isf", false),
     AutotuneAdditionalLog("autotune_additional_log", false),
 
-    SmsAllowRemoteCommands("smscommunicator_remotecommandsallowed", false),
+    SmsAllowRemoteCommands("smscommunicator_remotecommandsallowed", false, showInApsMode = false, showInPumpControlMode = false),
```
MilosKozak, line 73: Why this?

core/objects/src/main/kotlin/app/aaps/core/objects/wizard/BolusWizard.kt
```diff
@@ -86,9 +91,12 @@ class BolusWizard @Inject constructor(
     @Inject lateinit var processedDeviceStatusData: ProcessedDeviceStatusData
 
     var timeStamp: Long
+    // var phoneNumber: String
```
MilosKozak, line 94: review https://github.com/nightscout/AndroidAPS/pull/3507/files and cleanup the code

core/data/src/main/kotlin/app/aaps/core/data/configuration/Constants.kt
```diff
@@ -11,7 +11,7 @@ object Constants {
     const val notificationID = 556677
 
     // SMS COMMUNICATOR
-    const val remoteBolusMinDistance = 15 * 60 * 1000L
+    const val remoteBolusMinDistance = 5 * 60 * 1000L
```
MilosKozak, line 14: Why?

core/keys/src/main/kotlin/app/aaps/core/keys/BooleanKey.kt
mushroom-dev: This plugin is intended only for AAPSClient. It sends SMS commands to phone running AAPS. Both AAPS and control app are on the pump controlling phone so presence of that plugin would be confusing.

core/data/src/main/kotlin/app/aaps/core/data/configuration/Constants.kt
mushroom-dev: 15 minutes seems to be too much. In my case I've set some max carbs limit for single meal. It's intended to limit bolus to make sure carbs absorption kicks in on time. When meal is bigger than max carb then I split calculations into 2 meal with 5 minutes delay. 5 should also prevent accidental bolus at the same time from multiple caregivers

gerison77: Totally agree. I have the same process in place with the treatment on my two kids. Set bolus limit for safety, leading to mandatory split bolus for regular meals.

mushroom-dev:
@MilosKozak can you review again?

gerison77:
I have tested it now on 3.3.3.0-dev since it's release without any issues on two rigs.  

From a practical point of view, being able to track the bolus calculation via bolus calculator in the treatments tab proofs valuable in order to draw conslusions of futher treatments and for general analysis. Correct SMS commands proof to save time, especially since it prevents typos which occur quite often. 

I furher suggest to extend the funcionalty on pump commands, Connect/Disconnect Pump, Stop/Enable loop. Which I have to use regualrily on sport aktivities like supervising my kids on Karate Trainings, MTB Tours where the setting of temp targets or % of Porfile just doesn't suffice to handle the situation. I am looking forward to have this functionality as well.

gerison77:
Almost three months have passed since my last post, and I can now confirm: still no issues whatsoever.

From a practical perspective, the ability to track bolus calculations in the Treatments tab has proven consistently helpful for evaluating therapy decisions and overall analysis. The use of structured SMS commands continues to save time and prevents common input errors.

I stand by my recommendation to pull this—at least into dev.

If there are any concerns or objections, I’d appreciate hearing them so I can evaluate whether they’re practically relevant and warrant further attention.

gerison77:
Two months later, I can confirm the feature works well and meets expectations. Since it’s been pending here for quite a while, I’d like to discuss potential reasons against implementing it.

jgr-lab:
@MilosKozak  can you review?

gerison77:
I know this PR has been idle for quite some time, but I would like to bring the topic back on the table because the use case is still very relevant for caregivers.

Recently I found this project very inspiring:  
https://github.com/painbrain81/AAPS-SMS

Typical situations for us are: both kids racing their bikes, one of them at school or at a friend’s place, calling or sending a short message that they want to eat, and a bolus has to be given remotely, sometimes with or without using the full calculator.  In these real-time scenarios, having SMS bolus and similar features directly integrated in AAPSClient would significantly speed things up and reduce caregiver workload.

dmelliott20:
Any reason this PR can't make it to release? Could be useful for a number of us....

Philoul:
This PR uses a workaround (SMS command) to manage boluses from AAPSCLIENT.
Huge improvement concerning remote management (from AAPSCLIENT), are included within the next release of AAPS 😉.

MilosKozak:
this is replaced by master/client control

PR #3521, opened 2024-11-08
FIX: Register manual settings export

vanelsberg:
Minor fix:
In addition to unattended settings exports also register manual exports.


vanelsberg:
In addition to unattended settings exports also register manual exports.

PR #3523, opened 2024-11-11
CWF External data improvements

Philoul:
Update external views only if at least one external view is visible


Philoul:
2 full days with this commit and watch autonomy finally looks equivalent to previous external data management...
(I don't know if it's due to this commit or due to no "debugging mode on watch" nor "software update" often done during dev)

PR #3525, opened 2024-11-12
Fix Tidepool upload: filter out negative carbs

vanelsberg:
Solves Tidepool integration upload FAILED (response 400 shows invalid carbs)
FIX filter out negative carbs for uploading to Tidepool.


vanelsberg:
Merged with latest dev (A11) and tested ok.

vanelsberg:
Closed:
Does not merge correctly with dev :-|
Looking into this...

PR #3536, opened 2024-11-17
Watch Add User Action Tiles for Automation

Philoul:
Add Tiles for Up to 4 User Actions Automation


plugins/automation/src/main/kotlin/app/aaps/plugins/automation/AutomationFragment.kt
```diff
@@ -128,7 +129,10 @@ class AutomationFragment : DaggerFragment(), OnStartDragListener, MenuProvider {
         disposable += rxBus
             .toObservable(EventAutomationDataChanged::class.java)
             .observeOn(aapsSchedulers.main)
-            .subscribe({ eventListAdapter.notifyDataSetChanged() }, fabricPrivacy::logException)
+            .subscribe({
+                           eventListAdapter.notifyDataSetChanged()
+                           rxBus.send(EventWearUpdateTiles())
```
MilosKozak, line 134: This will executed only when the fragment is visible ;)

Philoul: Yes, it I put it here to have a synchro between automation plugin and tiles when rules are updated and see immediatly result within watch (I did the same within overview).
When none of both fragment are visible, synchro is done every 5 min.
So on temporary user action associated with an additional trigger, we can have tiles updated a few min after real time...

vanelsberg:
Merges OK. Works for me.

Issue #2335, opened 2023-01-10
AAPS/NSClient watchface improvement - add temp target display option


olorinmaia:
Absolutely loving the watch faces in the wear app from NSclient and AAPS. The only thing missing is an option to display temp-target. I know I can see it if I go to settings, but that only applies to AAPS watchface.

Maybe a thin grey line like in Nightscout could illustrate that a temp-target is active?

That way it will be easier to remember to turn of temp-target if it's displayed on the watch faces when active.

olorinmaia:
Don't need to be a visual thing either. Maybe if TT is active text "TT: target-value" could show on custom watch face when TT is active. @Philoul

olorinmaia:
Or another option, like on Widget:
If normal target display value with default font-color.
If target is adjusted by autosens or DynIsfSens display value with green font-color.
If temp target display value + (duration in hour and minutes') with yellow font-color.

olorinmaia:
@Philoul is this now easier to accomplish with latest changes in dev?

Philoul:
It's neither easier nor more complicated...
I just have to take some time to develop it 😉.
=> I have to add 3 views within custom watchface layout (one to manage main user, and 2 additional views for external users now available for the followers), update code to manage these views with text and colors, and update communication to send data.
Maybe add a new parameter to manage if Target should be visible or hidden in watchface (😱 we are trying to reduce them to simplify settings... see topic opened yesterday in ux-design channel 😂)

Then users will have to create new CWF or update existing one to include this new information.

Philoul:
See PR #3551 ;-)

olorinmaia:
Ty @Philoul

PR #3551, opened 2024-11-23
CWF Add TempTarget Management with TT Color

Philoul:
For feature request #2335


Philoul:
To test it I updated a bit Analog Watchface and I replaced BGI view by TempTarget View below Carbs...
White, Yellow Green according to TT (only tested White and Yellow...)
=> Note that this will not work if you have Target-Low and Target-High

[Analog G-Watch_TT.zip](https://github.com/user-attachments/files/17880527/Analog.G-Watch_TT.zip)
CustomWatchface

Philoul:
Green color when TT is adjusted by the loop works as expected 😉

olorinmaia:
I can confirm this is working  :) I tried to modify AAPSV2 to display TT instead of BGI or Rig Battery, but I don't seem to be able to make it work. The watchface doesn't appear after uploading zip-file :)

Philoul:
Check if Show Temp Target is enabled in preferences (within AAPS app in watch)

Philoul:
Be carefull by default font is white if no TT is set, if divider has white background you will not see TT...

PR #3572, opened 2024-12-01
Wear Fix TrendArrow

### LINKED ISSUE #3414: Trend arrow in Wear watchface does not account for smoothing
The trend arrow in the Wear watchfaces apparently do not account for smoothing. See the two attached images below. I've tested it both stock and personalized watchfaces and both have the issue. Collector is a Libre2Plus (CGM Brazilian version) with updates every 1 minute, sent by Juggluco with EverSense interface and received via the MM640g source.

Version: AAPS 3.2.0.1 (aeed6e2083-2023.10.24)

I think logs would not help, but please correct me if they are necessary.

Screenshot_20240824_212752_sysui
Screenshot_20240824_212805_AAPS

Philoul:
Fix #3414...
I never got this issue, but I modified code sent to watch to use same command than in OverView, So it should fix this issue ;-)


Philoul:
Strange, it includes my previous commit already merged in dev???
Will create a new PR clean...

PR #3605, opened 2024-12-11
Hide certain Dash settings instead of not clickable when not relevant

olorinmaia:
https://github.com/user-attachments/assets/c303fd05-89cd-4630-8cb5-b90d47061ee8


olorinmaia:
I'll see what I can do. Thanks for feedback.

PR #3612, opened 2024-12-12
UI : Better color for light mode

TheNeskik:
This PR is about changing some color when using light mode.
Some color choice make some button or information hardly readable (bottom button for example). 
That's why I'm proposing some color changes that improve readability.

Old color :
Screenshot_20241212_193046

New :
Screenshot_20241212_193526

### DIFF
--- a/core/ui/src/main/res/values/colors.xml
+++ b/core/ui/src/main/res/values/colors.xml
@@ -18,24 +18,24 @@
     <color name="secondaryColorDefault">#03dac6</color>
     <color name="secondaryLightColorDefault">#e3e3e3</color>
     <color name="secondaryDarkColorDefault">#848484</color>
-    <color name="primaryTextColorDefault">#ffffff</color>
+    <color name="primaryTextColorDefault">#000000</color>
     <color name="secondaryTextColorDefault">#000000</color>
     <color name="overviewPillColorDefault">#1c171c</color>
     <color name="colorOnPrimaryDefault">#FFFFFF</color>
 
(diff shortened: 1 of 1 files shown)


MilosKozak:
others: thoughts?

Philoul:
First to be cristal clear, I'm not using myself the light theme...

I think it's better to not have exactly the same background color for the buttons compare to graph background (Profile/target on the top and 4 buttons on the bottom...).
Maybe grey is a bit too dark currently, with not enough constrast with icon colors, but I think proposal is "too light" and we should still have visible differentiation. I think we should also have graph menu buttons background color also harmonized (it's another reason to not have the same background color compare to graphs)

Other comment, black font for profile and Target is much better...

TheNeskik:
You are right ! Buttons must be on different color to make it easier to find.

I also find Temp Basal hard to read but I didn't want to change it as it will differ from dark mode.

Screenshot_20241213_115818

kenzo44:
@TheNeskik looks good! Can you remove the pink background and just have solid white.

For the temp basal, we can change the color to a blue that works well with both light and dark

Philoul:
> For the temp basal, we can change the color to a blue that works well with both light and dark

I wouldn't change any color used on dark mode, everything has been harmonized (including watchface colors and tiles button).
The same colors look different if associated with a dark or a light background, so I think we could cheat a bit with the blue color to make it "look the same" on light mode (and improve constrast with background) even if a bit different compare to the blue used dark mode 🤔

Issue #3631, opened 2024-12-17
[Rc1] Minor issue open loop notifications on watch


olorinmaia:
Notification on watch could be improved to show same as on phone.

![Screenshot_20241217_192422_sysui](https://github.com/user-attachments/assets/5ec32803-b1be-4012-98a8-7df5bf7f8ef9)
![Screenshot_20241217-192457](https://github.com/user-attachments/assets/d0c727d4-4cfa-4cb4-af3d-b5c645b57bc1)

olorinmaia:
fe6e130

PR #3632, opened 2024-12-18
Improve wear open loop notification

olorinmaia:
My proposal to fix/improve https://github.com/nightscout/AndroidAPS/issues/3631

Before:
image

After:
image
image

### DIFF
--- a/core/objects/src/main/kotlin/app/aaps/core/objects/aps/DetermineBasalResult.kt
+++ b/core/objects/src/main/kotlin/app/aaps/core/objects/aps/DetermineBasalResult.kt
@@ -124,9 +124,9 @@ class DetermineBasalResult @Inject constructor(val injector: HasAndroidInjector)
             // rate
             var ret: String = if (rate == 0.0 && duration == 0) "${rh.gs(R.string.cancel_temp)} "
             else if (rate == -1.0) "${rh.gs(R.string.let_temp_basal_run)}\n"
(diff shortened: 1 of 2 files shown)


core/objects/src/main/kotlin/app/aaps/core/objects/aps/DetermineBasalResult.kt
```diff
@@ -124,9 +124,9 @@ class DetermineBasalResult @Inject constructor(val injector: HasAndroidInjector)
             // rate
             var ret: String = if (rate == 0.0 && duration == 0) "${rh.gs(R.string.cancel_temp)} "
             else if (rate == -1.0) "${rh.gs(R.string.let_temp_basal_run)}\n"
-            else if (usePercent) "${rh.gs(R.string.rate)}: ${decimalFormatter.to2Decimal(percent.toDouble())}% (${decimalFormatter.to2Decimal(percent * pump.baseBasalRate / 100.0)} U/h) " +
+            else if (usePercent) "${rh.gs(R.string.rate)}: ${decimalFormatter.to2Decimal(percent.toDouble())}% (${decimalFormatter.to2Decimal(percent * pump.baseBasalRate / 100.0)} ${rh.gs(R.string.profile_ins_units_per_hour)}) " +
```
MilosKozak, line 127: this is historicaly wrong. it's not possible to translate it correctly to languages that differs too much.
create new string like "Rate: %1$.2f%% (%2$.2f U/h)"
I believe many of such strings already exists can be reused

olorinmaia: For some reason turning this into a single rh.gs proved to be difficult. I will look at it later.

MilosKozak:
@Philoul can you verify?

Philoul:
I opened the loop a few min to test the PR 😉...
Looks good, same notification on watch compared to phone, and I could validate the proposal from the watch to apply proposed action...

PR #3650, opened 2024-12-22
Automation: add Steps Count trigger

tdrkDev:
Triggers while meets condition based on steps count per user-selected measurement duration (predefined set comes from `SC` model).

Unit test works and succeeds, real life test succeeds too (but AAPS is lowering the amount of steps for some reason, at my running speed, AAPS reports only 90-110 steps per 5 minutes).

I'd be happy to use `getLastStepsCountFromTime` from `StepsCountDao`, but it is not present even in AppRepository and I'm lazy, so I've used `getStepsCountFromTime().last { ... }`, which works as expected in a real life and in theory, according to SQL query in `StepsCountDao`.

This may be a useful trigger in case of forgotten "Exercise" TT or any similar one.


MilosKozak:
@Philoul @olorinmaia Anyone tested it?

olorinmaia:
@MilosKozak Tested OK. I walked around 400 steps. During that time I adjusted automation rule from 200 to 300, to 400 steps and it activated each time. I also tried reducing the time period it should consider, that also seemed to be working :) Great feature. Could see some uses here.

PR #3658, opened 2024-12-27
Add BR & IoB complication for Wear OS

keweki:
Adding a new Wear OS complication that display basal rate and insulin on bord without carbs on board
Usefull with AutoISF plugin (carbs on board are useless with UAM)

image
(bottom right complication)


olorinmaia:
Testet OK merged with latest dev.

Issue #3692, opened 2025-01-03
[3.3] Config Builder Samsung Tizen


reijoahola-nightscout:
This setting enables data broadcast to several CGM apps serving several types of watches, not only Samsung Tizen.
Known cases are
- G-Watch App (Samsung Tizen watches)
- G-Watch Wear App (all Wear OS 3 and 4 watches, not only Samsung Wear OS)
- GlucoDataHandler app (all Wear OS 3, 4 and 5 watches)
- GlucoDataAuto (AndroidAuto companion app)

I leave it for further discussion what the header and description text should be in this setting.

olorinmaia:
You could try asking AI for suggestions to this 😉 I agree we should find a better name and description

olorinmaia:
Suggested Setting Name:
"Companion App Broadcast"

Description:
Enables data broadcasting to companion apps on Samsung Tizen, Wear OS, and Android Auto devices, including G-Watch, G-Watch Wear, GlucoDataHandler, and GlucoDataAuto.

olorinmaia:
To illustrate:

| Current string | Improved string |
|--------|--------|
| ![image](https://github.com/user-attachments/assets/f11f0472-d703-481b-8911-1f3957d63e06) | ![image](https://github.com/user-attachments/assets/8c9d61e8-f8e2-4b6f-819c-94a98478fb52)|

reijoahola-nightscout:
It is OK for me :)

MilosKozak:
what is added value on android (compared to native AAPS possibilities)?

olorinmaia:
I didn't know about several of these apps (because of old description, I thought this only was for Tizen OS). I've tested them and they add nothing new entirely except for perhaps G-Watch that allows for showing BG graph + complications. Glucodatahandler have a few more options in complications than in-built with AAPS.
![image](https://github.com/user-attachments/assets/96491c6d-4743-43c8-8232-a46dcc707220)

More info:

https://play.google.com/store/apps/details?id=sk.trupici.gwatch.wear&hl=en
https://github.com/pachi81/GlucoDataHandler/blob/master/README.md
https://github.com/pachi81/GlucoDataAuto

reijoahola-nightscout:
> what is added value on android (compared to native AAPS possibilities)?

G-Watch Wear App is just another alternative for analog and digital watchface having CGM graph.

[...]

PR #3696, opened 2025-01-04
Improve strings for tizen plugin

olorinmaia:
For https://github.com/nightscout/AndroidAPS/issues/3692

New Plugin name:
`External Companion Apps`
New Plugin description:
`Enables data broadcasting to various external companion apps`

Current string
image | ![image]


Philoul:
Hum,
If plugin is dedicated to synchronize data with only one single external app, we can use an explicit name and explicit description.

Now that the same plugin is more generic and provide information for several external consumer apps, the generic name is fine, but I think we should also propose a generic description and don't try to be exhaustive in the external apps that can use it (to simplify maintainability of this description later...)

It's more within aaps doc (if we want to provide an exhaustive list in one place), or within external app doc like G-Watch documentation,  ...), to provide this kind of information 🤔.

olorinmaia:
So suggested name of plugin is in. 

Description could be changed to something like:

Enables data broadcasting to various companion apps on Samsung Tizen OS, Wear OS, and Android Auto devices.

Philoul:
I think no "and" at the end of the description which suggest the list is finished and exhaustive 🧐.
So either (without any name)
"Enables data broadcasting to various external  companion apps"

Or (if we think it's better to provide exemples to be a bit more precise), something like:
"Enables data broadcasting to various companion apps like G-Watch for Tizen or Wear OS, several Android Auto devices..."

The idea is if any additional new external apps include this broadcasting management, then the plugin description remains true and don't need to be updated...

olorinmaia:
I agree that it's smart to avoid putting an description that adds for the need to update string all the time.

Something like this could work:
`Enables data broadcasting to various external companion apps on Tizen OS, Wear OS and Android Auto. `

I can update wiki for this plugin to reflect what's actually supported at this moment.
https://androidaps.readthedocs.io/en/latest/Getting-Started/Watches.html#samsung-tizen

Known apps:
* G-Watch App (Samsung Tizen watches)
* G-Watch Wear App (all Wear OS 3 and 4 watches, not only Samsung Wear OS)
* GlucoDataHandler app (all Wear OS 3, 4 and 5 watches)
* GlucoDataAuto (AndroidAuto companion app)

PR #3725, opened 2025-01-09
Wear: add Loop States tile

tdrkDev:
Provides ability to change loop state (switch modes, disconnect/reconnect pump, suspend/resume loop) from WearOS tile.

I find this tile useful in my usage scenario when I put my watch on charge before going to shower. This tile helps to set "pump disconnected" loop mode directly from my watch instead of finding my phone in the room, opening AAPS, loop dialog and only then disconnecting pump.

The implementation has timestamp-based authorisation, so user can't do something unexpected when tile and DataHandlerMobile go out of sync, also all actions have additional confirmation on watch before actually running.


MilosKozak:
anyone tested?

olorinmaia:
Will test it now.

plugins/sync/src/main/res/values-ru-rRU/strings.xml
```diff
@@ -183,4 +183,27 @@
...
+    <string name="wear_action_loop_state_profile_invalid">Неверный профиль!</string>
+    <string name="wear_action_loop_state_unauthorized">Повторите попытку.</string>
+    <string name="wear_action_loop_state_invalid">Запрошен неверный тип цикла</string>
+
+    <string name="wear_action_loop_state_title">СОСТ. ЦИКЛА</string>
+    <string name="wear_action_loop_state_changed">%1$s</string>
+    <string name="wear_action_loop_state_changed_with_duration">%1$s\nДлительность: %2$d мин</string>
+
+    <string name="wear_action_loop_state_now_closed">Закрытый</string>
+    <string name="wear_action_loop_state_now_lgs">LGS</string>
```
olorinmaia, line 199: Since it's plenty of space for text in the wear tile confirmation box I would use `Low glucose suspend` instead of `LGS`

olorinmaia:
Tested OK for AAPS. Will check how it looks for AAPSClient

plugins/sync/src/main/res/values/strings.xml
```diff
@@ -221,4 +221,27 @@
...
+    <string name="wear_action_loop_state_profile_invalid">Profile is invalid!</string>
+    <string name="wear_action_loop_state_unauthorized">Please try again.</string>
+    <string name="wear_action_loop_state_invalid">Invalid loop state requested</string>
+
+    <string name="wear_action_loop_state_title">LOOP STATE</string>
+    <string name="wear_action_loop_state_changed">%1$s</string>
+    <string name="wear_action_loop_state_changed_with_duration">%1$s\nDuration: %2$d min</string>
+
+    <string name="wear_action_loop_state_now_closed">Closed</string>
+    <string name="wear_action_loop_state_now_lgs">LGS</string>
```
olorinmaia, line 236: Since it's plenty of space for text in the wear tile confirmation box I would use `Low glucose suspend` instead of `LGS`

olorinmaia:
Tested OK for AAPSClient, it's like with Automation user action tile, "No configuration", so it can't be used. 

Both User action and Loop state is visible for AAPSClient wear, I'm not sure if it's possible to hide them, but that is something that can be improved upon later also.

Thumbs up from me! :)

PR #3734, opened 2025-01-12
Add Basal / TBR rate complication for wear

olorinmaia:
With latest Wear OS there is multiple watchfaces that got 5+ slots for complications. That give us the possibility to show exactly what we want in each slot. 

There is no complication right now that only show current basal / TBR rate, this will add that.

See screenshots from wear below and how this can be used:

image
image

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -233,6 +233,24 @@
         </service>
 
         <service
+            android:name=".complications.BrComplication"
+            android:exported="true"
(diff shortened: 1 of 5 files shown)


MilosKozak:
The only concern I have is why the user would like to see running BR on watch. It's so detailed info that should be resolved on phone

olorinmaia:
I like to see if a TBR 0 is set when looking at the watch or when activity also seeing positive TBR. There are already BR complications, this is just another option showing only BR

PR #3742, opened 2025-01-15
fix display of negative IOB on wearos complications

keweki:
set MIN_FIELD_LEN_IOB from 3 to 4 for displaying the negative symbol when IOB is less than zero on complications for WearOS


MilosKozak:
I reverted this change as it break tests

Issue #3467, opened 2024-10-14
Equil pump driver


karzyg555:
I'm testing dev version and Equil pump integration. I've noticed 2 potential issues:
- When i try to inject single insulin dose equal or above 10 units i got communication problem between AndroidAPS and pump
- When insulin reservoirs will become empty and piston is going back, AndroidAPS app still shows progress of injection of insulin

MilosKozak:
Issue confirmed

PR #3757, opened 2025-01-18
Eqil: Fixed bug

EquilHack:
1.The communication protocol between AndroidAPS and the Equil pump was updated to correctly handle insulin doses of 10 units or higher. Additionally, a validation check was implemented in AndroidAPS to ensure that doses are within the Equil pump's supported range before initiating the injection.
 2.Fixed When insulin reservoirs will become empty and piston is going back, AndroidAPS app still shows progress of injection of insulin

https://github.com/nightscout/AndroidAPS/issues/3467


karzyg555:
Tested fix in AAPS version 3.3.1.3 and problem still exists. One difference in the behaviour is longer time before error is returned in the app.

PR #3764, opened 2025-01-19
Wear: allow negative carb entry from Carb + Treatment

olorinmaia:
Tested OK locally on my test rig using Wear OS 5

### DIFF
--- a/wear/src/main/kotlin/app/aaps/wear/interaction/utils/PlusMinusEditText.kt
+++ b/wear/src/main/kotlin/app/aaps/wear/interaction/utils/PlusMinusEditText.kt
@@ -103,7 +103,6 @@ class PlusMinusEditText @JvmOverloads constructor(
                 value = maxValue
             } else {
                 value = minValue
-                stopUpdating()
             }
         }
         updateEditText()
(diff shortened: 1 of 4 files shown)


wear/src/main/kotlin/app/aaps/wear/interaction/utils/PlusMinusEditText.kt
```diff
@@ -103,7 +103,6 @@ class PlusMinusEditText @JvmOverloads constructor(
                 value = maxValue
             } else {
                 value = minValue
-                stopUpdating()
```
MilosKozak, line 106: why this?

olorinmaia: I couldn't go below 0 when tapping on - (minus) when trying to add negative carbs without removing it.

olorinmaia: Input is still restricted to maxCarbs both positive and negative direction

olorinmaia: Thanks for feedback. I'll see if I can make it work without removing it. I notice now that it has that unwanted effect you mention.

PR #3802, opened 2025-02-02
Improve Sensor Change strings for BG Source Settings

olorinmaia:
This will clarify what this option really does, and hopefully not cause issues like below, which isn't the first time.
image

image

### DIFF
--- a/plugins/source/src/main/kotlin/app/aaps/plugins/source/AbstractBgSourcePlugin.kt
+++ b/plugins/source/src/main/kotlin/app/aaps/plugins/source/AbstractBgSourcePlugin.kt
@@ -24,7 +24,7 @@ abstract class AbstractBgSourcePlugin(
         parent.addPreference(category)
         category.apply {
             key = "bg_source_upload_settings"
-            title = rh.gs(R.string.bgsource_upload)
+            title = rh.gs(R.string.bgsource_settings)
             initialExpandedChildrenCount = 0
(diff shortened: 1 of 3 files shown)


chrisbev:
This is excellent for my use/problem.  I am not familiar with nightscout and the original log event "Sensor Change" to nightscout previous text to know if it needs retaining/including.

MilosKozak:
"Sensor change" is the name of event in NS. I'm afraid using different name in AAPS could be confusing

olorinmaia:
In AAPS UI, Treatments > Careportal and Nightscout +Button the event is named "CGM Sensor Insert". (Which is why I went for that phrase)
image
image

I'm not sure if it's worth mentioning NS in this setting, as it's purpose is mainly to register CGM Sensor Insert to AAPS Treatments Careportal based on received data from BG source local broadcast, which in turn will upload to NS if NS Sync is configured.

olorinmaia:
I welcome suggestions to better phrasing ofcourse! :) At the moment I can't see any clearer description.

olorinmaia:
If "CGM Sensor Insert" treatment to Careportal is uploaded to NS or not is controlled by a setting already exisiting in NSClientv1/v3 > Sync options > Upload data to NS.

MilosKozak:
`{"_id":"67a1e4afc134b0fee2de0c63","enteredBy":"undefined","eventType":"Sensor Change","created_at":"2025-02-04T09:58:05.534Z","utcOffset":0,"mills":1738663085534,"carbs":null,"insulin":null}`
right DB record uses "Sensor Change" but UI "CGM Sensor Insert"

Issue #3801, opened 2025-02-01
Spammy User Entry Care portal (CGM Sensor insert)


piecycle:
Since a couple of releases, the user entry tab in the Treatments section logs the same message every 5 minutes (see screenshot).

![Image](https://github.com/user-attachments/assets/af20895f-1aa3-4fab-b527-a660e1848c13)

I'm using a Dexcom G7 with xDrip as collector and data source in AAPS. The screenshot was taken from AAPS 3.3.1.3 (latest Dev from today, 01/02/2025), running the tsunami mod, but the issue appears to be unrelated to the mod.

olorinmaia:
Also see https://github.com/NightscoutFoundation/xDrip/discussions/3508 for possible solutions. Something has to be done, either on xDrip side or AAPS side, there is nothing decided yet.

reijoahola-nightscout:
I checked one earlier CGM Sensort Spam list. I am sure that earlier it was quite small, may ne 10 lines. Now it was over 110 lines ! Spamming seems to continue and increase the list.

vanelsberg:
> logs the same message every 5 minutes 

Current dev, using BYODA; not seeing this. So I'd say this is a xDrip+ problem only.

olorinmaia:
If using xDrip+ and G7 and experiencing this problem, please see https://github.com/nightscout/AndroidAPS/pull/3804 and test if you want. It fixed the issue for me. I haven't changed G7 session yet with this fix so that is still to confirm.

olorinmaia:
Fix is now merged to dev branch

olorinmaia:
Can confirm that new sensor start is received when changing G7. Still no spam. @MilosKozak I think this issue can be closed then.

olorinmaia:
@piecycle @MilosKozak This issue can be closed due to ce20cccf - tested OK on several sensor changes.

PR #3804, opened 2025-02-03
Fix "CGM Sensor Insert" spam in Careportal from G7 xDrip local broadcast

olorinmaia:
My proposal to fix issue with described here: https://github.com/NightscoutFoundation/xDrip/discussions/3508 and here: https://github.com/nightscout/AndroidAPS/issues/3801
 Also fixes issue with NS reports containing alot of unnecessary sensor changes.
image

Sensor start time is accepted only if changed more than 5 minutes to avoid a very long list of "CGM Sensor Insert" in NS and Careportal for G7 using xDrip+. 
 
This mainly seem to affect NSClientV3. (For testers). I've updated my live rig with G7, and problem is fixed for me, but I haven't tested with changing G7 yet.

### DIFF
--- a/plugins/source/src/main/kotlin/app/aaps/plugins/source/XdripSourcePlugin.kt
+++ b/plugins/source/src/main/kotlin/app/aaps/plugins/source/XdripSourcePlugin.kt
@@ -123,7 +123,7 @@ class XdripSourcePlugin @Inject constructor(
                 newSensorStartTime != null &&
                 abs(newSensorStartTime - lastStoredSensorStartTime) <= 300_000) {
                 aapsLogger.debug(LTag.BGSOURCE, "Sensor start time is within 5 minutes range, skipping update.")
-                lastStoredSensorStartTime
+                null
                 } else {
                     newSensorStartTime
                 }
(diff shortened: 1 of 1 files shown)


plugins/source/src/main/kotlin/app/aaps/plugins/source/XdripSourcePlugin.kt
```diff
@@ -113,8 +114,21 @@ class XdripSourcePlugin @Inject constructor(
...
+            val newSensorStartTime = getSensorStartTime(bundle)
+            // Retrieve last stored sensorStartTime from the database
+            val lastTherapyEvent = persistenceLayer.getLastTherapyRecordUpToNow(TE.Type.SENSOR_CHANGE)
+            val lastStoredSensorStartTime = lastTherapyEvent?.timestamp
+            // Decide whether to update sensorStartTime or keep the last stored one
+            val finalSensorStartTime = if (lastStoredSensorStartTime != null &&
+                newSensorStartTime != null &&
+                abs(newSensorStartTime - lastStoredSensorStartTime) <= 300_000) {
+                aapsLogger.debug(LTag.BGSOURCE, "Sensor start time is within 5 minutes range, skipping update.")
+                lastStoredSensorStartTime
```
MilosKozak, line 126: putting `null` here will save another DB query on storing

olorinmaia: dd790d5d

PR #3808, opened 2025-02-05
Wear: Add Duration support to Profile Switch and add it to Action Tile

olorinmaia:
* Now you can use Profile Switch on Wear to set duration. 
* Added Profile Switch to Action Tile
* Added constants for other parameters.

Tested OK locally on my test-rig. Run all tests in Android studios OK.

Some screenshots:
image
image
image


olorinmaia:
Cherry picked on my master branch and tested ok with aapsclient wear also + aaps wear.

PR #3812, opened 2025-02-07
Unbond Pod after deactivation

2flea:
Omnipod Dash unbonding

The current call in "DashDeactivatePodViewModel.kt" in function "discardPod" does not work. 
One rather needs to implement the call of "removeBond()" in "OmnipodDashManagerImpl.kt" and add 

".doOnComplete { bleManager.removeBond() }"

to Pod event "observeSendDeactivateCommand"


MilosKozak:
what's the difference?

2flea:
> what's the difference?

With the other/previous code the unbonding does not work or rather it is not called.

PR #3818, opened 2025-02-09
Start timer alarm after successful bolus.

mushroom-dev:
- no need to cancel alarm after bolus error
- my pump (AC Combo) sometimes connect immediately, sometimes it can take up to 3 minutes, if that happens time to meal is to short. I am not sure how many pumps have similar issue


kagilgin:
I would like to request that this update be changed to optional. Those who want it should run the timer at the time of the order to send insulin, those who want it should run the timer when the insulin delivery is completed. Also by default the bolus wizard already assigns the carbohydrate time according to the start of the insulin order, not according to the alarm. I think setting the timer at the start of insulin delivery is better for managing my sugar. @MilosKozak

danwhill:
I would second the request to make the timing optional (start vs end of bolus)  Starting it after insuline delivery is completed can delay the time of the user eating by 4 or 5 minutes for a large bolus, and could create a risk of a low blood sugar.  Pre-bolus timing is important to many of us.  @MilosKozak

Issue #3798, opened 2025-01-31
[Bug] - Bolus Wizard does not correctly display negative units


pnharris02:
The Bolus Wizard/Calculator now shows a positive number for IOB when you have positive IOB. In release 3.2.0.4, it was a negative number, which was logical because it had to be deducted from the calculated dose.

This was brought up in Discord by @barisk81 and discussed by myself and @robby
https://discord.com/channels/629952586895851530/1199783453793734677/1334593953734983743

I suggest, to reduce confusion, the actual IOB is shown in the horiz centre of the dialog, above the COB figures, then the inverse/negative number is shown in the right column, similar to v3.2

A second visual bug on this screen is that any negative number in the right column causes the "U" after the figure to wrap to the next line.  And in the case of IOB, the bottom of the U is also truncated slightly.

The attached screenshot pairs are from two different times, one in September to show 3.2 behaviour and one from this morning on 3.3.1.3, with a rough sketch of my suggested resolution

![Image](https://github.com/user-attachments/assets/76bd8b82-d816-4046-8522-25b46960b641)

![Image](https://github.com/user-attachments/assets/44f7e831-6c1c-4a73-a3bb-840854566fed)

![Image](https://github.com/user-attachments/assets/eaa74186-0dfd-4ff6-811d-cb94c866cdf8)

![Image](https://github.com/user-attachments/assets/8a6d7efa-adbb-4acf-b271-903ffed3d458)

robertrub:
Yes, it would be logical to keep the same math notation for all the numbers. If it is added to the bolus calculation, it should be + (or no sign) but if it will be subtracted, it should have the - sign in front.

olorinmaia:
Fix https://github.com/nightscout/AndroidAPS/pull/3831

Is merged to dev for those who want to test

pnharris02:
Thanks @olorinmaia. I'll download the latest dev and check it out when I get a chance.
I'm inferring you didn't like my idea of displaying the actual IOB (the inverse of what is shown in the calculations column) in the middle column of the page?
But that's okay if you've changed the sign. Thank you for fixing this.

olorinmaia:
@pnharris02 Your welcome :) I focused on getting correct math notation in wizard calc log so when you sum them all up it will be the bolus you will administer in the end. I think it's sufficient to show the IOB amount at the right when you use bolus calculator. It's not a bad idea, but there is always something about adding more text to the screen, it need to be really necessary imo if added.

[...]

PR #3831, opened 2025-02-15
Fix issues with math notation for IOB in wizard dialog/result and calculation of COB vs IOB

olorinmaia:
My proposal to fix issue mentioned in https://github.com/nightscout/AndroidAPS/issues/3798 and some more issues I found while looking.

* Fix issues with wrong math notation for IOB in Wizard/QW dialog in phone and wear

* Fix issue with wrong math notation for IOB in Wizard/QW Calc log in treatments 

* Fix issue with wrong calculation of COB vs IOB in wizard confirmation dialog box (One could argue to remove insulinFromBG from the equation, to compare solely COB vs IOB.?)

From my testing with both positive and negative IOB, the math for the sum of calculations is equal to the bolus amount in the end.

Comparisons:
| Before | After |
|--------|--------|
| imageimage| imageimage|
| Before | After (forgot 100% for comparison) |
| image | image|
| Before | After |
| imageimage | imageimage | 
| Before | After |
| image|image|

### DIFF
--- a/core/objects/src/main/kotlin/app/aaps/core/objects/wizard/BolusWizard.kt
+++ b/core/objects/src/main/kotlin/app/aaps/core/objects/wizard/BolusWizard.kt
@@ -355,7 +355,7 @@ class BolusWizard @Inject constructor(
             actions.add(
                 rh.gs(app.aaps.core.ui.R.string.cobvsiob) + ": " + rh.gs(
                     app.aaps.core.ui.R.string.formatsignedinsulinunits,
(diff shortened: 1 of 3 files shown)


pnharris02:
Thanks @olorinmaia. I'll download the latest dev and check it out when I get a chance.
I'm inferring you didn't like my idea of displaying the actual IOB (the inverse of what is shown in the calculations column) in the middle column of the page?
But that's okay if you've changed the sign.  Thank you for fixing this.

PR #3833, opened 2025-02-17
Improve UI of wizardinfo (calc log in treatments)

olorinmaia:
See pictures below.

| Before | After |
|--------|--------|
| image| image|


olorinmaia:
I'll open another PR with another way to fix this, but keep the height between rows.

Issue #3836, opened 2025-02-18
[Wear] Display/Rounding error when setting preset temp target from AAPS Wear


pnharris02:
There appears to be a rounding/display error when setting a preset temp target from the watch.
It only appears with some mmol values (eg. 9.8 and 9.6 but not the other numbers surrounding this) so I wonder if perhaps there is some sort of rounding bug when/if it is converted to mg/dL and back again??
See screenshot

![Image](https://github.com/user-attachments/assets/e9e89b34-27c7-496d-9a5e-6f616d81573e)

Also, while you're there, it would probably be good if the title was changed from "Temptarget:" to "Temp-Target" or something similar

I'm guessing this might fall in your area of expertise @Philoul

olorinmaia:
My proposal to fix this, works fine on my end: https://github.com/nightscout/AndroidAPS/pull/3838

pnharris02:
Looks good to me.  Thanks @olorinmaia

Philoul:
Solution proposed is fine.
Just little comment, I prefer "Temp-Target" shorter than "Temporary Target".
Keep in mind that a lot of langages use longer words compared to english (german, french, ...), and translators will generally translate long words by long words, shorter words by shorter words, without knowing where the string is used.
Short version or acronyme (Temp-Target, TT) is well known by all users because use everywhere in AAPS app and in doc 😉.

olorinmaia:
@Philoul Good point, updated PR.

olorinmaia:
@MilosKozak This can be closed due to https://github.com/nightscout/AndroidAPS/pull/3838

PR #3838, opened 2025-02-18
Improve and unify strings for TempTarget in Wear and add check for invalid TT range

olorinmaia:
Proposal to fix https://github.com/nightscout/AndroidAPS/issues/3836 and:

* Improve and unify strings for TempTarget on wear, add mg/dl or mmol/l and min to duration to strings
* Add check for invalid temp target range
* Fix default values for TempTarget range
* Tested and verified all TT functions for mmol/l and mg/dl on wear vs phone app still works like before

mmol/l TempTarget:
| Before | After |
|--------|--------|
| image| image| 

mg/dl TempTarget:
| Before | After |
|--------|--------|
|image|image|

low - high range TempTarget:
| Before | After |
|--------|--------|
|image|image|

Error if invalid TempTarget range:
image


MilosKozak:
* I'd unify TempTarget vs Temp-Target. I think TempTarget is mostly used in AAPS
* you cannot change %s to %f without adjusting the code

olorinmaia:
Can use TempTarget instead if you prefer.

I'm abit confused. It's working fine. It's the presentation in UI that's abit messy? I don't know how to fix what you suggest in bullet point two.

Do you think TT values are stored incorrectly went sent from watch? If so I'll close this one and wait for a better solution.

Philoul:
TT parameter sent within these strings is a double (so the %s was probably an error which didn't produce crash but just an automatic convertion from double to string)
It can probably be improved using 2 differents strings according to units and prevent a "xxx.0" value for mgdl users (`%2$s` replaced by `%2$.0f` or `%2$.1f` according to unit), or by managing the rounding and conversion to string by code 🤔

The 2 other TT strings should also be updated to be consistant...

olorinmaia:
@Philoul @MilosKozak What about something like this?

```
            EventData.ActionTempTargetPreCheck.TempTargetCommand.PRESET_EATING   -> {
                val eatingSoonTTDuration = preferences.get(IntKey.OverviewEatingSoonDuration)
                val eatingSoonTT = preferences.get(UnitDoubleKey.OverviewEatingSoonTarget)
                val formattedGlucoseValue = if (presetIsMGDL)
                    String.format(Locale.getDefault(), "%.0f mg/dL", eatingSoonTT)
                else
                    String.format(Locale.getDefault(), "%.1f mmol/L", eatingSoonTT)
                val reason = rh.gs(app.aaps.core.ui.R.string.eatingsoon)
                message += rh.gs(R.string.wear_action_tempt_preset_message, reason, formattedGlucoseValue, eatingSoonTTDuration)
                rxBus.send(
                    EventMobileToWear(
                        EventData.ConfirmAction(
                            title, message,
                            returnCommand = EventData.ActionTempTargetConfirmed(presetIsMGDL, eatingSoonTTDuration, eatingSoonTT, eatingSoonTT)
                        )
                    )
                )
            }
```

string = `TempTarget Reason: %1$s Target: %2$s Duration: %3$s min`

How it looks:
| mg/dl| mmol/l |
|--------|--------|
| image| image|

MilosKozak:
action.duration is not string

olorinmaia:
> action.duration is not string

73d469f

PR #3853, opened 2025-02-26
Insulin Management within Profile

Philoul:
This PR is a first step to replace dia management within profile (deprecated in profile interface) by a global insulin management within profile:
- ICfg is fully included within ProfileSealed, PureProfile, SingleProfile, json import/export...
- Profile migration (using profile's dia and Current selected InsulinPlugin for peak value) has been included within MainApp (but this migration should be completed in the next step, see below)
- Dialogs (ProfileViewer) and ProfilePlugin has been updated to show Insulin Name, peak and dia
- Autotune plugin has also been updated to use Profile insulin during run.
- All unit test has been updated and passed

Within this first step when a profileswitch is done, it will use Insulin Recorded within profile (dia and peak), so it will no more use current InsulinPlugin peak value.
A clone Profile will also clone insulin included into source profile (this will be improved on next step)
A new profile will use CurrentInsulin (concerning peak)
UI concerning dia value is still within ProfilePlugin (will be moved to InsulinPlugin in next step).

Next steps I have in mind:
- Replace the 4 current InsulinPlugins by one unique InsulinPlugin (and finish Migration code within MainApp). This plugin will allow the user to manage his single or various insulins (pump, pens)
- Update ProfilePlugin to replace DIA +/- User Interface by Insulin selection menu
- Include Insulin selection for external bolus
- ...


Philoul:
Note, this PR will replace draft PR #3782 
- I managed several tests, several updates of code back and forth between this branch and  latest dev branch. It looks stable on my side
- I still let the draft PR #3782 opened to have in mind the target solution concerning future proposed UI for Insulin Management

olorinmaia:
I will try to help test this PR by adding it to my test rig. Virtual pump though but still useful?

olorinmaia:
I have setup AAPS test-rig with Ultra-Rapid Oref.

If I change DIA on AAPSClient and save, the following happen:
Peak changes from 55 -> 75 min
Insulin changes from Ultra-Rapid Oref to Rapid-Acting Oref
Profile switching from AAPSClient after changing DIA then seem to set wrong insulin type / peak on AAPS.

Philoul:
> If I change DIA on AAPSClient and save, the following happen:

Did you install this PR on both sides (master loop and AAPSClient)?
- insulin management is different so update should currently be done on both sides (but good catch, I will have check AAPSClient sync in both directlions, and probably improve management...)
- profile's insulin is currently not managed within old insulin plugin

This PR is only a first step, (working locally), but I didn't test NSClient synchro.
- in target solution you can't change dia from profile, you change insulin

Philoul:
A new item has been added in my todo list, Check AAPS <=> Syncho (insulin sync and profile sync, with same or different versions...).

olorinmaia:
I built AAPSClient with same branch👍

Since AAPSClient can't select which insulintype in config builder, it could be that it reverts to some default value when changing DIA and saving, but it's sending both DIA and INS type to AAPS, since AAPS is updated. So there is probably some choice we need to make here. Do we want AAPSClient to be able to change INS also? I don't see a problem with that, but it would make sense if it was part of the actual profile dialog then perhaps instead of config builder, but that was next step right?

Philoul:
> Do we want AAPSClient to be able to change INS also?

Good question, to be discussed...
In my mind changing insulin type (replace one insulin by another insulin) can only be done in AAPS, and should normaly be synchronized with a reservoir/pod change (replace insulin within pump by another one).
- So this kind of change should normaly be limited to AAPS app only

The other modification can be update of an existing insulin (without changing insulin within pump).
- this should be done from insulin plugin only (to adjust peak/dia) with a modification of one existing insulin. 
- Then all profiles using the modified insulin should be updated, with new values automatically (one confirmation popup for all), when updated insulin is saved within insulinPlugin
Again, this should probably not be available from NSClient with insulinPlugin hidden.

Finally, a ProfileSwitches managed within AAPS UI (Overview, Activity, Automation, SMS...) should only be available between profiles sharing the same insulin.
- To be able to select a profile with a different insulin compared to insulin included into current running profile, you should manage PS from ProfilePlugin (with a dedicated warning/confirmation if insulin included into new selected profile is different...)

That's why I included within future InsulinPlugin a tab to list and show the profiles using selected insulin 😉.

Again these are my thoughts, but of course opened to discuss on UI, features...

Philoul:
@olorinmaia I now included the proposed InsulinPlugin (unique) to manage all type of insulin.
So you select insulin within profilePlugin (no more DIA setting within ProfilePlugin)
Be carefull, I didn't remove previous 4 insulinPlugin, so you should select manually InsulinPlugin to test everything
I'm not sure how AAPSClient will work (able or not to select an insulin...)
Tell me if you see strange things on AAPS, and other strange things in AAPSClient...

olorinmaia:
@Philoul I've tested your latest commits, and it seem to be working fine, with some minor issues with deleting / saving insulin profiles. I really like this approach that you can have multiple insulin profiles and can change this in your profile. We however use only Fiasp, but I understand the need for it. It simplify UX abit also, by being able to set all this in profile.

I'm wondering though, if we should hide the possibility to change Insulin Profile in AAPSClient, and only show What InsType, Peak, DIA is set for each profile when activated. If this is to be changed, it should be done on AAPS. Today in AAPSClient you can change DIA, but not insulintype or peak. One could say we take something from AAPSClient, but we also give more information as you now can see ins type, dia and peak when you open profile.

Problem is that changes to Insulin profiles seem to sync very slow with AAPSClient, which could lead to issues. It seem to sync whenever AAPS is restarted. If that could be improved I see less issues with Insulin Profiles being available to change on AAPSClient.

You can try to add a few new Insulin Profiles on AAPS and delete a few, and then activate profile. And notice that on AAPSClient, you won't see it, until you restart AAPS and wait some.

Philoul:
@olorinmaia With my latest commits:
- the 4 previous InsulinPlugins has been removed (only remain one unique InsulinPlugin)
- in AAPSClient versions, Insulin type cannot be changed within Profile Plugin (TextView)
Can you verify if now synchronization is better or not between AAPS and AAPSClient ?

olorinmaia:
Hi @Philoul :) I've tested some more and sync seem to be better now as you got less chance of breaking stuff when you can't change the INS parameters in AAPSClient.

I think I've found a few bugs, will try to help you recreate them.

1) If you modify a Insulin that is used in a Profile and save. Then select another Insulin and back to the one you just saved, you will see that it got "0" Profiles. Only way to get it back is to restart AAPS.

2) When creating a new Insulin the UX is a bit messy if the one selected before you click "+" to create got a connected Profile.

| First you can see that the "new" insulin is used in 1 profile | When saving new Insulin you get this pop-up |
|--------|--------|
| image| image | 

3) If you change which Insulin a profile uses. The Insulin-plugin doesn't show the correct number of profiles the insulin is being used in. It doesn't help to activate profile either. Must restart AAPS for it to count correctly.
image
image
image

4) Not sure exactly how I got this problem, but I think I edited a free peak insulin peak time, then a new Insulin was created with a "automated" name which I didn't set. And now I'm not able to delete it. See video below.

https://github.com/user-attachments/assets/81b72f74-40a2-4491-8199-55dc6c5181a3

Just some feedback from my testing, could be my rig that got some issues with DB also. But I hope it will be of help improving UI/UX for this new feature.

Philoul:
@olorinmaia Thx for comment
I will rework UX to simplify code and fix the bugs...

The expexted behaviour was:
- when you edit an insulin used in one or more profiles, you get the confirmation popup. (but this confirmation popup shoud not be shown on a new insulin). If you confirm all profiles using selected insulin should be updated.
- when you create a new insulin (button +), number of profile should be 0 (it seems you have one profile so a bug which can explain why you got the popup...).
- red background should occurs (with toast message) only when you cannot save the edited insulin (empty name, name already used by another insulin, or edited insulin already exists with same peak and same dia).

Philoul:
@olorinmaia UI bugs should be fixed now (I hope...)
additional UI updates:
- if an insulin is used on one or several profiles, the remove button is hidden (an insulin can be removed only if it is not used nowhere)
- Current Active profile is now included in the list of profiles on the top (<Active>).

Philoul:
Hum, my working branch was ok in my emulator. but it crashed after merge with latest dev...

Philoul:
Crash fixed with my latest commits (due to wrong preferences migration).
- after migration from 3.3.2.0 it can be necessary to select profiles once within Profile plugin to see dedicated insulin (if dia is different than current active profile's dia...)
- Several fixes also done within UI

Philoul:
Replaced by #3952

Issue #3821, opened 2025-02-11
[Bug] High Battery Usage Due to Frequent WakeLocks in Wear OS Complications


TebbeUbben:
I have noticed that using AndroidAPS complications on my Wear OS watch significantly increases battery consumption - reducing battery life by _**several hours**_. Upon analyzing battery logs, I found that AndroidAPS acquires a WakeLock every 15 seconds, which prevents the device from entering a more power-efficient sleep mode.

The issue originates from [this line](https://github.com/nightscout/AndroidAPS/blob/dece9288370c7f54b3d20456173789edf9d94afb/wear/src/main/kotlin/app/aaps/wear/complications/BaseComplicationProviderService.kt#L263) in `BaseComplicationProviderService`. The intention behind this logic is to ensure complications update correctly when displaying the time elapsed since the last CGM reading. However, this approach keeps the CPU awake unnecessarily, leading to excessive power drain.

A more battery-efficient way to achieve the same behavior is to use `ComplicationText.TimeDifferenceBuilder`. This allows the system to intelligently update the complication only when needed, reducing unnecessary wake-ups and code executions and allowing the watch to stay in a low-power state when not in use.

I’ve implemented this approach for the `LongStatusComplication`, and it has worked well in my testing. My changes can be found in the [wear-complications-timedifference](https://github.com/nightscout/AndroidAPS/commit/b65d3e4660d0cdcd24af2cbc06852e9b6e763af3) branch.

Additionally, newer Wear OS versions and Jetpack libraries introduce [dynamic expressions](https://developer.android.com/training/wearables/data/dynamic), which provide even more flexibility for complication updates while optimizing battery usage.

MilosKozak:
@Philoul

olorinmaia:
@MilosKozak @TebbeUbben  I think this issue has been resolved by 7b445d7d43 and other improvements being made. It can be closed in my opinion after https://github.com/nightscout/AndroidAPS/pull/5039 is eventually merged, which is the last two complications not using system-driven auto-updating age (TimeDifferenceComplicationText)

olorinmaia:
https://github.com/nightscout/AndroidAPS/pull/5039 merged - if you agree, this issue can now be closed @TebbeUbben @MilosKozak

PR #3873, opened 2025-03-06
Complication Update

Philoul:
Test to improve consumption

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -157,7 +157,7 @@
                 android:value="LONG_TEXT" />
             <meta-data
                 android:name="android.support.wearable.complications.UPDATE_PERIOD_SECONDS"
-                android:value="300" />
+                android:value="1800" />
         </service>
 
         <service
--- a/wear/src/main/kotlin/app/aaps/wear/comm/DataHandlerWear.kt
+++ b/wear/src/main/kotlin/app/aaps/wear/comm/DataHandlerWear.kt
@@ -4,11 +4,13 @@ import android.app.Notification
 import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.app.PendingIntent
+import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
 import android.os.Build
 import android.os.Bundle
 import android.os.SystemClock
+import android.support.wearable.complications.ProviderUpdateRequester
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
 import androidx.localbroadcastmanager.content.LocalBroadcastManager
@@ -26,6 +28,8 @@ import app.aaps.core.keys.DoubleKey
 import app.aaps.core.keys.IntKey
 import app.aaps.core.keys.Preferences
 import app.aaps.wear.R
+import app.aaps.wear.complications.BaseComplicationProviderService
+import app.aaps.wear.complications.LongStatusComplication
 import app.aaps.wear.interaction.WatchfaceConfigurationActivity
 import app.aaps.wear.interaction.actions.AcceptActivity
 import app.aaps.wear.interaction.actions.ProfileSwitchActivity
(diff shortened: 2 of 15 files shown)


Philoul:
@TebbeUbben can you please review this PR ? (I tried to implement your proposal to fix #3821 for Complication management, but I never use complication on my side and never worked before on complication (so not sure I correctly implemented your proposal...)

@olorinmaia Can you please test this PR and check if it works correctly or if there are issues compare to previous implementation.

On my side I made quick verification (but don't know if it's regression or not...)
- Update of complication is done every 5 minutes (but can be late compare to real update of data). As example for BG Value, time ago is updated every 5 minutes and not every minutes (so you can see BG value with 1' when value has been received 2 or 3 min ago), same for IOB (this could probably be fixed by an adjusted duration of these complication...)
- If watch is always on, complication is not updated in Ambient mode (update is only done when watch is again in Interactive mode)
- Not sure if code works correctly for Uploader battery and Graphical complication (it seems good but did check for long duration...)

olorinmaia:
I've tested PR one day and compared to another day with latest dev, and I can't see any difference in battery consumption either for watch or phone. I don't have good tools to monitor this, except for what comes with android OS. Complications works like before.

MilosKozak:
any update on this?

Philoul:
I tried to implement Tebbe's proposal (not a specialist of complication 😉) but didn't found noticeable improvements concerning watch autonomy...

I didn't made any new version or update since this PR.

njmo:
my sons battery also drains very fast. Android AAPS used since morning about 47% ( phone working for 7hours), we are struggling to have phone charged whole day,  we are using 3.3.2. I don't recall having those problem earlier. And it is not everyday, some days it works whole day and we have 40% late evening, but sometimes in the middle of the day it is dead. Always AAPS have 50%+ drain in battery info app

Philoul:
@njmo 
This PR is dedicated for users with complication (so not using AAPS Watchfaces).
Can you test this PR and check if watch autonomy is improved?

Note that can build and install only the wear apk and keep 3.3.2.0 running in your phone during your tests.

njmo:
we are not using any wearable application for aaps. samsung watch is connected to the phone but only for notification purposes

Philoul:
Will update soon this PR with the 4 additional complications for Ext data...
It seems there is an autonomy improvement with it (one user tested it in discord...)

TebbeUbben:
> Will update soon this PR with the 4 additional complications for Ext data... It seems there is an autonomy improvement with it (one user tested it in discord...)

Hi @Philoul! I appreciate your work on this. Sorry for not being of further assistance as I'm currently a little busy with my degree and private life. I just wanted to add that since updating my Pixel Watch 2 to WearOS 6 / Android 16, the watchface doesn't format the complication correctly anymore (`hh:mm` instead of `x min`). It might be required to use [dynamic values](https://developer.android.com/training/wearables/complications/exposing-data#providing-dynamic-values) here instead.

screenshot-2025-10-20-14-33-43

Philoul:
@TebbeUbben Strange, 
Which complication do you use in your screenshot ? (It seems to be LongStatusComplication but when I look at code behind, it shouldn't be (0:02) but (2') https://github.com/nightscout/AndroidAPS/blob/937e2bccc9c43c1a1566a8ec2793bcae4ff11dba/wear/src/main/kotlin/app/aaps/wear/interaction/utils/DisplayFormat.kt#L41
Could you also share me the watchface used (I don't have Watchfaces with TYPE_LONG_TEXT complication available...) to let me test it on my watch ?

I would appreciate a quick review of my implementation of your proposal in this PR (I'm not a specialist of complication, so I tried to understand and adapt your code, but not sure I did it the right way...)

TebbeUbben:
> @TebbeUbben Strange, Which complication do you use in your screenshot ? (It seems to be LongStatusComplication but when I look at code behind, it shouldn't be (0:02) but (2')
> 
> https://github.com/nightscout/AndroidAPS/blob/937e2bccc9c43c1a1566a8ec2793bcae4ff11dba/wear/src/main/kotlin/app/aaps/wear/interaction/utils/DisplayFormat.kt#L41
> 
> 
> Could you also share me the watchface used (I don't have Watchfaces with TYPE_LONG_TEXT complication available...) to let me test it on my watch ?
> I would appreciate a quick review of my implementation of your proposal in this PR (I'm not a specialist of complication, so I tried to understand and adapt your code, but not sure I did it the right way...)

As far as I can tell, your PR correctly implements the TimeDifferenceComplicationText. And you’re absolutely right: I’m using the LongStatusComplication. The reason you don’t see the faulty formatting in the AndroidAPS code is that I’ve modified it to use a TimeDifferenceComplicationText.

In the standard AndroidAPS setup, the app acquires a WakeLock every 15 seconds to update the complication, which keeps the watch from going to sleep. The idea behind using TimeDifferenceComplicationText was to let the watch face handle the time difference calculation and formatting on demand, rather than forcing constant updates that drain the battery. Please see ##3821.

However, it seems that Wear OS 6 broke the formatting for this complication type. So, we may need to switch to the newer and more flexible “dynamic text” complications instead.

Philoul:
PR updated with the 4 additional complications for External data.

Philoul:
PR probably outdated due to Android x library update and important refactoring of complications...

PR #3879, opened 2025-03-12
Fix "version committed" text visibility to follow AAPS color theme

vanelsberg:


ga-zelle (commented):
works fine for me, commited and uncommitted
Thanks for persisting.

PR #3884, opened 2025-03-16
Mark SMBs as 'automated' when uploading to Tidepool

ConstantinMatheis:
### Motivation
Currently, the Tidepool plugin uploads all boluses with `"subType": "normal"`. This leads to confusion when viewing the data on https://app.tidepool.org/ or when generating the Tidepool PDF report, since it cannot be distinguished if the bolus was administered manually, or automatically by the loop.

This PR proposes to upload boluses with `Type.SMB` as `"subType": "automated"` (see Tidepool documentation [here](https://tidepool.stoplight.io/docs/tidepool-api/5d4b8be93bf4f-automated-bolus-automated)).

### Testing the fix
The PDF report now shows manual boluses as a black bar, and SMBs as grey bars:
Screenshot from 2025-03-16 17-30-16

When viewing directly on https://app.tidepool.org/, the SMBs have a different color and contain the label `Automated` when hovering over it:
Screenshot from 2025-03-16 17-20-53
Screenshot from 2025-03-16 17-20-50


ConstantinMatheis:
Not sure why the build fails immediately, should I create the PR in a different way?

PR #3898, opened 2025-03-20
fix: Connection to RileyLink sometimes can become broken

### LINKED ISSUE #3897: Connection to RileyLink sometimes can become broken with no automatic recovery till AAPS restart
For both OrangePro and MicroLink devices and 3.2.0.4 and 3.3.1.3 AAPS versions sometimes RileyLink and AAPS connectivity is broken and only restart of the AAPS helps to recover from this situation. This is a very annoying issue for many months. So recently I decided to check the code and try to find the root cause of the issue using AAPS logs.

I took last example when AAPS couldn't communicate with RyleyLink for whole night but restart of AAPS solved the issue immeditely. I dive into the logs of the AndroidAPS and find interesting behavior.

I can see that after communication is broken the queue size of RFSpyReader.mDataQueue is constantly growing and that non-zero size looks like a root cause of stuck.

Let's follow one call of RFSpy.writeToData method work when RFSpyReader.mDataQueue is not empty:
1) we actually calling writeToDataRaw method;
2) inside writeToDataRaw we first try to empty reader queue, otherwise we will probably get response not for our command but for some old one:
```
        // FIXME drain read queue?
        byte[] junkInBuffer = reader.poll(0);

        while (junkInBuffer != null) {
            aapsLogger.warn(LTag.PUMPBTCOMM, ThreadUtil.sig() + "writeToData: draining read queue, found this: "
                    + ByteUtil.INSTANCE.shortHexString(junkInBuffer));
            junkInBuffer = reader.poll(0);
        }
```
But the main problem is that inside RFSpyReader.poll method we are doing nothing in case queue is not empty so we never get into while loop:
```
   fun poll(timeout_ms: Int): ByteArray? {
        aapsLogger.debug(LTag.PUMPBTCOMM, "${ThreadUtil.sig()}Entering poll at t==${SystemClock.uptimeMillis()}, timeout is $timeout_ms mDataQueue size is ${mDataQueue.size}")
        if (mDataQueue.isEmpty()) {
            ...main logic here...
        }
        return null
```
So that is why our attempt to drain the queue is not working.
3) after that inside writeToDataRaw we pushing command (rileyLinkBle.writeCharacteristicBlocking) and then listening to the result of the command:
```
        return reader.poll(responseTimeout_ms);
```
But since the queue is already not empty - we can't get real response - we immediately get null result and return.
4) So RFSpy.writeToData receiving null response and saying "No response from RileyLink" while real response was in the end of the queue. I know that because during next call queueSize will be increased by 1 with that last lost RileyLink response.

Here is how that looks in the logs:
```
00:00:58.524 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16946638, timeout is 0 mDataQueue size is 733
00:00:58.525 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpy.writeToDataRaw():209]: writeToData (raw=03 06 0C 59)
00:00:58.868 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16946982, timeout is 7500 mDataQueue size is 734
00:00:58.869 [pool-12-thread-1] E/PUMPBTCOMM: [RFSpy.writeToData():231]: writeToData: No response from RileyLink
00:00:58.971 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16947085, timeout is 0 mDataQueue size is 734
00:00:58.972 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpy.writeToDataRaw():209]: writeToData (raw=03 06 0F 62)
00:00:59.315 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16947429, timeout is 7500 mDataQueue size is 735
00:00:59.317 [pool-12-thread-1] E/PUMPBTCOMM: [RFSpy.writeToData():231]: writeToData: No response from RileyLink
00:00:59.420 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16947534, timeout is 0 mDataQueue size is 735
00:00:59.424 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpy.writeToDataRaw():209]: writeToData (raw=03 06 10 1A)
00:00:59.770 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16947884, timeout is 7500 mDataQueue size is 736
00:00:59.773 [pool-12-thread-1] E/PUMPBTCOMM: [RFSpy.writeToData():231]: writeToData: No response from RileyLink
00:00:59.875 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16947989, timeout is 0 mDataQueue size is 736
00:00:59.876 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpy.writeToDataRaw():209]: writeToData (raw=03 06 11 13)
00:01:00.306 [pool-12-thread-1] D/PUMPBTCOMM: [RFSpyReader.poll():35]: pool-12-thread-1[285]Entering poll at t==16948420, timeout is 7500 mDataQueue size is 737
00:01:00.307 [pool-12-thread-1] E/PUMPBTCOMM: [RFSpy.writeToData():231]: writeToData: No response from RileyLink
```
I think solution here is to let "drain queue" code work well - do not check mDataQueue size in case timeout_ms is 0. More aggressive proposal here can be completely remove if statement, but this will have bigger potential impact (while still looks valid).
This is how line [RFSpyReader:36](https://github.com/nightscout/AndroidAPS/blob/dece9288370c7f54b3d20456173789edf9d94afb/pump/rileylink/src/main/java/app/aaps/pump/common/hw/rileylink/ble/RFSpyReader.kt#L36) might look like:
```
        if (mDataQueue.isEmpty() || timeout_ms == 0) { //0 timeout is used for drain queue in RFSpy.writeToDataRaw before sending new command
```

I also propose to think of removing extra latency in [RFSpy.java](https://github.com/nightscout/AndroidAPS/blob/dece9288370c7f54b3d20456173789edf9d94afb/pump/rileylink/src/main/java/app/aaps/pump/common/hw/rileylink/ble/RFSpy.java#L218) before we go to polling of the response, because we really do not want to enter the poll when result of the sent operation will be already there because in this case we will not read that response from queue successfully because of "mDataQueue.isEmpty()" check in the beginning.
```
        SystemClock.sleep(100);

        return reader.poll(responseTimeout_ms);
```

mifi100:
Make draining of mDataQueue work by fixing poll method filter condition and additionally remove extra sleep to get into polling as soon as possible to start listening before result will come.

fixes #3897

### DIFF
--- a/pump/rileylink/src/main/java/app/aaps/pump/common/hw/rileylink/ble/RFSpy.java
+++ b/pump/rileylink/src/main/java/app/aaps/pump/common/hw/rileylink/ble/RFSpy.java
@@ -215,8 +215,6 @@ public class RFSpy {
             return null; // will be a null (invalid) response
         }
 
-        SystemClock.sleep(100);
-
         return reader.poll(responseTimeout_ms);
     }
 
--- a/pump/rileylink/src/main/java/app/aaps/pump/common/hw/rileylink/ble/RFSpyReader.kt
+++ b/pump/rileylink/src/main/java/app/aaps/pump/common/hw/rileylink/ble/RFSpyReader.kt
@@ -33,7 +33,7 @@ class RFSpyReader internal constructor(private val aapsLogger: AAPSLogger, priva
     // This timeout must be coordinated with the length of the RFSpy radio operation or Bad Things Happen.
     fun poll(timeout_ms: Int): ByteArray? {
         aapsLogger.debug(LTag.PUMPBTCOMM, "${ThreadUtil.sig()}Entering poll at t==${SystemClock.uptimeMillis()}, timeout is $timeout_ms mDataQueue size is ${mDataQueue.size}")
-        if (mDataQueue.isEmpty()) {
+        if (mDataQueue.isEmpty() || timeout_ms == 0) { //0 timeout is used for drain queue in RFSpy.writeToDataRaw before sending new command
             try {
                 // block until timeout or data available.
                 // returns null if timeout.


andyrozman (approved):
I don't see any problem with the code. This is the code that I have taken over when I started project, and we did some small changes when added Omnipod support, but it was mostly unchanged from original author (Geoff I think). 

If @mifi100 did tests and his solution fixes the problem, then let's merge it... If some users have problem with it, then we come back to it... The code like it was might have been needed to be like this in previous versions of firmware (when I started project I had version 0.85 or something like that... And those RL are no longer available (or functioning)...

mifi100:
Hi @andyrozman , 
The issue this change is aimed to fix is sporadic on my phone. I will be able to confirm that fix is working and stable after around one week. Will keep you updated about the results.

PR #3919, opened 2025-03-29
Add expand double

Philoul:
And fix expand color


olorinmaia:
Even better. :) I wonder, is it possible to only show this icon if there is anything to expand? E.g. Profile section only got 1 item.

I cherry picked this one also and working good on master branch.

Philoul:
Yes I know...
Profile, Loop and Objectives are alone in their category.
"Nothing to expand" can be difficult (if all plugins are selected, there is nothing to expand...), but showing the expand icon only if there are several plugins within one category is probably easier...

Issue #3931, opened 2025-04-03
DASH no longer shows the "silence alert" button to disable pod's alerting beeps


vanelsberg:
**AndroidAPS 3.3.3.0-dev**
**Build**: 4d18c1a2
**Flavor**: Fullrelease

When the DASH pod is alerting (_expiration reminder, shutdown, low reservoir_) a button should appear on the DASH tab för user to "silence" the Pod beeps. This button is no longer enabled (made visible). See WiKi [Silencing Pod Alerts](https://wiki.aaps.app/en/latest/CompatiblePumps/OmnipodDASH.html#silencing-pod-alerts)

**Problem description:**
The "Silence Alerts" button on DASH tab is not visible on when Pod is alerting (even after refreshing state).
User can not silence Pod beeping every X minutes.

**Additional information**
Noticed this problem also on [build commit 40068c81ffd5bff07c98763f84cee95f5cb78acc](https://github.com/nightscout/AndroidAPS/commits/dev/) but the problem may also be on earlier commits.

[AndroidAPS_LOG_1743700440381.log.zip](https://github.com/user-attachments/files/19594719/AndroidAPS_LOG_1743700440381.log.zip)

vanelsberg:
Update: after reopening AAPS and moving to the DASH tab, very briefly (less then 0.1 secs?) seeing a "flash" of the "silence" button.
(I'd say the button view gets enabled but then on screen refresh when opening the DASH tab the button view is disabled again?)

@MilosKozak Could this be cause by one of the dev library updates done recently? Not sure where to look.... 🤔

Gohtraw:
Silence Alert reappears after going to "POD MGMT" and tapping "Play Test Beep"

olorinmaia:
I don't experience this in android 12 and Samsung galaxy s10+

vanelsberg:
Interesting: will retry reproducing on my S22 (android 14)

vanelsberg:
Think I found the problem. Working on a fix - soon.

PR #3939, opened 2025-04-09
Fix DASH alert silencing button not showing

vanelsberg:
This PR fixes issue https://github.com/nightscout/AndroidAPS/issues/3931

Under circumstances DASH alert silencing button is not showing it can affect users that migrated from EROS to DASH where in settings "automatic alert silencing" was enabled for EROS.

DASH button is disabled due preferences setting (common for EROS and DASH) for automatic silencing Pod alerts (not supported by DASH). This fix ignores the EROS related preference setting.


olorinmaia:
Tested OK with these steps:
* Switched to EROS and set automatic silencing ON
* Switch back to DASH
* Triggered low reservoir alert
* Silence button visible and working

vanelsberg:
Merges ok with current dev.
On my main phone now shows the silencing button. Silenced "Pod about to expire" alert succesfully.

PR #3952, opened 2025-04-17
Insulin Management outside ProfilePlugin

Philoul:
- New Insulin Plugin
- Dia tab removed from ProfilePlugin and from PureProfile
- ProfileSwitch updated to include Insulin Selection
- Migration updated (from 3.3.2.0 or 3.3.3.0)
- Tests updated


olorinmaia:
Hi @Philoul! :)

I tested this PR upon latest dev + AAPSClient and here are some feedback. Also is this PR a replacement for https://github.com/nightscout/AndroidAPS/pull/3853 ?

If the issues below doesn't appear on your end, is there something I need to do before testing this PR than building upon latest dev?

1) DIA and Insulin doesn't sync to AAPSClient, this Insulin Activity in overview is not working. I'm using NSClientV3 on my testrig. 

| AAPSClient overview | AAPSClient Profile | AAPSClient Profile Switch | 
|--------|--------|--------|
| image | image | image | 

2) For some reason deleted insulin templates are appearing again. I named the template I created Lyumjev, but the marked ones keep appearing in AAPS:
image

Philoul:
@olorinmaia Yes, this PR is a new version with Profiles and insulin completely independant.
When you manage a Profile Switch you select profile on one side and insulin on the other side.

This PR is a first step (already very big), and I have 3 additional PR based on this branch:
- one to include insulin selection into Insulin dialog in overview (this branch also include database synchro for insulin configuration of boluses within database
- one for NS synchronization (to send/receive insulin included within ProfileSwitch, EffectiveProfileSwitch and boluses through NS)
- one for Autotune to include insulin selection for AT calculation.

So issue concerning NS synchro should be fixed with another PR.

Philoul:
@olorinmaia Concerning your second issue, I don't know if it could be a side effect of NS synchro (especially if AAPS version behind source of data is a previous version with dia managed within profile.
Do you have these creation in AAPS version. in AAPSCLIENT version or in both?

The only automatic insulin creation are done in 2 situations:
- during migration (insulin plugin or default peak is given by previous selected insulin plugin and an new insulin is created for each dia value found within profiles
- if a profileswitch is received with an unknown insulin (insulin information is included into EffectiveProfileSwitch).

olorinmaia:
@Philoul I tried disabling NSClient plugin on AAPS and deleted all other profiles and insulin templates. Here is a video of the issue that still persist on my testrig. Seems like insulin template with custom name gets overwritten by "default name". 

Can you reproduce this on your end?

https://photos.app.goo.gl/wAtthmf4pn2qM4tg9

Philoul:
Thanks, I could reproduce the issue (bug with migration, previous key was not removed properly so migration was done again on each launch) 🙈

Philoul:
@olorinmaia Issue fixed with my latest commits
Thanks

Philoul:
PR Closed (Outdated)

PR #4008, opened 2025-05-31
Equil: Fixed race condition preventing commands from being sent

richart77:
This PR fixes a race condition in the Equil driver where enqueued commands were failing to execute. The status check function (running every 10 seconds) was invoking a getter that set the current command to null, even when a command was queued. This caused the driver to remain idle after establishing connection with the pump, as the ready() function invoked after connection saw that the command was null, often resulting with annoying connection errors during boluses/temp basal changes.

### DIFF
--- a/pump/equil/src/main/kotlin/app/aaps/pump/equil/EquilPumpPlugin.kt
+++ b/pump/equil/src/main/kotlin/app/aaps/pump/equil/EquilPumpPlugin.kt
@@ -106,7 +106,7 @@ class EquilPumpPlugin @Inject constructor(
     private val bolusProfile: BolusProfile = BolusProfile()
 
     private val disposable = CompositeDisposable()
-    private var statusChecker: Runnable
+    private lateinit var statusChecker: Runnable
 
     override fun onStart() {
         super.onStart()
@@ -179,7 +179,7 @@ class EquilPumpPlugin @Inject constructor(
     init {
         pumpDescription = PumpDescription().fillFor(pumpType)
         statusChecker = Runnable {
-            var cmd = commandQueue.performing()
+            val cmd = commandQueue.performing()
 
             if (commandQueue.size() == 0 && cmd == null) {
                 if (indexEquilReadStatus >= 5) {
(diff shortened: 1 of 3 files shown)


pump/equil/src/main/kotlin/app/aaps/pump/equil/manager/EquilManager.kt
```diff
@@ -593,7 +593,7 @@ class EquilManager @Inject constructor(
         getTempBasal()?.let { equilTempBasalRecord ->
             val tempBasalStartTime = DateTime(equilTempBasalRecord.startTime)
             val tempBasalEndTime = tempBasalStartTime.plus(equilTempBasalRecord.duration.toLong())
-            return (time.isAfter(tempBasalStartTime) || time.isEqual(tempBasalStartTime)) && time.isBefore(tempBasalEndTime)
+            return (time!!.isAfter(tempBasalStartTime) || time.isEqual(tempBasalStartTime)) && time.isBefore(tempBasalEndTime)
```
MilosKozak, line 596: Not necessary

Hartes: My IDE constantly gave me errors about it which i found kind of weird because it compiled anyways but I added them just to make the type checker happy

pump/equil/src/main/kotlin/app/aaps/pump/equil/EquilPumpPlugin.kt
```diff
@@ -194,7 +194,6 @@ class EquilPumpPlugin @Inject constructor(
                 }
 
             } else {
-                equilManager.readStatus()
```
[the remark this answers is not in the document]
richart77: You're right, they're equal. My thought process was that instead of making a getter that returns Unit and invokes some other functions inside on each "get" it is more natural to actually turn it into a function, for more imperative approach

pump/equil/src/main/kotlin/app/aaps/pump/equil/manager/EquilManager.kt
richart77: My IDE constantly gave me errors about it which i found kind of weird because it compiled anyways but I added them just to make the type checker happy

PR #4028, opened 2025-06-12
Hotfix: Bluetooth Connection for OmniPod Dash on Android 16

### LINKED ISSUE #4027: No Pod activation possible since Android 16 Update
Did the Pixel Android 16 Update Yesterday. 
Connection to Dash wored fine so far.
Today i tried a Pod change. 
Deactivating old Dash worked fine.
But there is no Activation for New Pod possibe

Bonding Mode didn't work too

What are possibilitiea to check?

TebbeUbben:
### For Users: How to Install the Fix
If you're using OmniPod Dash and have updated to Android 16, you may encounter connection issues upon activating a new pod. This hotfix resolves those problems until a new official release of AndroidAPS becomes available.

To install the fix, please follow [these step-by-step instructions](https://github.com/nightscout/AndroidAPS/issues/4027#issuecomment-2968324285) provided by @olorinmaia.

If you need assistance, you're welcome to open a [GitHub issue](https://github.com/nightscout/AndroidAPS/issues) or reach out via our [Discord community](https://discord.gg/4fQUWHZ4Mw).

---

### Summary
Android 16 introduced changes to the internal behavior of the `BluetoothGatt` class. According to the documentation, the `BluetoothGatt::connect` method is intended specifically for reconnecting to a remote device after a connection has been dropped - i.e., reusing an existing `BluetoothGatt` instance.

In earlier Android versions, calling `BluetoothGatt::connect` on a `BluetoothGatt` object that was already connecting or connected would quietly return `true`. However, the new behavior in Android 16 is stricter: the method will now return `false` unless the `BluetoothGatt` instance is in an "idle" state.

This has implications for the OmniPod Dash driver. After calling `BluetoothDevice::connectGatt`, which already starts a connection attempt, a subsequent call to `BluetoothGatt::connect` now returns `false`, even though the connection might still be proceeding. The driver mistakenly interprets this as a failure and disconnects.

I've updated the code to no longer rely on the return value of `BluetoothGatt::connect`. If the connection fails, we should already get a `null` result from `BluetoothDevice::connectGatt`, or the connection timeout logic will kick in appropriately.

### Notes
There’s still an open question as to why devices that were already paired before the system update continued to work correctly, while only fresh pairings seem to fail. I’ll leave that puzzle for another time - it’s getting late, and at least now I can go to sleep with a working loop again. 😅

### Fixes
Fixes #4027


nickb24:
This is an excellent find. I just ran this on a Pixel 7 Pro with Android 16 BP2A.250605.031.A2 and it works.

Philoul:
@TebbeUbben Do you know if other pump drivers can have the same issue?

olorinmaia:
Good job on finding a solution so quick. 

I'm assuming this won't cause issues for android 15 and below? Are someone able to test this also? If not I'll activate a pod on my testrig with this PR on android 12

TebbeUbben:
After a good night of sleep, I figured out why existing pairings continued to work while fresh ones failed.

The driver does reuse instances of `BluetoothGatt` - but only after pairing is complete. During activation (i.e., when the pod is not yet paired), any failed connection attempt leads to the `BluetoothGatt` being closed and discarded. As a result, each connection attempt in the activation phase effectively starts from scratch: `BluetoothDevice::connectGatt` is called again, followed by `BluetoothGatt::connect`, which now returns false under Android 16.

In contrast, once a pod is paired, the BluetoothGatt instance is retained and reused. For these cases, `BluetoothDevice::connectGatt` is not called again - so calling `BluetoothGatt::connect` directly works as expected.

This explains the observed behavior:

- Existing pods: First connection may fail, but retries succeed due to reuse.
- New pods: Each attempt acts like an initial one, triggering the stricter behavior and repeated failures.

I've now restructured the connection logic to ensure that we only call `BluetoothGatt::connect` if the `BluetoothGatt` instance already exists and was not freshly created. This avoids the faulty pattern where `BluetoothGatt::connect` is called prematurely after a new `connectGatt()`.

TebbeUbben:
> @TebbeUbben Do you know if other pump drivers can have the same issue?

I've done a quick code search for all occurrences of `connectGatt` and didn't find any other drivers exhibiting the same faulty pattern. The others seem to consistently close and discard any existing `BluetoothGatt` instance before initiating a new connection, which avoids the issue under Android 16.

> I'm assuming this won't cause issues for android 15 and below? Are someone able to test this also?

I’d expect no regressions here, but a quick check on older Android version would provide peace of mind.

lanceblaise:
Sorry for the n00b question here, but I don't really understand GitHub. Is the only fix the workaround right now, using a backup device to connect to the pod then sending settings to the main (Android 16) device?

lanceblaise:
> > Sorry for the n00b question here, but I don't really understand GitHub. Is the only fix the workaround right now, using a backup device to connect to the pod then sending settings to the main (Android 16) device?
> 
> @lanceblaise You can do that, or you can do what this pull request is saying to change in the code. image

Thank you for the reply. How and where do I change that code? Obviously in Android Studio, but I'm not sure how or where exactly to do that. Thanks again for any assistance. My sugars have been over 400 all night and shots aren't helping much. Hoping to get this resolved today as I don't own an extra phone.

TebbeUbben:
> Sorry for the n00b question here, but I don't really understand GitHub. Is the only fix the workaround right now, using a backup device to connect to the pod then sending settings to the main (Android 16) device?

Open the terminal in Android Studio (might look a little different depending on your operating system):
image

Type:
1. `git fetch` and press ENTER, wait for it to finish
2. `git checkout 39dd2e6151daafa3c98634eec77361b124779aad` and press ENTER, wait for it to finish

Rebuild the AndroidAPS APK.

lanceblaise:
> > Sorry for the n00b question here, but I don't really understand GitHub. Is the only fix the workaround right now, using a backup device to connect to the pod then sending settings to the main (Android 16) device?
> 
> Open the terminal in Android Studio (might look a little different depending on your operating system): image
> 
> Type:
> 
> 1. `git fetch` and press ENTER, wait for it to finish
> 2. `git checkout 39dd2e6151daafa3c98634eec77361b124779aad` and press ENTER, wait for it to finish
> 
> Rebuild the AndroidAPS APK.

Thank you. I am going to try this in an hour when I return home. Fingers crossed I do it correctly. 🙂

lanceblaise:
> > Sorry for the n00b question here, but I don't really understand GitHub. Is the only fix the workaround right now, using a backup device to connect to the pod then sending settings to the main (Android 16) device?
> 
> Open the terminal in Android Studio (might look a little different depending on your operating system): image
> 
> Type:
> 
> 1. `git fetch` and press ENTER, wait for it to finish
> 2. `git checkout 39dd2e6151daafa3c98634eec77361b124779aad` and press ENTER, wait for it to finish
> 
> Rebuild the AndroidAPS APK.

I followed your directions, but got an error on the build process.  Any idea on what went wrong?  It took almost 30 minutes to build then the errors popped up.

Screen Shot 2025-06-12 at 10 26 58 AM

lanceblaise:
> Click on 'sync'

Ok. Do I need to perform the above steps again? Git fetch and git checkout?

lanceblaise:
> Try following these steps: https://androidaps.readthedocs.io/en/latest/GettingHelp/TroubleshootingAndroidStudio.html#gradle-resync

same error

lanceblaise:
> Installing directly from Google Drive is always finicky. Save the APK locally to the phone and install that way. Anyways this is beyond the scope of this PR at this point. Now this is how to install AAPS as usual.

It won't even save to my phone. I tried saving to my phone first. Something is wrong with the build.  Also meerkat is showing me this...

lanceblaise:
Installed the update and now I get a new error in AAPS. 
Screenshot_20250612-114706

lanceblaise:
Tried a new pod and now I get this error...

Screenshot_20250612-115418

lanceblaise:
Screenshot_20250612-120645

lanceblaise:
> What worked for me was the commit from last night: `6153bc4b257222a1d29a2871d6fa4223ce4e4ac7`
> 
> Didn't try the new one from today. Maybe try the first commit. So you can run the same commands but change the number.

Ok. I will give that a try.

TebbeUbben:
@lanceblaise Thanks for your engagement. Unfortunately, the discussion in this thread started drifting off-topic, so I had to remove some comments to keep the PR focused. For further questions or support, please use the appropriate channels, such as Discord or GitHub Issues. Appreciate your understanding!

lanceblaise:
> @lanceblaise Thanks for your engagement. Unfortunately, the discussion in this thread started drifting off-topic, so I had to remove some comments to keep the PR focused. For further questions or support, please use the appropriate channels, such as Discord or GitHub Issues. Appreciate your understanding!

Ok. Just want to make sure people know that this "fix" is not working. 
Screenshot_20250612-134450

TebbeUbben:
> Ok. Just want to make sure people know that this "fix" is not working.

Thanks for your feedback. I’ve just tested my changes on a spare Pixel 8 running Android 16, and I was able to successfully activate a pod. My main device (Pixel 9 Pro, A16) also continues to work without any issues.

From your earlier screenshots, it looks like there was a pod fault due to mechanical errors. While that’s certainly frustrating, it also confirms that your phone was able to connect to the pod - otherwise, it wouldn’t have been able to retrieve its state.

To make sure the updated version is actually installed on your phone, could you please double-check the commit hash shown in the build number on the "About" screen of the app?

If the issue persists, it would really help if you could upload log files so we can better understand what’s going on and support you in troubleshooting. Please open a separate issue for that.

lanceblaise:
> > Ok. Just want to make sure people know that this "fix" is not working.
> 
> Thanks for your feedback. I’ve just tested my changes on a spare Pixel 8 running Android 16, and I was able to successfully activate a pod. My main device (Pixel 9 Pro, A16) also continues to work without any issues.
> 
> From your earlier screenshots, it looks like there was a pod fault due to mechanical errors. While that’s certainly frustrating, it also confirms that your phone was able to connect to the pod - otherwise, it wouldn’t have been able to retrieve its state.
> 
> To make sure the updated version is actually installed on your phone, could you please double-check the commit hash shown in the build number on the "About" screen of the app?
> 
> If the issue persists, it would really help if you could upload log files so we can better understand what’s going on and support you in troubleshooting. Please open a separate issue for that.

So when I went through the normal AAPS pairing process I got a pop up asking me to pair to the pod like you normally get a request to pair to a regular Bluetooth device like headphones or something. I let it pair like that, and then it went back to aap trying to pair and failed. Could that be the issue?

Yes under about, the build is the first 10 numbers of that code you gave me followed by a hash and todays date. 

Do you want the logs from my phone or Android Studio?

Thank you. 

Screenshot_20250612-154353~2

olorinmaia:
Cherry picked 39dd2e61 and 6153bc4b onto master branch and installed apk on my test-rig and activated a pod. Tested so far OK on android 12 with bonding OFF. I assume you don't need to use the bond option ON on A16 with these changes either. Pod is running fine. Will let you know if I find any issues. 

Would be interesting to hear if those experiencing issues with Android 15 and use the manual bonding and restarting, if the issue is resolved for them with these commits and turning bonding off..?

Also got an A15 phone I can test with if need be, on that phone I don't experience any issues without this PR.

image

image

nickb24:
I've been running with this change since it was released yesterday on a Pixel 7 Pro on Android 16 with the bonding option Off. No issues with perfect connection quality.

lanceblaise:
> I've been running with this change since it was released yesterday on a Pixel 7 Pro on Android 16 with the bonding option Off. No issues with perfect connection quality.

Did you get a regular Bluetooth pairing request when you did this? Like the same way you would get a pairing request for headphones etc.

nickb24:
> > I've been running with this change since it was released yesterday on a Pixel 7 Pro on Android 16 with the bonding option Off. No issues with perfect connection quality.
> 
> Did you get a regular Bluetooth pairing request when you did this? Like the same way you would get a pairing request for headphones etc.

I didn't. That should only happen if you have the Dash bond enabled. No where in this code change could a bond initialization occur.

lanceblaise:
> > > I've been running with this change since it was released yesterday on a Pixel 7 Pro on Android 16 with the bonding option Off. No issues with perfect connection quality.
> > 
> > 
> > Did you get a regular Bluetooth pairing request when you did this? Like the same way you would get a pairing request for headphones etc.
> 
> I didn't. That should only happen if you have the Dash bond enabled. No where in this code change could a bond initialization occur.

Ugh. Thank you! I turned that on last night during my initial tries at fixing this Bluetooth issue because it was suggested on Facebook. Like an idiot I never turned it off. This was causing my issue today. So glad I mentioned it and you responded telling me why that had popped up. I feel like such a moron! Thank you so much for helping me through out the day with the build and figuring this major mistake on my part!

TebbeUbben:
> Would be interesting to hear if those experiencing issues with Android 15 and use the manual bonding and restarting, if the issue is resolved for them with these commits and turning bonding off..?

That would be interesting indeed! Personally, I've never experienced any issues with Android 15 on both my Pixel 8 and Pixel 9 Pro.

nickb24:
> > Would be interesting to hear if those experiencing issues with Android 15 and use the manual bonding and restarting, if the issue is resolved for them with these commits and turning bonding off..?
> 
> That would be interesting indeed! Personally, I've never experienced any issues with Android 15 on both my Pixel 8 and Pixel 9 Pro.

I've been the only advocate that I know so far that's been asking for help investigating the issue on Android 15. However since my phone doesn't have the issue I couldn't investigate further. You can follow my discussions on the GitHub issue or Discord. 

As far as I can tell, the error logs look similar. Maybe more errors in service discovery on Android 15.

borijess:
How do I merge this hotfix with 3.3.2.0?

olorinmaia:
> How do I merge this hotfix with 3.3.2.0?

Please ask for help for this on Discord or in the issue: https://github.com/nightscout/AndroidAPS/issues/4027

lanceblaise:
> > Would be interesting to hear if those experiencing issues with Android 15 and use the manual bonding and restarting, if the issue is resolved for them with these commits and turning bonding off..?
> 
> That would be interesting indeed! Personally, I've never experienced any issues with Android 15 on both my Pixel 8 and Pixel 9 Pro.

Same here. For me Android 15 was the most stable version that I ever had AAPS running on.

borijess:
> > How do I merge this hotfix with 3.3.2.0?
> 
> Please ask for help for this on Discord or in the issue: #4027

I think I got it. I did Fetch and Pull and did pull hotfix-a16-dash, then a rebase and it updated. Is this correct? Then build the signed APK.

borijess:
Besides the obvious the pod connects. How would you know the app hotfix was applied correctly to the APK after installed on phone?

MJDSys:
> Would be interesting to hear if those experiencing issues with Android 15 and use the manual bonding and restarting, if the issue is resolved for them with these commits and turning bonding off..?

I only use AAPS on Android 15 with a Pixel 9 Pro Fold after some bug fixes were released.  With manual bonding on, I had to do the restart dance, with manual bonding off everything worked great.  I wonder if Google quietly released something that fixed it, or newer devices don't experience the bug?

I also upgraded to Android 16 and with this branch everything seems to be 👍 .

Gohtraw:
@TebbeUbben Thank you! I merged the hotfix into AutoISF and my setup is running without a hitch.

Android Studio was complaining that things might go wrong if the user revokes bluetooth permissions so I added some 
@RequiresPermission(Manifest.permission.BLUETOOTH_CONNECT) before the functions that need it

kubatom:
@TebbeUbben I can confirm it works on Android 16, Pixel 7! Amazing work, you're a lifesaver

olorinmaia:
Update from me.

Activated pods on two different rigs, Samsung galaxy S10+ (Android 12) and Samsung Galaxy S23 (Android 15) with 100% connection and deactivated in the end without any issues.

I've seen several users on Facebook reporting they fixed their issue with Android 16 Activation of Dash pods with these commits. Good job. It's looking good!

ToreBj:
I'm adding a code review for this fix which was done with the support from Chatgpt o4-mini-high:

That little if (!gatt.connect()) … guard was simply there as an early sanity check—it ensured that your call to BluetoothGatt.connect() actually queued a new connection attempt before you went off and started waiting for the asynchronous callback. In other words:

Original concern:
On some Android versions (or in some error states), gatt.connect() can immediately return false if the BLE stack can’t even start a connect (for example, if the GATT instance isn’t in a state where it can open the ACL link). Throwing right away meant “don’t hang around waiting for something that will never happen”—you’d fail fast instead of waiting out your timeout.

Why it backfired on Android 16:
Google tightened the rules so that, on a fresh BluetoothGatt (one you just got from connectGatt()), connect() will return false until you’ve first disconnected it once. But that doesn’t mean “no connection attempt was started”—it only means “you called connect() too soon on a non-idle GATT.” Your real handshake was already in motion via connectGatt().

Since you already have robust timeout-and-callback logic in waitForConnection(...), that immediate false was giving a false negative on Android 16. Removing the throw lets you:

Always kick off the link via connectGatt(...)

Call gatt.connect() unconditionally (harmless on fresh GATTs)

Rely on waitForConnection(...) to catch true failures (no callback, timeout or manual stop)

So in short, the guard was a good “fail-fast” check on older platforms, but on A16 its semantics changed under your feet—causing premature exceptions rather than real failures—so it made sense to drop it and lean entirely on your callback-driven timeout.

Me:
And there's a good reason to think that this will work correctly for A15? What about android versions before that?


ChatGPT :
Yes—it’s safe on Android 15 and even all the way back through the earlier releases, because:

On Android 15 and below, BluetoothGatt.connect() was effectively a no-op that always returned true (as long as you’d called connectGatt() first). So your old code never saw a false there, and dropping the check changes nothing in practice.

On really old Androids (pre-BLE days like Jelly Bean), you’d only ever call connectGatt() and didn’t even have a separate connect() call in your app. Once BLE arrived (KitKat/Marshmallow), the behavior has been consistent: connect() returns true whenever the stack accepted the request.

By removing the “throw if connect() is false” guard, you:

Still kick off the link with connectGatt() (that’s the real allocator and opener).

Call gatt.connect(), which on A15− and A16+ is harmless when ignored.

Rely solely on your waitForConnection(...) latch/timeout logic to detect actual failures (no callback, timeout or manual stop).

So you’ll get the same behavior you had on A15 (and earlier) plus correct behavior on A16, all with one unified code path.

My conclusion is that TebbeUbben has done a very good assessment of the problem and made a code change that I have tested myself.

My testing was done on a Google Pixel 8a on AAPS 3,2,0,4 Boost version lagging a bit behind. I had no problems with A15
no bond problems.

ToreBj:
Hi, I only reviewed this change and I only tested this change because it alone fixes the problem.

https://github.com/nightscout/AndroidAPS/commit/6153bc4b257222a1d29a2871d6fa4223ce4e4ac7

However I discovered that @olorinmaia had cherrypicked two fixes and by accident I looked at the other fix:

https://github.com/nightscout/AndroidAPS/commit/39dd2e6151daafa3c98634eec77361b124779aad

which effectively nullifies the first one?
This must be a mistake and I would like to hear what others think @TebbeUbben .

Chatgpt's take on this:
A16fix




Re-introduces the throw on !gatt.connect(), precisely the check we removed to work around Android 16’s stricter semantics.
Misorders the retry logic so the first throw will fire before it ever calls connectGatt() again, making the subsequent retry dead code.
Defeats the whole purpose of relying on waitForConnection(...) as the single source of truth for success/failure.

In short, applying this patch would:

- Break Android 16 again (you’d once more see connect() return false and immediately throw),

- Add nonsensical retry logic that never executes, and

- Throw away the robust timeout–and–callback approach you already have.

My recommendation is to leave the Android 16 fix in place (i.e. drop any if (!gatt.connect()) throw …) and, if you need retries on a null gattConnection, do that before you throw––but don’t re-check connect() itself. The real connect-success test remains your waitForConnection(...) logic.

ToreBj:
I rest my case, both fixes work independently. I was mistaken about how the commits were done in the "A16 fix" build.

TebbeUbben:
> Hi, I only reviewed this change and I only tested this change because it alone fixes the problem.
> 
> [6153bc4](https://github.com/nightscout/AndroidAPS/commit/6153bc4b257222a1d29a2871d6fa4223ce4e4ac7)
> 
> However I discovered that @olorinmaia had cherrypicked two fixes and by accident I looked at the other fix:
> 
> [39dd2e6](https://github.com/nightscout/AndroidAPS/commit/39dd2e6151daafa3c98634eec77361b124779aad)
> 
> which effectively nullifies the first one? This must be a mistake and I would like to hear what others think @TebbeUbben .
> 
> Chatgpt's take on this: A16fix
> 
> Re-introduces the throw on !gatt.connect(), precisely the check we removed to work around Android 16’s stricter semantics. Misorders the retry logic so the first throw will fire before it ever calls connectGatt() again, making the subsequent retry dead code. Defeats the whole purpose of relying on waitForConnection(...) as the single source of truth for success/failure.
> 
> In short, applying this patch would:
> 
> * Break Android 16 again (you’d once more see connect() return false and immediately throw),
> * Add nonsensical retry logic that never executes, and
> * Throw away the robust timeout–and–callback approach you already have.
> 
> My recommendation is to leave the Android 16 fix in place (i.e. drop any if (!gatt.connect()) throw …) and, if you need retries on a null gattConnection, do that before you throw––but don’t re-check connect() itself. The real connect-success test remains your waitForConnection(...) logic.

My initial commit simply ignored the result of `BluetoothGatt::connect` after identifying this call as the root of the problem. However, further investigation showed, that it simply must not be used in conjunction with `BluetoothDevice::connectGatt`, but only when reusing an existing instance of `BluetoothGatt`. In these cases, the result of `BluetoothGatt::connect` is actually meaningful and if it returns `false`, we can safely assume that the connection attempt failed and trigger an exception. This is what my second commit does. If there already is an instance of `BluetoothGatt`, it uses `BluetoothGatt::connect` and relies on its return value to determine an error state. Otherwise `BluetoothDevice::connectGatt` is used and `BluetoothGatt::connect` is not called.

Also, the retry-logic is not implemented in `Connection::connect`, which is designed to only connect once or throw an exception if it fails.

Furthermore, I’d kindly ask you not to copy and paste unreviewed AI-generated responses. If I want to consult ChatGPT, I can (and often do) do that myself. What’s most helpful in code reviews is your own reasoning or a specific concern - especially since AI tools often miss important context or nuance. Thanks for understanding!

t1dude:
I've tested this on two setups:

Pixel 9 Pro with Android 16
Samsung Galaxy Z Flip 5 with Android 15

Both setups are working fine, proving that the fix does not break anything in Android 15. In both setups I am not using the bonding feature, as it is not needed with these two phones. 

However I have seen one issue, not sure if it's related to this fix: when I change pods, AAPS will still try to send a TBR command, leading to an alarm that TBR cannot be set while I'm in the middle of preparing a new pod. It seems AAPS is not aware that the previous pod has been disconnected and that there is no active pod.

Jared-Is-Coding:
I was rudely introduced to this bug at 1am after my Pixel 7a forced the Android 16 update on me. Went to start a new pod and could not. Was repeatedly getting the "retry" button appearing when attempting to initialize the pod. Did a quick search around, found myself on this branch + PR.

Previously:
- Pixel 7a + Android **15** + AAPS 3.3.1.0 -> Working fine
- Pixel 7a + Android **16** + AAPS 3.3.1.0 -> Initialization failing. Repeatedly arriving at the "Failed to connect to pod" screen with a retry button, same as @lanceblaise 

Now:
- Pixel 7a + Android 16 + AAPS 3.3.3.0-dev-a + this PR's commits -> Working as expected. Omnipod Dash initialized, connected, and looping. Receiving commands as expected. Will update when I hit the end of this Pod's life cycle.

Additional reference:
- I do not use the bonding fix setting; that is disabled.

Thank you! @TebbeUbben

PR #4035, opened 2025-06-16
Android 16 fix Locale messages in logs

nickb24:
After upgrading to Android 16, I noticed that the logcat logs are filled with messages saying "Updating configuration, locales updated from [] to [en_US]"

image

The messages are being triggered because the code is creating empty Configuration() objects and then setting locales on them, which causes Android 16 to log these updates.

This PR fixes that issue and at the same time updated the depreciated `Locale(language)` and `Locale(lang, country)` located in the LocaleHelper.kt


Gohtraw:
Thank you - works like a charm for me.

PR #4046, opened 2025-06-25
Add Juggluco support to xDrip+ source strings

### LINKED ISSUE #4042: Add Juggluco to BG source list
As Juggluco imitates xDrip communications and can be used as a BG source (if AAPS is set to xDrip), I think it would be logical to have _xDrip / Juggluco_ as a BG source (instead of just _xDrip_ )

Cleboost:
Updated translations in all supported languages to reference both xDrip+ and Juggluco as sources for BG/glucose values. This clarifies that the app can now receive data from either xDrip+ or Juggluco.

*fix #4042

### DIFF
--- a/plugins/source/src/main/res/values-bg-rBG/strings.xml
+++ b/plugins/source/src/main/res/values-bg-rBG/strings.xml
@@ -4,8 +4,8 @@
     <string name="ns_client_bg">КЗ от NS</string>
     <string name="ns_client_bg_short">NS КЗ</string>
     <string name="description_source_ns_client">Изтегля стойности на КЗ от Nightscout</string>
-    <string name="source_xdrip">КЗ от xDrip+</string>
-    <string name="description_source_xdrip">Получаване на данни за КЗ от xDrip+.</string>
+    <string name="source_xdrip">КЗ от xDrip+ / Juggluco</string>
+    <string name="description_source_xdrip">Получаване на данни за КЗ от xDrip+ или Juggluco.</string>
     <string name="dexcom_app_patched">BYODA</string>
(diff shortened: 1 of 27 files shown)


olorinmaia:
It is only needed to update the English string. The rest will need to be updated through crowdin if PR gets merged to dev.

You also need to target dev branch, not master 👍

Cleboost:
So I delete all the changes and keep only the English string?

olorinmaia:
Yes. I'll also leave the same comment as I did in the issue here. @MilosKozak have the final say in PRs. 🙂👍

Wiki doesn't say much about Juggluco being able to broadcast directly to AAPS. If that is so, it makes more sense to make an own option for Juggluco. In case adjustments must be made to either xDrip or Juggluco in the future. But I'll leave that up to Milos to decide.

olorinmaia:
Some more info taken from Juggluco wiki https://www.juggluco.nl/Jugglucohelp/exchangehelp.html

**xDrip broadcast**
Send the same broadcast as turned on in xDrip with "Settings->Inter-app settings->Broadcast locally". Apps like AndroidAPS listen to it. Unlike the original xDrip broadcast which was broadcast every 5 minutes, Juggluco broadcasts this broadcast every minute.

swissalpine:
And the minute-by-minute broadcast is already a significant difference: it can cause AAPS to crash and hang on weaker smartphones. This broadcast may need an option (if necessary) to evaluate only every 5th value and use it for a loop calculation. In any case, it does matter whether the data comes from xDrip or Juglucco. It's not just about the superfluous naming (Juglucco itself says that it imitates the xDrip broadcast, so every user should know for themselves that I then also select xDrip).

PR #4082, opened 2025-08-02
Fix a typo in an insulin name

psfinaki:
It is Humulin, not Humalin: https://www.drugs.com/humulin-r.html


Philoul:
When you see a typo, only the main language (english, it's within `.../values` folder) should be included within your commit, never modify translations.
Translations are managed by translators/proofreaders within crowdin.

In your example, it's the string included into `.../main/res/values/exam.xml` file.

Philoul:
And if typo is within a translation then the proposal should be done in crowdin in the corresponding language.

Note that when a "main string" is created or modified within androidaps repo dev branch, then it's is shown as "not translated" in crowdin 😉.

psfinaki:
@Philoul sure - thanks for the explanations! I untranslated translations and left only the main string now :)

PR #4093, opened 2025-08-10
3330 dev b/dev mini graphs

ga-zelle:
This is about the layout of the subgraphs.

Change 1: depending on the default font size of the phone the bottom label of all the subgraphs is partly obscured. So I moved the vertical labels somewhat inwards.

Change 2: after moving the labels in change 1 the margin between the subgraphs can be reduced. Now the screen can show more of the subgraph(s).

grafik


MilosKozak:
I'd like to avoid changes to original libs. Any update can break this

Issue #4134, opened 2025-09-08
AAPS CI keeps failing (dev branch)


vanelsberg:
08/09/2025:
For the past week AAPS CI build on the dev branch keeps failing.

<img width="952" height="610" alt="Image" src="https://github.com/user-attachments/assets/957e2970-885c-4539-a02c-31edf4e6dff8" />

<img width="935" height="470" alt="Image" src="https://github.com/user-attachments/assets/2aa8c05d-9185-4e48-827f-e4d5274c887a" />

MilosKozak:
it's local issue. Mine is working fine https://app.circleci.com/pipelines/github/nightscout

qq254344379:
I also encountered a situation where the browser could not compile Dev.

<img width="1492" height="858" alt="Image" src="https://github.com/user-attachments/assets/4794f30a-25ab-413b-9584-9eb3af5f154a" />

vanelsberg:
> I also encountered a situation where the browser could not compile Dev.

Confirming. At the moment, while building dev/fullRelease I also see exactly this error.
But it does not always fail at the same location (after code changes?)

Angus-repo:
Hello everyone,

I’ve tested it and was able to reproduce the issue on my own computer.

When using Android Studio’s interface on the dev branch to build the APK, everything works fine.
The main difference with the AAPS-CI command line is that AAPS-CI builds both the app and wear simultaneously, which can sometimes cause ordering issues.

I will adjust it so that AAPS-CI builds the app first, then the wear, which should prevent this problem.

vanelsberg:
> I will adjust it so that AAPS-CI builds the app first, then the wear, which should prevent this problem.

Interesting. Wil have a look.
Actually this should not be a problem. My Docker Autobuild solution (uses command line build) can still build App en Wear-app in a single build without problems.

vanelsberg:
> I’ve tested it and was able to reproduce the issue on my own computer.

1) What I have seen so far (for me), for current dev is building fulllDebug without problems.
But as soon as I try to build **fullRelease it fails**? Tried deleting cache but that does not solve.... 🤔

2) Also seeing some problems when building with the latest version of Android Studio b.t.w:
- Building signed APK's is no problem.
- Assemble App run configuration is OK
A build failure occurs for "Assemble Project"?

Angus-repo:
@vanelsberg 

I have opened a PR #4139  with this change.  
Please help test the above cases to confirm stability.

[...]

PR #4139, opened 2025-09-10
aaps-ci: adjust assemblefullRelease order for app and wear

Angus-repo:
AAPS-CI previously invoked `assemblefullRelease` on both app and wear
modules together, which sometimes caused build order issues.
This change enforces building app first, then wear, to ensure stability.

Related to #4134 

### Test Cases

**Branch**
- dev
- master

**Variant**
- fullRelease
- fullDebug
- aapsclientRelease
- aapsclientDebug
- aapsclient2Release
- aapsclient2Debug
- pumpcontrolRelease
- pumpcontrolDebug

### DIFF
--- a/.github/workflows/aaps-ci.yml
+++ b/.github/workflows/aaps-ci.yml
@@ -188,7 +188,7 @@ jobs:
 
       - name: Build APKs
         run: |
-              ./gradlew assemble${{ env.BUILD_VARIANT }} \
+              ./gradlew :app:assemble${{ env.BUILD_VARIANT }} :wear:assemble${{ env.BUILD_VARIANT }} \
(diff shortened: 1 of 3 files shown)


MilosKozak:
is there a way to use debug keys from studio?

Angus-repo:
> is there a way to use debug keys from studio?

Yes,
the Android Studio debug key at the path ~/.android/debug.keystore can be used by AAPS-CI to build the APK.

However, please note that when converting with aaps-ci-preparation.html, you need to add the .jks extension — for example, rename it to debug.keystore.jks — so that it can be selected in Option 2.

vanelsberg:
@Angus-repo Merged this into my local dev branche for testing. Congrats - I think you solved the problem!

@MilosKozak I am ok with merging this PR into dev!

PR #4149, opened 2025-09-11
overlay bg parabola as line graph

ga-zelle:
What does it do? In the main graph it shows the result of the parabola which best fitted the recent BG history. It goes backwards from now by as many minutes as were included in this optimal fit. If showing predictions is enabled, too, then it also shows the extrapolation from now for the next 20 minute as a dotted line. If the fit was unsatisfactory at this point in time then no curve is shown.

For the AutoISF algorithm this parabola is a key information. Having used this display for some weeks it was surprising how a very short glance shows what is happening right now. While the BG points may go fairly straight up or down the fitted parabola may already show that the end of that trend is near.

For the other APS algorithms it is one more alternative to the existing predictions even when the code does not take any action based on it. Users who feel confused can deselect it in the graph menu. For the future I plan to use the parabola data to calculate smoother values for delta and short average delta benefitting all algorithms.


ga-zelle:
Here is an example screenshot with the extrapolation of the parabola partially hidden by the graph menu:

grafik


A bitlater it looks like this: 

grafik

MilosKozak:
generaly I'm OK with this but as there is more and more code related only to one specific alg, we should make this code (and others) performed only when it's needed

MilosKozak:
check this https://github.com/nightscout/AndroidAPS/commit/a6dfb3cf81f0cee133451f1fa6cf8b0130a536a8

ga-zelle:
Yes, inserting visibility makes things a bit easier than my solution. The main impact will be when I add more graphics for key AutoISF parameters. I'll adapt my approach so all use the same for the same purpose.

On a first read I have 2 comments:
1. var_sens should also be visible when AUTOISF plugin is active
2. setting visibility in "init" - wouldn't that miss it when things change over time? Or am I misinterpreting when"init" gets called?

MilosKozak:
> Yes, inserting visibility makes things a bit easier than my solution. The main impact will be when I add more graphics for key AutoISF parameters. I'll adapt my approach so all use the same for the same purpose.
> 
> On a first read I have 2 comments:
> 
> 1. var_sens should also be visible when AUTOISF plugin is active
update the rule it in your PR
> 2. setting visibility in "init" - wouldn't that miss it when things change over time? Or am I misinterpreting when"init" gets called?
enums don't  change over time and it's not possible to use injected modules directly in enum definition

MilosKozak:
I did one more update to allow resolve visibilty dynamically

ga-zelle:
Tonight I wanted to pull it all together. However, GlucoseStatusProvider no longer contains the AutoIsf pecific results for the parabola. It appears you moved that and I have no idea how to get that into OverviewMenusImpl. I tried with GlucoseStatusAutoIsf, but then dagger does not know about it ...

This is beyond my Kotlin skills. I could try with some hint. Gemini speaks double dutch.

Never mind, I got over that hurdle.

ga-zelle:
1. I did not manage to include my "runningAutoIsf" flag into the init section. It always complained about no APS selected and even crashed on start.
2. I added that no graph is calculated if it is classified as invisible.
3. A non-critical but awkward condition can be created by activating VSENS in AutoISF and then switching to AMA plugin. The related menu entry is of course not there to switch it off. I managed to get rid of the empty graph frame and the NaN labels but the header for VSENS remains.
4. This may be unrelated, but the progress bar at the bottom stays visible in SMB plugin. AMA and AUTOISF plugins are OK

ga-zelle:
With my last edit the header VSENS is no longer shown even when plugin is switched which makes it invisible implicitly.

MilosKozak:
unmergeable now

Issue #3367, opened 2024-06-15
Timeline in History screen is not Always correct


francoisqueru35:
In the history screen, while switching scale, the timeline is not correct for 6 and 12 hours.
For 6 : the current hour appears two times (see screenshots attached)
For 12 hours :  the graph is one hour wrong (i.e 7 instead of 8 on the screenshots 

Any idea why?
![Screenshot_20240615-084422](https://github.com/nightscout/AndroidAPS/assets/171514589/602f2b08-2294-494f-b6da-42797d9761dc)

![Screenshot_20240615-081518](https://github.com/nightscout/AndroidAPS/assets/171514589/4ea18561-04d1-4441-a8bb-89a14aa498de)
![Screenshot_20240615-081534](https://github.com/nightscout/AndroidAPS/assets/171514589/f640c395-2a91-4f23-81b9-75ac82f77b48)
![Screenshot_20240615-081629](https://github.com/nightscout/AndroidAPS/assets/171514589/e444f0bb-fa5f-4e17-82dc-631a2912a9df)
![Screenshot_20240615-081546~2](https://github.com/nightscout/AndroidAPS/assets/171514589/5204e9dd-4b5b-470a-be92-db824aa292fe)
![Screenshot_20240615-081557~2](https://github.com/nightscout/AndroidAPS/assets/171514589/caa88064-2020-4f94-b2ee-d19731e8284e)
Reporting bugs
--------------
- **Note the precise time the problem occurred** and describe the circumstances and steps that caused
  the problem
- Note the Build version (found in the About dialog in the app, when pressing the three dots in the
  upper-right corner).
- Obtain the app's log files, which can be found on the phone in
  _/storage/emulated/0/Android/data/info.nightscout.androidaps/_
  See https://wiki.aaps.app/en/latest/Usage/Accessing-logfiles.html
- Open an issue at https://github.com/nightscout/AndroidAPS/issues/new

fezulin:
I observe the same behavior in both version 3.2.0.4 and version 3.3.0.0.
In the first image: On the left is a part of the "History" (This is not a screenshot from the main view!). On the right is a screenshot of the "Treatments".
1.) Left grey arrow: 
Apparently, two SMBs were delivered around 5:30 AM. 
2.) Right grey arrow: 
According to the "Treatment", the SMBs were given at 6:30 AM. One hour later! (And this is the time I remember.)
3.) Red square: 
The "08" is missing. So, an entire hour is missing. 
[...]

Issue #2854, opened 2023-10-02
NSClient History time-axis wrong scaling


ben80:
Problem
--------------
The time-axis in the history window is not scaled properly for all hourly-steps.
The time difference between the vertical grid lines is not always constant.

Examples - wrong grid
--------------
See screenshots attached.

Screenshot_20230927-062114.png: between 4 and 5 o'clock only one hour difference instead of 2
![Screenshot_20230927-062114](https://github.com/nightscout/AndroidAPS/assets/8597725/2885f2cb-af98-4b5c-bb96-3bdacf07e296)

Screenshot_20231002-082624.png: between 7 and 10 o'clock 3 hours difference instead of 2
Screenshot_20231002-082746.png: between 8 and 12 o'clock 4 hours difference instead of 3

Examples - plot scaling seems to be ok
--------------

MilosKozak:
wasn't it day of DST change?

ben80:
No, that's what i first thought, also.
But DST change will be oct 29th 2023 only (in one month).

I think the plot is scaled correctly. 
But the ticks of the vertical grid lines are not equidistant within the graph.
Should be e.g.:  0 - 2 - 4 - 6 ... o'clock
But is:                0 - 2 - 5 - 7 ... o'clock

MilosKozak:
https://github.com/nightscout/AndroidAPS/issues/2729

MilosKozak:
it's a bug in library

PR #4221, opened 2025-10-12
Small fix to time range in the history browser

psfinaki:
As was discussed in a Discord [thread](https://discord.com/channels/629952586895851530/629954570394533889/1412413761972732015) - this aligns the time range extraction to the way it's done in [PreparePredictionsWorker](https://github.com/nightscout/AndroidAPS/blob/master/workflow/src/main/kotlin/app/aaps/workflow/PreparePredictionsWorker.kt#L66).

### DIFF
--- a/app/src/main/kotlin/app/aaps/activities/HistoryBrowseActivity.kt
+++ b/app/src/main/kotlin/app/aaps/activities/HistoryBrowseActivity.kt
@@ -264,7 +264,10 @@ class HistoryBrowseActivity : TranslatedDaggerAppCompatActivity() {
 
     private fun adjustTimeRange(start: Long) {
         historyBrowserData.overviewData.fromTime = start
-        historyBrowserData.overviewData.toTime = historyBrowserData.overviewData.fromTime + T.hours(rangeToDisplay.toLong()).msecs()
+        historyBrowserData.overviewData.toTime =
+            historyBrowserData.overviewData.fromTime +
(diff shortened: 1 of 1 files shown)


ga-zelle:
@psfinaki When I cycle through a day using the 6 hour window it becomes quite obvious that you should have applied the fix one line earlier, i.e. already to the start time of the window.

The original bug was submitted several times. Those issues can be closed like https://github.com/nightscout/AndroidAPS/issues/3367 https://github.com/nightscout/AndroidAPS/issues/2854

psfinaki:
Hmmm @ga-zelle alright, I might have misunderstood something then. But then also why it's different on the main screen?

But thanks for the links, I will go through them and (re)submit a fix this week if needed.

ga-zelle:
My approach to codeing is 
1. develop a plan how to code it 
2. develop a plan how rto test it
3. do the code 
4. do the test

In this case the test is really easy because there are only 7 possible outcomes for 7 possible time axes and they can all be executed quickly. So I went through 24h/12h/6h width and in there through those sub windows. You will see that your fixed the end of each window but (some?)  start value labels are still wrong. The reason is in main window it starts from end time including the 100s offset and then subtracts the window length to get the start value. Therefore start inherits the offset. In History Browser it starts at the beginning and then adds the length of the window. You added the offet to the end only. If you add it to the start then the end will inherit the offset and everything is fine.

Issue #4201, opened 2025-10-07
Safety issue: can't stop an ongoing bolus after closing AAPS


vanelsberg:
**AndroidAPS 3.3.3-dev**: Lates dev builds
**Flavor**: fullrelease

**Problem:**
When executing a manual bolus the resulting pop-up screen will show bolus progress and a "STOP" button to cancel the onging bolus. _Closing the app (using Android UI "Close All" or swiping) does not stop the bolus by design. However, when I reopen the AAPS app it only shows the (still) ongoing bolus at the bottom of the overview screen without a STOP button to cancel._

Considering safety, one could argue that this is not acceptable

**Additional info:**
To STOP the bolus then user has no other option then to forcefully close AAPS through settings, restart the phone, disconnect the pump or 
(when running the AAPS Wear app) using the option to cancel the bolus from the watch.

_Screenprints:_

**Ongoing bolus with STOP button enables canceling the ongoing bolus.**

![Image](https://github.com/user-attachments/assets/f5f6fa0d-1814-4d58-adba-179ba4b04492)

**After closing and reopening AAPS: ther is no way to STOP the ongoing bolus**

![Image](https://github.com/user-attachments/assets/15cc5246-7b30-4d1b-a791-8930e97f3919)

andyrozman:
I would say that is by design... Closing AAPS shouldn't cancel the bolus, unless user presses the button... 

Now... Good question... Why did you do the "Close All" while bolus was executing... You need to wait until all commands are executed and then you can close your AAPS... 

That small thing there is just showing that command is executed. You can cancel command only while you are in original AAPS instance and when command was issued... AND this is especially true if you are using pump which has no other way of cancelling command (pod pumps without screen and buttons).

vanelsberg:
Yes. By design and this is correct.
> Why did you do the "Close All" while bolus was executing...

I did to show that _when you do_, there is no way to stop the ongoing bolus.
- Think about a user entering a large bolus, then realizing this was wrong and panicking - swipes/closes AAPS to stop it...

When close/open AAPS I still can STOP the bolus from the watch. Can't see any reason why this would not be possible from AAPS?

andyrozman:
- Think about a user entering a large bolus, then realizing this was wrong and panicking - swipes/closes AAPS to stop it...

Yes simple solution, press that little button with "Stop" text on it. I don't see anything wrong with how it works... But that is for Milos to decide...

robertrub:
[...]

Issue #4083, opened 2025-08-02
Feature request: make Bolus iondicator clickable


twinko:
It would be great if the bolus indicator were always clickable and opened the injection progression window, where the bolus could be stopped. This would significantly improve usability, since most users probably don’t know that pressing the back button brings the window back. Especially when SMBs are being delivered and you want to stop them (e.g., when the algorithm isn’t adapting quickly enough), this feature would be very helpful.

Indicator:

![Image](https://github.com/user-attachments/assets/6a12561f-3763-4562-88a7-bb86caf72505)

injection progression window:

![Image](https://github.com/user-attachments/assets/47bafba4-82d2-4366-b024-a7d8b4578f06)

tyrex1975:
Replacing the 'Going to deliver XX.XU' modal with the SMB 'BOLUS X.XXU' main screen footer version, along a clickable Stop on the footer, would be ace.

olorinmaia:
https://github.com/nightscout/AndroidAPS/pull/4229

PR #4229, opened 2025-10-15
Optimize the insulin injection progress dialog functionality.

Angus-repo:
Both of these issues involve similar logic adjustments, so they were modified together.
During testing, it was found that the VirtualPump did not pass the injected insulin into the while loop, which caused the progress dialog title to display 0.0 after killing and reopening the app.
Therefore, this was corrected as part of the same fix.

Please review, thank you.

#4083 

https://github.com/user-attachments/assets/ec1dfa27-a601-4ff5-bfaa-64f8ac37775f

#4201

https://github.com/user-attachments/assets/ba707b34-aeff-468f-b996-b7b14c74077e


vanelsberg:
@Angus-repo Merged this PR with current dev#9fe9498b commit.
Works perfect. Thanks much!

PR #4238, opened 2025-10-19
Add autotune info to About screen

PieterGit:
I tried enabling autotune in the dev build and created `enable_autotune`  in `AAPS/extra` folder as a follow up to https://github.com/nightscout/AndroidAPS/commit/659d14b267a242ec18e82526c31f230ba4284cae

I noticed it was not listed in the About screen so I made this PR to add the information in the About screen. See example here:
image

Unfortuately I don't see the Autotune plugin enabled in the build, so I think there is something more to be done.


andyrozman:
As I understand autotune is plugin... Why should it have extra section here? Don't you need to select it in Config screen (like all other plugins/modules)?

PieterGit:
@andyrozman no it is not just another plugin. It used to be enabled with dev versions if you had a file engineering_mode in the extra directory. Recently that was changed to a specific filename for autotune. I was hoping it will be possible to use autotune with future releases. Perhaps @MilosKozak can indicate why he made the filename check and if it need to be added to the About page or not.

MilosKozak:
`showInList` is probably mandatory. but showing in About is not necessary

MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/24ef11d88d98d6c698047840c65d3e682eb37309

Issue #3984, opened 2025-05-16
AAPS 3.3.3.0_dev_b froze


robertrub:
This happened several times. I look at AAPS and find out that is has frozen during the night.
I thought that it was related to Play Protect but it froze again this morning (a bit before 7 AM) and Play protect is OFF.

As you can see, there was a "catch-up" in BGs when I force closed, cleared cache and restarted AAPS.

![Screenshot_20250516_092543_AAPS.jpg](https://github.com/user-attachments/assets/46e7859c-bc18-4c27-af40-51d162332de5)

The logs after force close and restart.

![AndroidAPS_LOG_1747379627382.log.zip](https://github.com/user-attachments/files/20240118/AndroidAPS_LOG_1747379627382.log.zip)

TheNeskik:
Got an other freeze to this night.

Since last time I enabled everything to log as I don't know what can cause those freeze.

![Screenshot_20250516_063630_AAPS.jpg](https://github.com/user-attachments/assets/1f99de36-7be5-4ba6-a86d-706d699d3dae)

![AndroidAPS_LOG_1747370162514.log.zip](https://github.com/user-attachments/files/20241731/AndroidAPS_LOG_1747370162514.log.zip)

I just updated to latest dev this morning after the freeze, I will see if it happen again

MilosKozak:
> This happened several times. I look at AAPS and find out that is has frozen during the night. I thought that it was related to Play Protect but it froze again this morning (a bit before 7 AM) and Play protect is OFF.
No reason is visible in logs

WesSec:
I experienced the same 2 times on (86fb). Checked the logs and also no reason visible. It didnt crash, it just hung on the latest sate. (it showed last bg was 3 min ago, so upon checking first i didnt notice it was broken). I pulled latest commits and see if i can get a logcat dump when it happens again:

Extras:
Phone: Samsung S21FE, OneUI 6.1 (Android 14)

TheNeskik:
It happened on older commit (f737) but nothing in log. Can't replicate issue as it is random.

I'm on Android 15 (galaxy S23)

robertrub:
`It didnt crash, it just hung on the latest state.`

Same here. No crash, just froze.  No reaction to taps etc.

vanelsberg:
> Same here. No crash, just froze. No reaction to taps etc.
Only AAPS freezes - other apps and phone are then still responsive?

TheNeskik:
[...]

PR #4244, opened 2025-10-20
Changed wait condition, in waitForFinishedThread

HuismanErik:
Changed wait condition, in waitForFinishedThread. 
Change waiting while workIsRunning AND waitingForDisconnect to waiting while workIsRunning OR waitingForDisconnect.

Hopefully this will fix freezing of the app 
See this issue: https://github.com/nightscout/AndroidAPS/issues/3984


HuismanErik:
This doesn't work

PR #4262, opened 2025-10-27
Delay KeepAliveWorker scheduling until app lifecycle is fully started

HuismanErik:
**Problem Description**

After a crash and reinstallation of AndroidAPS, logs showed that the KeepAliveWorker was no longer running.
Analysis revealed that the worker was being scheduled too early during MainApp.onCreate(), before WorkManager and Dagger initialization were fully completed.
This could cause the worker to silently fail to register, leading to missing background tasks and potential startup hangs.

**Solution / Fix Summary**

The scheduling of KeepAliveWorker has been moved to a lifecycle-aware callback, ensuring it starts only after the app is fully initialized and active.
This prevents race conditions with WorkManager setup and guarantees reliable scheduling after app startup.

### DIFF
--- a/app/src/main/kotlin/app/aaps/MainApp.kt
+++ b/app/src/main/kotlin/app/aaps/MainApp.kt
@@ -9,6 +9,8 @@ import android.net.wifi.WifiManager
 import android.os.Build
 import android.os.Handler
 import android.os.HandlerThread
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.ProcessLifecycleOwner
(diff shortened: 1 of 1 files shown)


MilosKozak:
WorkManager should be initialized before `onCreate` is called or?

HuismanErik:
The problem is that KeepAliveWorker was being scheduled too early, before injections and other dependencies were ready. The fix moves the scheduling to a point after the app is fully initialized.

MilosKozak:
So what I mentioned is not true?

HuismanErik:
It might have been a rare timing issue — I only saw it happen once after a reinstall.
If it’s not reproducible, we can probably close this PR

PR #4275, opened 2025-11-02
Date util refactor

Gohtraw:
I was looking for the reason the DST change was a bit messy and refactored DateUtil.


MilosKozak:
Good move but add tests to prove it provides the same results

MilosKozak:
Old files can be moved to test directory, to make them compile only for testing

Gohtraw:
Thanks, will do that tomorrow.
There are also some DST transitions I want to analyse.

PR #4283, opened 2025-11-04
Medtrum Fix Cancel Bolus

### LINKED ISSUE #3490: Medtrum Nano / stop bolus gives error
Last dev. Systematically, if I stop an ongoing bolus, I get a pump unreachable alert.

Either stopping the bolus messes with the BT connection, or triggers a wrong alert message. In both cases, it is a bug.

Logs are below. Check around 18:00

AndroidAPS_LOG_1729699332528.log.zip

Philoul:
Fix #3490

Cancel Bolus by user was returned as Failed, just modify to return a success


MilosKozak:
i don't like to compare comments.
what about to use `medtrumPump.bolusStopped`?

Philoul:
> what about to use `medtrumPump.bolusStopped`?

Much better, thanks!
I was focussed to find the right place to fix the issue, and the root cause, but I saw reading your comment that bolusStopped was dedicated to manual action done by the user.

I just received my new nano pump 2 days ago and I'm not yet very familiar with the nano driver code ;-)

PR #4285, opened 2025-11-05
Equil Maximum basal rate limit 10 U/hr

hhfcvmars:
During the pairing of the Equil pump body, the pump configuration was written via the configuration interface, which limited the maximum basal rate to 1.5 U/hr. As a result, subsequent commands attempting to set a basal rate exceeding 1.5 U/hr were rejected by the pump.


MilosKozak:
this value is overwritten by your AAPS settings. check the code

MilosKozak:
you have to change preferences or change reservoir to apply new settings

PR #4286, opened 2025-11-05
Todo/insulin_concentration

Philoul:
Issue comes from this update within ProfileSealed:
```
override val dia: Double
     get() = if (this is EffectiveProfile) iCfg.insulinEndTime / 1000.0 / 60.0 / 60.0 else error("Requesting DIA on wrong profile")
```

I updated `toPureNsJson` to put the same test before dia access, but I don't know if we can just remove dia within PureNsJson and PureProfile or if there are risks of side effects...

```
    override fun toPureNsJson(dateUtil: DateUtil): JSONObject {
        val o = JSONObject()
        o.put("units", units.asText)
        if (this is EffectiveProfile) o.put("dia", dia)
        o.put("timezone", dateUtil.timeZoneByOffset(utcOffset).id ?: "UTC")
        // SENS
```

### DIFF
--- a/core/data/src/main/kotlin/app/aaps/core/data/model/ICfg.kt
+++ b/core/data/src/main/kotlin/app/aaps/core/data/model/ICfg.kt
@@ -97,7 +97,7 @@ data class ICfg(
         if (bolus.amount != 0.0) {
             val bolusTime = bolus.timestamp
             val t = (time - bolusTime) / 1000.0 / 60.0
-            val td = dia * 60 // always >= MIN_DIA
+            val td = dia * 60 //getDIA() always >= MIN_DIA
             val tp = peak.toDouble()
             // force the IOB to 0 if over DIA hours have passed
             if (t < td) {


MilosKozak:
Well pure profile should not contain dia anymore as it's stored elsewhere and we should handle it consistently across whole app.
I'd recommend to refactor Autotune to handle it the same way and if not possible, add dia to json afterward manually

Philoul:
Here dia was included by the call `toPureNsJson`
It was just not with an `EffectiveProfile` class within ProfileSealed, but a `Pure` class provided by Autotune.

image


This call is done to update, from Autotune, the profile list within ProfilePlugin (update selected profile or add a new profile with AT results)

I will provide soon the Autotune plugin refactoring (work already done between 90% and 100%). 
- I just have to confirm with you the user interface (and link with InsulinPlugin) to manage iCfg selection for Autotune calculation (current iCfg within pump or another one if user want to compare AT results with different iCfg).
- provide the new insulinPlugin/Editor to be able to select (or not) iCfg within Autotune.

Philoul:
in my latest commits, dia has been removed from ProfileSealed and PureProfile, and replaced by iCfg? (not nullable for EffectiveProfile)
- important number of files modified due to dia usage in a lot of places ;-)
- I didn't update ProfilePlugin currently and SimpleProfile (but it's the next step)

Tell me if you are ok with this proposal (I will then fix tests linked to ProfileSealedTest and all other tests using profile...)

PR #4291, opened 2025-11-06
Concentration Helper proposal

Philoul:
Here a proposal to simplify presentation of Insulin (bolus, Reservoir level, SMBs...) or rates using a concentration helper (ch) instead of rh
- if U100 selected, nothing change within strings
- if another concentration is used then both values are presented (with U and CU)

I provided example of usage within Virtual Pump fragment, Insight Pump Fragment and FillDialog

Please comment ConcentrationHelper docjava and functions proposed (here I started from my previous branch but probably some functions should be reworked due to new architecture ;-)

### DIFF
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/rx/events/EventOverviewBolusProgress.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/rx/events/EventOverviewBolusProgress.kt
@@ -1,7 +1,9 @@
 package app.aaps.core.interfaces.rx.events
 
 import app.aaps.core.interfaces.R
+import app.aaps.core.interfaces.insulin.ConcentrationHelper
 import app.aaps.core.interfaces.pump.BolusProgressData
+import app.aaps.core.interfaces.pump.PumpInsulin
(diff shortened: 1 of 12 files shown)


MilosKozak:
I'd not mention concentrated units anywhere in UI except 
* pump driver fragment - real values
* dialogs where we change units

this will help keep the code clean and UI will match the code

Philoul:
In my latest commit, I simplified ConcentrationHelper to only provide strings within Pump Drivers (with IU and CU), and FillDialog (U with Volume in µl)

MilosKozak:
I'd not display concentrated amount in BolusProgress. It's AAPS core which should present only U100

Philoul:
> I'd not display concentrated amount in BolusProgress. It's AAPS core which should present only U100

Agree, I updated the code to manage CU->IU convertion (using PumpInsulin class) to convert delivered amount value provided by pump drivers (value in CU) to BolusProgress dialog (should be IU for presentation and percentage calculation) 🤔

Philoul:
Will rework ConcentrationHelper for only UserInterface presentation (in `FillDialog` with insulin volume, `EventOverviewBolusProgress` for iU presentation or in pump fragments)

Issue #3657, opened 2024-12-26
[Dev] Timestamps in SMB log is shown in milliseconds


olorinmaia:
![image](https://github.com/user-attachments/assets/0021b849-cca7-4e42-8ca0-63ce1aabe2b1)

![image](https://github.com/user-attachments/assets/e7c08565-4291-4ca3-a748-0b35290c2a05)

![image](https://github.com/user-attachments/assets/1589dea0-006b-4d29-8773-a0601b55eab7)

olorinmaia:
Fixed in dev

PR #4295, opened 2025-11-09
DisplayAsDate to make ms since epoch timestamps human-understandable

Gohtraw:
Created an annotation to make timestamps more intuitive ( https://github.com/nightscout/AndroidAPS/issues/3657 )
@olorinmaia please tell me if I missed a spot.


MilosKozak:
as this extends previous function I'd just replace `dataClassToHtml`

olorinmaia:
@Gohtraw Thanks for looking into this. I've merged your PR into latest dev and from what I can see Meal data section is fixed.
"IOB data" and "Result" are still showing timestamp in milliseconds on my test phone.

MilosKozak:
whay is necessary to have both: dataClassToHtml and dataClassToHtmlDisplayAsDate ?

MilosKozak:
closed for inactivity

Issue #4192, opened 2025-10-02
Preselection of hypo TT in carb UI


widmann:
Recently, I went on a day-long cycling tour and set a 70% profile with an activity TT for 10 hours. After two hours I got slightly low, so I ate some carbs, and entered them. Unfortunately, I didn't notice the preselected "start hypo TT". There was no problem for the next hour, but then the target was reset to the default (instead of activity TT). The loop gave me too much insulin, and I had a severe hypo.

Suggestion: Do not preselect "Start hypo TT" if the currently set TT is *higher and longer* than the hypo TT would be.

gerison77:
I recently encountered the same issue. In a caretaker configuration, a low profile combined with a high temporary target (TT) set 1.5 hours prior to exercise caused an interaction problem.
The caretaker treated the low with carbohydrates, which effectively shortened the TT duration. When the TT expired, the system reverted to the default target, thereby blocking the expected automation from being triggered.

olorinmaia:
When low, I think Hypo TT should automatically be selected when entering carbs, unless:
There is already a high TT > 100 mg/dl or 5,5 mmol/L with a duration longer than preset Hypo TT duration.

This will avoid Hypo TT cancelling out Activity TT or custom high TT with a longer duration than preset Hypo TT. 

We experience this issue from time to time with Hypo TT cancelling out Activity TT when entering carbs when low. I think this proposal will fix issue as described in this ticket and other scenarios. It's recommended to not allow SMB when a high TT is set (above 100mg/dl)

olorinmaia:
@MilosKozak @widmann this issue can be closed due to merging of https://github.com/nightscout/AndroidAPS/pull/4339

widmann:
Thank you for providing the patch! It is much appreciated.

olorinmaia:
> Thank you for providing the patch! It is much appreciated.

You are welcome and thanks for reporting this issue. I also helped myself as we experience this from time to time.

Feel free to close the issue 😉

PR #4339, opened 2025-11-20
Improve auto-check-HypoTT in CarbsDialog when low

olorinmaia:
Fixes and improves upon issue described in https://github.com/nightscout/AndroidAPS/issues/4192

When low, I think Hypo TT should automatically be selected when entering carbs, unless:
There is already an active high TT > 100 mg/dl or 5,5 mmol/L with a duration longer than preset Hypo TT duration.

This will avoid Hypo TT cancelling out Activity TT or custom high TT with a longer duration than preset Hypo TT.

We experience this issue from time to time with Hypo TT cancelling out Activity TT when entering carbs when low. I think this proposal will fix issue as described in the mentioned issue and other scenarios. It's recommended to not allow SMB when a high TT is set (above 100mg/dl)

Built and tested OK.

Some screenshots when BG is low and opening Carbs Dialog:
image
image
image
image

### DIFF
--- a/core/data/src/main/kotlin/app/aaps/core/data/configuration/Constants.kt
+++ b/core/data/src/main/kotlin/app/aaps/core/data/configuration/Constants.kt
@@ -10,6 +10,9 @@ object Constants {
     const val defaultDIA = 5.0
     const val notificationID = 556677
 
+    // OpenAPS algorithm
+    const val ALLOW_SMB_WITH_HIGH_TT = 100
+
     // SMS COMMUNICATOR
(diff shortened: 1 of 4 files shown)


olorinmaia:
@MilosKozak Please review again when you got time.

I've tested that new functionality still works in Carbs Dialog when low.
Also tested that `allowSMB_with_high_temptarget` works as before after replacing value 100 with the new `Constants.ALLOW_SMB_WITH_HIGH_TT` in DetermineBasalSMB.kt and CarbsDialog.kt

PR #4347, opened 2025-11-20
Hide InsulinAge pref for Patch pump

Philoul:
Hide Insuline Age thresholds in overview preferences when a patch pump is used


MilosKozak:
when we hide preferences ... did you check, that the code using it will not act?

Philoul:
No, you are right...
I just saw statusLight hide InsulinAge everywhere, that's why I proposed to hide preferences.
Will check iage usage within AAPS and comment here (or update/close this PR 😉)

Philoul:
I checked quickly and for me it's fine.
keys are only used (get) in StatusLightHandler (Insulin age views hidden with patched pumps).

PR #4392, opened 2025-12-07
Fix NS Sync for SiteRotation

### LINKED ISSUE #4284: Site Management Sync issue on NS
Site is managed as an update of Therapy Event
- NS Sync works immediatly for the first event (without site location) but not later when site has been updated
- if Phone is offline during the site change, and then internet connection is enabled, then evenrything is correctly uploaded to NS

Will see if I can fix this issue with an event sent to NS plugin on location or note update of the TE to force NS upload...

Philoul:
Site rotation is managed as un update of a therapy Event
Force a new NS upload in SiteRotationDialog
Force database Sync on Location, Arrow or Note update
Trial to Fix #4284


Philoul:
Upload in NS works fine.
Not sure for Synchronization in remote device (I see in NSV3 the updated TherapyEvent downloaded in remote), I thought it worked on first tests but stopped after a full sync...

Philoul:
Now it works with my latest commit...

PR #4393, opened 2025-12-07
Fix insulin concentration conflicts with dev

Philoul:
I tried to fix conflicts between insulin_concentration and dev branch

Could you please review this proposal (a bit trick for me especially within `persistenceLayerImpl`)


Philoul:
Due to the number of files modified, probably better to manage the conflicts on your side and then compare with my proposal...

MilosKozak:
do you trust it or should i do it by myself?

Philoul:
I trust the files I modified on my side, but it was tricky for me on other files modified by you on both sides
- to prepare insulin concentration on one hand
- updates done on main dev branch on the other hand
(Mainly within persistance layer, persistanceLayerImpl, and the corresponding calls).

Build works and I didn't got crashed on my tests, but I'm not 100% sure I'm on latest target solution...

Philoul:
So you can fix conflicts quickly on your side without taking care of my work and I can update if something linked to insulin management is missing 😉.

PR #4397, opened 2025-12-08
feat: Expand insulin pump compatibility by serial number prefix

hhfcvmars:
Add serial number-based support check for insulin pumps. Only pumps with serial numbers starting with '0', '1', '3', 'A', or 'D' require firmware version validation. All other pump models are now supported by default.


MilosKozak:
which pumps are affected and need this change?

hhfcvmars:
> which pumps are affected and need this change?

I have confirmed with their developers that this configuration supports all pump models, and the serial numbers comply with their naming conventions.

PR #4425, opened 2025-12-19
Always Show Cannula Usage

Philoul:
I think this information is usefull even with Patch Pump, so I propose to make it always visible.
On my side I manually record the overall usage in Site rotation Note, and overall usage is not always the Insulin quantity put into Reservoir of patch pump...


olorinmaia:
Can you illustrate what this means as a dash user?

Philoul:
@olorinmaia 


In this screenshot, 161U has been delivered by the patch.

With tube pumps, cannula age and insulin age can be different but cannula should also be changed every 3 days.

PR #4439, opened 2025-12-24
insulin_concentration bug_fix

Philoul:
- Fix Progressbar missing in BolusProgressDialog (due to wrong `detailedBolusInfo.id`)
- ~~Fix missing convertions in PumpWithConcentration~~ (I finally propose to replace Double by PumpInsulin)

Note that we still miss in EventOverviewBolusProgress the convertion of `delivered` value currently sent as `Double` from pump driver and not sent as `PumpInsulin`
- Progress bar works, but with wrong calculated percentage (with U200, bolus is finished once 50% has been delivered, so correct amount in CU but not converted to IU and wrong percentage)
- I will have to replace `Double` value by `PumpInsulin` value in  EventOverviewBolusProgress constructor and provide concentration to manage convertion CU -> IU included in PumpInsulin object to fix BolusProgressDialog with `delivered` sent...
 
My plan is to use ch (concentrationHelper) for everything related to userInterface (at least to provide concentration to be able to manage convertion and standardize presentation everywhere...


MilosKozak:
@Philoul this has a conflict now

PR #4470, opened 2026-01-04
Hide AutotuneAutoSwitchProfile option in released branch

Philoul:
Autotune Automation rule is only available in Dev + Engineering mode, so we should hide option to "AutoSwitch" on Autotune enabled in Released version


MilosKozak:
fix this in compose if needed

PR #4475, opened 2026-01-06
Update FillDialog

Philoul:
Provide information of Volume of insulin in µl
- Show requested Prime/Fill standard value (U100) and corresponding insuline volume
- if other concentration is used, then show a warning with a correction factor with updated Insulin value (IU) and corresponding volume
image


MilosKozak:
@Philoul remember all UI changes. it will be necessary to do it again for 4.0

PR #4480, opened 2026-01-10
Fix BolusProgress on Watch

### LINKED ISSUE #4472: [3.4] Bolus Progress notification on Wear does not show total amount of insulin to be delivered
Due to Bolus Progress refactoring in ab620d32 for 3.4 the string below was removed and replaced with a string only showing what has been delivered and not the total amount to be delivered. The code for several pumps were also adjusted.

<string name="bolus_delivered_so_far">%1$.2f U / %2$.2f U delivered</string>

Problem is that on watch you then don't see the total amount to be delivered.

Screenshot below.

Image

I've fixed this myself with this commit, this is simply reverting the change from Bolus Progress refactoring:
https://github.com/nightscout/AndroidAPS/commit/e278382b7adc5dc2c73768a7b37652ee95d0accb

Philoul:
Add missing Total amount on Wear Watch BolusProgress Notification
BolusProgressDialog on phone unchanged
Dash  driver use now EventOverviewBolusProgress for bolus delivery percentage
Fix #4472 

Same bolus on the 2 screenshots below.
image

image

@olorinmaia could you check with Dash pump if now issue is fixed (I updated Dash Driver to use EventOverviewBolusProgress for bolus progress string (instead of creating the string locally into Dash Driver)

### DIFF
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/pump/BolusProgressData.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/pump/BolusProgressData.kt
@@ -20,6 +20,7 @@ object BolusProgressData {
         bolusEnded = false
         stopPressed = false
         status = ""
+        wearStatus = ""
         percent = 0
     }
 
@@ -48,6 +49,7 @@ object BolusProgressData {
      * Last received status update
      */
     var status = ""
+    var wearStatus = ""
     var percent = 0
 
     var bolusEnded = false
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/rx/events/EventOverviewBolusProgress.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/rx/events/EventOverviewBolusProgress.kt
@@ -8,12 +8,13 @@ import kotlin.math.min
 /**
  * Custom status message and percent
  */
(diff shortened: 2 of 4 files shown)


olorinmaia:
I've tested this one day now with Dash and seem to be working fine and a good solution for my issue. I recommend patching this to master.

My only feedback is that I think this should be changed from:
```
    <!--    Events-->
    <string name="bolus_delivering">Delivering %1$.2fU</string>
    <string name="bolus_delivered_so_far">%1$.2fU / %2$.2fU delivered</string>
    <string name="bolus_delivered_successfully">Bolus %1$.2fU delivered successfully</string>
```

To:
    
```
     <!--    Events-->
    <string name="bolus_delivering">Delivering %1$.2f U</string>
    <string name="bolus_delivered_so_far">%1$.2f U / %2$.2f U delivered</string>
    <string name="bolus_delivered_successfully">Bolus %1$.2f U delivered successfully</string>
```

There could be reasons you don't want a space before "U" but at least we can fix this with translations in Norwegian to what we prefer. So not a big deal!

Philoul:
Here I removed the space to be sure the "U" of total amount will not be in a new row.

These spaces should normally be a "Non breaking space" `&#160;` everywhere in all strings we manage values with units...
I propose to update with the non breaking space here let's see later how we manage all the other strings in AAPS...

Issue #4462, opened 2026-01-02
[Wear] [Enhancement] QuickBolus Activity


mcnelson:
Hello all,

Thank you for this brilliant piece of software. I've enjoyed using AAPS on Wear in particular, although I think there is a big opportunity for feature development. One such area is administering boluses manually. I decided to develop a feature called QuickBolus. It's a one-screen carbs + insulin UI optimized for a tiny circular watch screen. It's a launcher activity so that it can be bound to hardware buttons on a watch. I'd love to pull this back into the project if there is interest. It's become indispensable for me, although I realize everyone manages T1D differently and this may not align with AAPS' design goals. Zero hard feelings either way!

I made a scrappy demo: https://vimeo.com/1150900096

I've chipped away at this for about a month, and am very open to feedback. Happy to open PRs if it's a 👍.

Philoul:
> I made a scrappy demo: https://vimeo.com/1150900096

Can you please share your video another way (impossible to see it without connecting to Vimmeo ???). Or just share screenshots here, probably the best way to understand your proposal...
- We already have a Tile for Bolus + Carbs from Watch (can you compare your proposal with current existing one)?
- Also, please remember that we cannot offer the QuickBolus feature without guaranteeing 100% compliance with security constraints (bolus constraints respected, user confirmation, which cannot be performed accidentally or unintentionally...).

mcnelson:
Of course, here's a gif. Apologies for emulator lag / not a clear circle viewport..

![Image](https://github.com/user-attachments/assets/ff0d107c-3f9f-4adf-a026-860b713692db)

The intent is to offer a more streamlined option to complement (not replace) the existing Wear treatment flow:

- Dial up/down units in steps of 0.5 (warns when exceeding max bolus limit)
- Toggle 'eating' to auto-calculate carbs from units × IC ratio
- Manual carbs override with up/down buttons in steps of 5g
- Shows what IOB will be with selected dose (indispensable for my use case)
- Hold center button for 3s to commit - progress ring fills during hold. Must keep finger on button.
- Ring then shows current bolus progress, or activity closes if recording carbs only
- Displays wizard suggestion as colored value (e.g., 0.7 U in cyan); snaps to suggestion when dialing past it
- Carb deficit shown as yellow checkbox with pre-filled grams
 
[...]

PR #4484, opened 2026-01-11
Wear: Display New IOB in Wizard Result if IOB is used in calculations

olorinmaia:
Inspired from: https://github.com/nightscout/AndroidAPS/issues/4462

We often look at calculations on Wear to see current IOB to make sure when adding the calculated bolus from Bolus Wizard that the new IOB doesn't exceed a certain amount.

I've made this improvement to the Wizard Result in Wear. It will only show if IOB is used in the calculations.

Some screenshots.

**IOB "positive"**
Current IOB: 0,58
Calculated Bolus: 0,44
New IOB after Bolus: 0,98

| Current IOB | Result with new IOB |
|--------|--------|
| image | image | 

**IOB "negative"**
Current IOB: -0,13
Calculated Bolus: 0,90
New IOB after Bolus: 0,77

| Current IOB | Result with new IOB |
|--------|--------|
| image | image | 

**When IOB is not used in calculations:**
image

### DIFF
--- a/wear/src/main/kotlin/app/aaps/wear/interaction/actions/WizardResultFragment.kt
+++ b/wear/src/main/kotlin/app/aaps/wear/interaction/actions/WizardResultFragment.kt
@@ -177,7 +177,7 @@ class WizardResultFragment : Fragment() {
         // New IOB = Total Insulin - Current IOB (because IOB is subtracted in wizard)
         // Only calculate if IOB was actually used (totalIob is not NaN means it was included)
         val newIob = if (!totalIob.isNaN()) {
-            totalInsulin - totalIob
+            totalInsulin - totalIob // totalIob is negative so this is addition
         } else {
             null  // Don't show New IOB if IOB wasn't used in calculation
         }


wear/src/main/kotlin/app/aaps/wear/interaction/actions/WizardResultFragment.kt
```diff
@@ -173,14 +173,32 @@ class WizardResultFragment : Fragment() {
 
         builder.addCalculationRow(calculationRowsContainer, WizardCalculationRow(getString(R.string.wizard_result_total), totalInsulin))
 
+        // Calculate new IOB if IOB was used in calculation
+        // New IOB = Total Insulin - Current IOB (because IOB is subtracted in wizard)
+        // Only calculate if IOB was actually used (totalIob is not NaN means it was included)
+        val newIob = if (!totalIob.isNaN()) {
+            totalInsulin - totalIob
```
mcnelson: Tiny suggestion-- to an outsider like me, this looks weird. This is addition because totalIob is negative, correct? Maybe comments could be more specific?

```suggestion
            totalInsulin - totalIob // totalIob is negative so this is addition
```

olorinmaia: I agree it's abit confusing, but you are correct. I tried explaining the reason why the math must be like this with this comment:
// New IOB = Total Insulin - Current IOB (because IOB is subtracted in wizard)

totalIob is a positive number if IOB in AAPS is negative.. to make it more confusing, but you get the idea here. I'm not sure if your comment is better than what I already use.. But I agree it can be improved.

PR #4467, opened 2026-01-03
Omnipod Dash Drift detection

nl-ruud:
# Basal Drift Detection for Omnipod Dash

## Issue
The insulin delivered by the Omnipod Dash drifts from the programmed basal rate (related to #4158). This PR adds tracking to quantify the drift for future correction (will be a separate PR).

> ⚠️ First PR for AndroidAPS and GitHub.

## Description
Add basal drift tracking to the Omnipod Dash driver.

### Changes
- **Separate basal and bolus delivery**
  - Track bolus pulses separately
  - Calculate basal pulses: `total pulses − bolus pulses`
- **Expected basal delivery calculation**
  - Integrates expected delivery over time
  - Accounts for hourly basal rate changes and temporary basals
- **Drift computation**
  - `basalDrift = actual − expected`
  - Logged on each status update (`DefaultStatusResponse` and `AlarmStatusResponse`)

#### Example log
```text
PUMP_BASAL act=8,10U (tot=33,25U bol=25,15U) exp=9,42U err=-1,32U dErr=-0,02U
```

**Log fields:**

| Field | Description |
|------|------------|
| `act` | Actual basal units delivered (`tot` − `bol`) |
| `tot` | Total units delivered |
| `bol` | Bolus units delivered |
| `exp` | Expected basal units delivered |
| `err` | Error / cumulative basal drift (`act` - `exp`) |
| `dErr` | Delta of `err` on this pump status response |

### New State Fields
- `bolusPulsesDelivered`: cumulative bolus pulses
- `basalExpected`: running total of expected basal delivery

### New Computed Properties
- `basalPulsesDelivered` = total pulses − bolus pulses
- `basalDelivered` = basal pulses delivered converted to insulin units
- `basalDrift` = actual − expected insulin units (positive = over-delivery, negative = under-delivery)

## Tests
- Confirmed `basalDrift` matches manual calculations
- Confirmed calculation on a temp basal change
- Confirmed calculation on a basal program change
- Validated calculations with Nightscout 'day to day' report: bolus insuline equals `ΔbolusPulsesDelivered * 0.05`, basal insuline is exactly `ΔbasalDrift` off from `ΔbasalDelivered`. Same difference between TDD and `ΔtotalPulsesDelivered`. QED.

Issue #4158, opened 2025-09-15
Too frequent basal updates may cause basal insulin delivery problems with Omnipod Dash


SetaHemuli:
- AAPS version 3.3.1.3
- Using closed loop with SMB enabled
- Pump: Omnipod Dash
- Glucose sensor: FSL3
- BG values coming every 1 minute from Juggluco

**Problem description:**

Too frequent basal rate changes may cause basal insulin delivery problems when using Omnipod Dash pump (may also affect other pump models).

Omnipod Dash uses a timer for giving small 0.05u doses of basal, with an interval defined by set basal rate. For example, basal rate of 0.6u/h results in 0.05u dose of insulin given every 5 minutes. Dose is given at the end of each timer interval, ie. if basal rate is set to 0.6u/h, the first dose will be given after 5 minutes have elapsed since setting basal rate.

When using 1-minute blood glucose values from Juggluco, AAPS loop updates once every minute. This may lead to basal rate changes happening once a minute. Every time basal rate is updated to the Dash, also the timer on the pump is reset to zero. If basal rate is changed too often, it may lead to a situation where zero basal is actually delivered, since the pump timer keeps resetting.

I have verified that insulin delivery from the Dash indeed stops by changing basal rates too often. I disabled AAPS Loop, and manually changed temp basal rates from AAPS about once every 30 seconds. Basal rates were changed from 0.8U/h to 3.0U/h. By continuously switching the basal rate, no insulin was delivered for a full 10 minutes (verified by listening for the "click" sound the pump makes when delivering 0.05U).

**Suggestions for improvement:**

- Disable 1-minute Loop updates when using Omnipod Dash pump. Only allow loop updates / basal rate changes once every 5 minutes
- Include a warning in AAPS documentation which describes that 1-minute BG updates may lead to incorrect basal delivery, particularly when using Omnipod Dash pump

planiitis:
If I understand the issue correctly, then limiting updates to once every 5 minutes will still result in a basal rate deficiency of up to 0.05 U every five minutes = 0.6 U/h.

It seems to me the correct approach would be to keep an accumulator in the pump driver that increases by the amount of basal insulin missing after every basal rate change (calculated from the pump's reported total amount delivered vs the value and duration of the last basal rate, subtracting any bolus of course) and inserting a bolus of 0.05 U each time this accumulator passes 0.05 U.

nl-ruud:
[...]

PR #4499, opened 2026-01-18
Omnipod Dash Drift resolution

nl-ruud:
## Omnipod Dash basal drift resolution

This pull request introduces a new "basal correction" feature for the Omnipod Dash pump integration. The main goal is to detect under-delivery of basal insulin and automatically trigger a small correction bolus when needed. The implementation tracks delivered basal and bolus pulses, calculates expected vs. actual insulin delivery, and manages correction logic with safety checks and cooldowns. 

## Recap of the issue

The Omnipod Dash pump exhibits behavior that causes it to deliver less basal insulin than AAPS expects (issue #4158). This is effectively a hardware limitation of the pump.

The Dash uses an internal timer to determine when a basal pulse of 0.05 U is delivered. Once the timer interval elapses, the pulse is delivered. However, this timer is restarted whenever a basal rate change occurs (and possibly also when a bolus or SMB is delivered).

When used in combination with looping, this leads to structural under-delivery of basal insulin, as the algorithm adjusts the basal rate frequently. While the Medtrum pump compensates for this behavior internally, the Dash does not, and neither does the current Dash driver implementation in AAPS.

The issue is most apparent during the night. During daytime operation, SMBs often result in a basal rate of 0, which masks the effect. In observed usage, this results in approximately 10% of the expected Total Daily Dose (TDD) not being delivered over a 24-hour period. Additionally, glucose targets are often not reached overnight, particularly after meals with prolonged glucose impact (e.g. pasta).

## Approach

This pull request introduces automatic basal drift compensation by correcting the pump lag with small correction boluses. A correction bolus of 0.05 U is delivered as soon as the detected basal deficit reaches 0.025 U. Several safety constraints are applied. For example, corrections are not delivered during a temporary basal of 0 unless that temporary basal is the direct result of a recently delivered bolus.

With this approach, the deviation between expected and delivered basal insulin is kept within a bounded range of approximately −0.050 U to +0.025 U.

With this fix applied:
- The TDD reported by AAPS (under *Statistics*) matches the total amount actually delivered by the pump (Dash menu → *Pod management* → *History*).
- Overnight glucose targets are consistently reached.

Without this fix, a clear discrepancy between reported and delivered TDD can be observed.

### Basal Correction Feature

- Added a new custom command, `CommandDeliverBasalCorrection`, and integrated it into the command queue and handling logic in `OmnipodDashPumpPlugin` to trigger a basal correction bolus when under-delivery is detected.
- Implemented the `deliverBasalCorrection` method to handle the actual delivery of the correction bolus, including safety checks (reservoir level, bolus in progress, cooldowns, etc.).

### Basal Delivery Tracking and Drift Calculation

- Added logic in `OmnipodDashPodStateManagerImpl` to track basal and bolus pulses separately, compute actual delivered basal, and calculate "drift" (difference between expected and actual basal delivery). This includes new fields in the pod state and methods for integrating expected delivery over time. 
- The `needsBasalCorrection` method determines when a correction is needed based on drift, cooldowns, and other safety criteria.

### Pod State Update and Logging Refactor

- Refactored pod state update logic to centralize updates in a new `updatePodState` method, which also updates basal tracking fields and logs basal delivery/drift for debugging. This method is now used in both default and alarm status response handlers and removes code duplication from these methods. 

Logging example:

```text
PUMP_BASAL act=8,10U (tot=33,25U bol=25,15U) exp=9,42U err=-1,32U dErr=-0,02U
```

Legend for the log fields:

| Field | Description |
|------|------------|
| `act` | Actual basal units delivered (`tot` − `bol`) |
| `tot` | Total units delivered |
| `bol` | Bolus units delivered |
| `exp` | Expected basal units delivered |
| `err` | Error / cumulative basal drift (`act` - `exp`) |
| `dErr` | Delta of `err` on this pump status response |

### Interface and State Enhancements

- Updated the `OmnipodDashPodStateManager` interface and its implementation to support new fields and methods for basal correction tracking (`lastBasalCorrectionTime`, `basalCorrectionInProgress`, `needsBasalCorrection`). 

These changes together enable the system to automatically detect and correct small basal under-deliveries, improving insulin delivery reliability and safety.

## Tests Performed

To verify the fix, we compared the total insulin delivered at 00:00 at the start of a day with the total at 00:00 at the end of the same day. 

Before this fix, the difference consistently resulted in a lower number than the TDD reported by AAPS, confirming under-delivery. After applying the fix, this discrepancy no longer occurs: AAPS TDD matches the pump’s delivered total.

### DIFF
--- a/app/src/main/kotlin/app/aaps/implementations/ConfigImpl.kt
+++ b/app/src/main/kotlin/app/aaps/implementations/ConfigImpl.kt
@@ -44,6 +44,7 @@ class ConfigImpl @Inject constructor(
     private var ignoreNightscoutV3Errors: Boolean? = null
     private var doNotSendSmsOnProfileChange: Boolean? = null
     private var enableAutotune: Boolean? = null
+    private var enableOmnipodDriftCompensation: Boolean? = null
     private var disableLeakCanary: Boolean? = null
 
     override fun isEngineeringModeOrRelease(): Boolean = if (!APS) true else isEngineeringMode() || !isDev()
@@ -54,5 +55,6 @@ class ConfigImpl @Inject constructor(
     override fun ignoreNightscoutV3Errors(): Boolean = ignoreNightscoutV3Errors ?: (fileListProvider.get().ensureExtraDirExists()?.findFile("ignore_nightscout_v3_errors") != null).also { ignoreNightscoutV3Errors = it }
     override fun doNotSendSmsOnProfileChange(): Boolean = doNotSendSmsOnProfileChange ?: (fileListProvider.get().ensureExtraDirExists()?.findFile("do_not_send_sms_on_profile_change") != null).also { doNotSendSmsOnProfileChange = it }
     override fun enableAutotune(): Boolean = enableAutotune ?: (fileListProvider.get().ensureExtraDirExists()?.findFile("enable_autotune") != null).also { enableAutotune = it }
+    override fun enableOmnipodDriftCompensation(): Boolean = enableOmnipodDriftCompensation ?: (fileListProvider.get().ensureExtraDirExists()?.findFile("omnipod_drift_compensation") != null).also { enableOmnipodDriftCompensation = it }
     override fun disableLeakCanary(): Boolean = disableLeakCanary ?: (fileListProvider.get().ensureExtraDirExists()?.findFile("disable_leakcanary") != null).also { disableLeakCanary = it }
 }
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/configuration/Config.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/configuration/Config.kt
@@ -34,6 +34,7 @@ interface Config {
     fun ignoreNightscoutV3Errors(): Boolean
     fun doNotSendSmsOnProfileChange(): Boolean
     fun enableAutotune(): Boolean
+    fun enableOmnipodDriftCompensation(): Boolean
 
     /**
      * Disable LeakCanary (memory leaks detection). By default it's enabled in DEBUG builds.
--- a/pump/omnipod/common/build.gradle.kts
+++ b/pump/omnipod/common/build.gradle.kts
@@ -24,6 +24,8 @@ dependencies {
     api(libs.androidx.navigation.fragment)
     api(libs.com.google.android.material)
 
+    testImplementation(project(":shared:tests"))
+
     ksp(libs.com.google.dagger.compiler)
     ksp(libs.com.google.dagger.android.processor)
 }
--- a/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManager.kt
+++ b/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManager.kt
@@ -138,6 +138,10 @@ interface OmnipodDashPodStateManager {
 
     fun updateLowReservoirAlertSettings(lowReservoirAlertEnabled: Boolean, lowReservoirAlertUnits: Int): Completable
 
+    var lastBasalCorrectionTime: Long?
+    var basalCorrectionInProgress: Boolean
+    fun needsBasalCorrection(): Boolean
+
     data class ActiveCommand(
         val sequence: Short,
         val createdRealtime: Long,
--- a/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
+++ b/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
@@ -2,6 +2,7 @@ package app.aaps.pump.omnipod.common.bledriver.pod.state
 
 import android.os.SystemClock
 import app.aaps.core.data.model.BS
+import app.aaps.core.interfaces.configuration.Config
 import app.aaps.core.interfaces.logging.AAPSLogger
 import app.aaps.core.interfaces.logging.LTag
 import app.aaps.core.interfaces.rx.bus.RxBus
@@ -43,7 +44,8 @@ import javax.inject.Singleton
 class OmnipodDashPodStateManagerImpl @Inject constructor(
     private val logger: AAPSLogger,
     private val rxBus: RxBus,
-    private val preferences: Preferences
+    private val preferences: Preferences,
+    private val config: Config
 ) : OmnipodDashPodStateManager {
 
     private var podState: PodState
@@ -322,6 +324,8 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
     }
 
     override fun needsBasalCorrection(): Boolean {
+        if (!config.enableOmnipodDriftCompensation()) return false  // Semaphore file check
+
         val correctionThreshold = -PodConstants.POD_PULSE_BOLUS_UNITS / 2  // -0.025U
         
         if (!isActivationCompleted) return false  // Don't correct during activation/priming
@@ -735,6 +739,7 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
     }
 
     override fun onStart() {
+        logger.info(LTag.PUMP, "Omnipod Dash drift compensation: ${if (config.enableOmnipodDriftCompensation()) "enabled" else "disabled"}")
         when (getCommandConfirmationFromState()) {
             CommandConfirmationSuccess, CommandConfirmationDenied -> {
                 val now = SystemClock.elapsedRealtime()
--- a/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
+++ b/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
@@ -48,7 +48,8 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
     private val config: Config
 ) : OmnipodDashPodStateManager {
 
-    private var podState: PodState
+    /** Internal (rather than private) to allow unit testing within this module. */
+    internal var podState: PodState
 
     init {
         podState = load()
@@ -261,9 +262,24 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
         // Compute drift (actual - expected: positive = over-delivery, negative = under-delivery)
         get() = basalDelivered - (podState.basalExpected ?: basalDelivered)
 
-    private fun integrateExpectedDelivery(startTime: Long, endTime: Long): Double? {
+    /**
+     * Integrates expected basal delivery (in units) over [startTime, endTime].
+     * Internal (rather than private) to allow unit testing within this module.
+     */
+    internal fun integrateExpectedDelivery(
+        startTime: Long,
+        endTime: Long,
+        timeZoneOffset: Int?,
+        tempBasal: OmnipodDashPodStateManager.TempBasal?,
+        basalProgram: BasalProgram?
+    ): Double? {
         logger.debug(LTag.PUMP, "integrateExpectedDelivery: period ${(endTime - startTime) / 1000.0}s")
-        
+
+        if (timeZoneOffset == null) {
+            logger.error(LTag.PUMP, "integrateExpectedDelivery: pod timezone unknown, skipping")
+            return null
+        }
+
         // Validate time period
         if (startTime > endTime) {
             logger.error(LTag.PUMP, "Invalid time period: startTime=$startTime > endTime=$endTime")
@@ -284,7 +300,7 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
         // Add basal program segment boundaries
         basalProgram?.segments?.forEach { segment ->
             // Calculate day boundaries in pod's local timezone, not UTC
-            val timeZoneOffset = podState.timeZoneOffset ?: 0
+
             val dayStartLocal = ((startTime + timeZoneOffset) / 86400_000L) * 86400_000L - timeZoneOffset
             var segmentStart = dayStartLocal + segment.startSlotIndex.toLong() * 30 * 60_000L
             
@@ -299,6 +315,10 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
                 segmentStart += 86400_000L
             }
         }
+
+        // Offset to convert UTC epoch to pod-local time for rateAt (which uses device timezone internally)
+        val deviceOffsetMs = TimeZone.getDefault().getOffset(startTime)
+        val podTimeAdjustmentMs = timeZoneOffset - deviceOffsetMs
         
         // Integrate over each segment
         return boundaries.sorted().windowed(2).mapIndexed { index, (boundaryStart, boundaryEnd) ->
@@ -309,7 +329,7 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
             val rate = tempBasal?.let { tb ->
                 val tempBasalEnd = tb.startTime + tb.durationInMinutes * 60_000L
                 tb.rate.takeIf { segmentMid in tb.startTime until tempBasalEnd }
-            } ?: basalProgram?.rateAt(segmentMid) ?: return null  // Abort if rate unknown
+            } ?: basalProgram?.rateAt(segmentMid + podTimeAdjustmentMs) ?: return null  // Abort if rate unknown
             
             val delivery = rate * segmentHours
             
@@ -339,7 +359,9 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
         // Compute drift once for efficiency
         val drift = basalDrift
 
-        // Reset if drift exceeds boundaries (over-delivery or severe under-delivery)
+        // Reset if drift exceeds boundaries (over-delivery or severe under-delivery).
+        // Thresholds are intentionally tight: a reset is preferred over risking over-correction.
+        // Even so, reaching this point is genuinely exceptional and has not been observed in practice; widen if needed.
         if (drift >= PodConstants.POD_PULSE_BOLUS_UNITS * 2 || drift <= -PodConstants.POD_PULSE_BOLUS_UNITS * 2) {
             logger.warn(
                 LTag.PUMP,
@@ -703,7 +725,8 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
         store()
     }
 
-    private fun calculateBolusPulseIncrease(
+    /** Internal (rather than private) to allow unit testing within this module. */
+    internal fun calculateBolusPulseIncrease(
         previousTotalPulses: Short,
         newTotalPulses: Short,
         previousBolusPulsesRemaining: Short?,
@@ -768,7 +791,8 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
         }
     }
 
-    private fun updatePodState(
+    /** Internal (rather than private) to allow unit testing within this module. */
+    internal fun updatePodState(
         deliveryStatus: DeliveryStatus,
         podStatus: PodStatus,
         totalPulsesDelivered: Short,
@@ -790,9 +814,13 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
 
         // Update basal expected delivery
         podState.basalExpected = podState.basalExpected?.let {
-            integrateExpectedDelivery(podState.lastUpdatedSystem, now)?.let { delta ->
-                it + delta
-            }
+            integrateExpectedDelivery(
+                startTime = podState.lastUpdatedSystem,
+                endTime = now,
+                timeZoneOffset = podState.timeZoneOffset,
+                tempBasal = tempBasal,
+                basalProgram = basalProgram
+            )?.let { delta -> it + delta }
         } ?: basalDelivered.takeIf { isActivationCompleted }
         
         // Update bolus pulses delivered (exclude basal corrections)
--- a/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
+++ b/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
@@ -244,7 +244,7 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
         get() = podState.bolusPulsesDelivered
 
     private val basalPulsesDelivered: Short?
-        // Compute basal pulses delivered by subtracting bolus pulses from total pulses delivered
+        // Compute basal pulses (including corrections) by subtracting user bolus pulses from total
         get() = pulsesDelivered?.let { total ->
             bolusPulsesDelivered?.let { bolus ->
                 (total - bolus).toShort()
@@ -320,6 +320,63 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
             logger.debug(LTag.PUMP, "  total integrated delivery: ${"%.4f".format(total)}U")
         }
     }
+
+    override fun needsBasalCorrection(): Boolean {
+        val correctionThreshold = -PodConstants.POD_PULSE_BOLUS_UNITS / 2  // -0.025U
+        
+        if (!isActivationCompleted) return false  // Don't correct during activation/priming
+        if (isSuspended || isPodKaput) return false
+
+        // Check cooldown: 2 minutes minimum between corrections (prevents rapid corrections while allowing timely response)
+        podState.lastBasalCorrectionTime?.let {
+            if (System.currentTimeMillis() - it < 2 * 60 * 1000L) return false
+        }
+
+        // Compute drift once for efficiency
+        val drift = basalDrift
+
+        // Reset if drift exceeds boundaries (over-delivery or severe under-delivery)
+        if (drift >= PodConstants.POD_PULSE_BOLUS_UNITS * 2 || drift <= -PodConstants.POD_PULSE_BOLUS_UNITS * 2) {
+            logger.warn(
+                LTag.PUMP,
+                "Resetting basal drift: drift=${"%.3f".format(drift)}U, " +
+                    "expected=${"%.3f".format(podState.basalExpected ?: 0.0)}U -> actual=${"%.3f".format(basalDelivered)}U"
+            )
+            podState.basalExpected = basalDelivered
+            store()
+            return false
+        }
+
+        // No correction if drift too small
+        if (drift > correctionThreshold) {
+            return false
+        }
+
+        // Safety check: don't correct when TBR = 0 (algorithm explicitly requested zero insulin)
+        // Except zero temp due to bolus delivery, allow corrections in that case
+        if (tempBasal?.rate == 0.0) {
+            val timeSinceLastBolus = podState.lastBolus?.startTime?.let { System.currentTimeMillis() - it }
+            if (timeSinceLastBolus == null || timeSinceLastBolus >= 5 * 60 * 1000L) {
+                return false
+            }
+        }
+
+        // Correction triggers when: -0.10U < drift <= -0.025U and no zero-TBR (unless recent bolus)
+        return true
+    }
+
+    override var lastBasalCorrectionTime: Long?
+        get() = podState.lastBasalCorrectionTime
+        set(value) {
+            podState.lastBasalCorrectionTime = value
+            store()
+        }
+
+    override var basalCorrectionInProgress: Boolean
+        get() = podState.basalCorrectionInProgress
+        set(value) {
+            podState.basalCorrectionInProgress = value
+        }
 
     override val lastStatusResponseReceived: Long
         get() = podState.lastStatusResponseReceived
@@ -733,10 +790,10 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
             }
         } ?: basalDelivered.takeIf { isActivationCompleted }
         
-        // Update bolus pulses delivered
+        // Update bolus pulses delivered (exclude basal corrections)
         podState.bolusPulsesDelivered = podState.bolusPulsesDelivered?.let {
             podState.pulsesDelivered
-                ?.takeIf { podState.lastBolus?.deliveryComplete == false }
+                ?.takeIf { podState.lastBolus?.deliveryComplete == false && !podState.basalCorrectionInProgress }
                 ?.let { previousTotalPulses ->
                     (it + calculateBolusPulseIncrease(
                         previousTotalPulses,
@@ -967,6 +1024,9 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
         var lastBolus: OmnipodDashPodStateManager.LastBolus? = null,
 
         var bolusPulsesDelivered: Short? = null,  // Cumulative count of bolus pulses for basal tracking
-        var basalExpected: Double? = null  // Initialized to actual on first drift calculation
+        var basalExpected: Double? = null,  // Initialized to actual on first drift calculation
+        
+        var lastBasalCorrectionTime: Long? = null,  // Timestamp of last basal correction attempt (for cooldown)
+        @Transient var basalCorrectionInProgress: Boolean = false  // Transient flag: true while basal correction is delivering
     ) : Serializable
 }
--- a/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
+++ b/pump/omnipod/common/src/main/kotlin/app/aaps/pump/omnipod/common/bledriver/pod/state/OmnipodDashPodStateManagerImpl.kt
@@ -239,6 +239,88 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
             store()
         }
 
+    private val bolusPulsesDelivered: Short?
+        // Track pulses delivered via bolus for basal delivery detection
+        get() = podState.bolusPulsesDelivered
+
+    private val basalPulsesDelivered: Short?
+        // Compute basal pulses delivered by subtracting bolus pulses from total pulses delivered
+        get() = pulsesDelivered?.let { total ->
+            bolusPulsesDelivered?.let { bolus ->
+                (total - bolus).toShort()
+            }
+        }
+
+    private val basalDelivered: Double
+        // Compute basal delivered in insulin units
+        get() = (basalPulsesDelivered ?: 0) * PodConstants.POD_PULSE_BOLUS_UNITS
+
+    private val basalDrift: Double
+        // Compute drift (actual - expected: positive = over-delivery, negative = under-delivery)
+        get() = basalDelivered - (podState.basalExpected ?: basalDelivered)
+
+    private fun integrateExpectedDelivery(startTime: Long, endTime: Long): Double? {
+        logger.debug(LTag.PUMP, "integrateExpectedDelivery: period ${(endTime - startTime) / 1000.0}s")
+        
+        // Validate time period
+        if (startTime > endTime) {
+            logger.error(LTag.PUMP, "Invalid time period: startTime=$startTime > endTime=$endTime")
+            return null
+        }
+        
+        // Build set of time boundaries where rate changes
+        val boundaries = mutableSetOf(startTime, endTime)
+        
+        // Add temp basal start/end if within period
+        tempBasal?.let { tb ->
+            val tempStart = tb.startTime
+            val tempEnd = tb.startTime + tb.durationInMinutes * 60_000L
+            if (tempStart in startTime until endTime) boundaries.add(tempStart)
+            if (tempEnd in startTime until endTime) boundaries.add(tempEnd)
+        }
+        
+        // Add basal program segment boundaries
+        basalProgram?.segments?.forEach { segment ->
+            // Calculate day boundaries in pod's local timezone, not UTC
+            val timeZoneOffset = podState.timeZoneOffset ?: 0
+            val dayStartLocal = ((startTime + timeZoneOffset) / 86400_000L) * 86400_000L - timeZoneOffset
+            var segmentStart = dayStartLocal + segment.startSlotIndex.toLong() * 30 * 60_000L
+            
+            // If segment already passed today, start checking tomorrow
+            if (segmentStart <= startTime) {
+                segmentStart += 86400_000L
+            }
+            
+            // Add all occurrences of this segment boundary until endTime
+            while (segmentStart < endTime) {
+                boundaries.add(segmentStart)
+                segmentStart += 86400_000L
+            }
+        }
+        
+        // Integrate over each segment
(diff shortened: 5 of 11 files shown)


vanelsberg:
This PR contains unrelated commits from the dev branche, which I think is undesired?
Not an expert on this: maybe consult with @MilosKozak on how to fix.

vanelsberg:
In addition, it is stated: "It should be reviewed on top of PR https://github.com/nightscout/AndroidAPS/pull/4467"
This makes merging this PR with AAPS-CI impossible. Is there a reason this PR's are separate?
Please consider combining PR#4467 and this PR into one?

vanelsberg:
Nice 👍
Will need to review and test implementation in detail and test to see how this PR does.

vanelsberg:
This PR contains lots of file changes and commits.
To enable proper reviewing, please consider squashing all commits into one?

nl-ruud:
@vanelsberg I think I fixed it, there are two commits now in this PR: the first one equals 4467 (quantifying the basal drift) and the second one is resolving the basal drift. does this make sense? I think that also would mean we can close #4467 as the distinction is now just clear in this PR.

vanelsberg:
> I think I fixed it, there are two commits now in this PR....

Yes! This looks a lot better, making it easier to follow implementation. Helps reviewing/testing and - eventually - when ok, acceptance for merging with dev.

(When 4467 is no longer of use, I think best to close it to prevent confusion?)

pump/omnipod/dash/src/main/kotlin/app/aaps/pump/omnipod/dash/driver/pod/state/OmnipodDashPodStateManagerImpl.kt
```diff
@@ -239,6 +239,140 @@ class OmnipodDashPodStateManagerImpl @Inject constructor(
...
+        // Add temp basal start/end if in period
+        tempBasal?.let { tb ->
+            val tempStart = tb.startTime
+            val tempEnd = tb.startTime + tb.durationInMinutes * 60_000L
+            if (tempStart in (startTime + 1) until endTime) boundaries.add(tempStart)
+            if (tempEnd in (startTime + 1) until endTime) boundaries.add(tempEnd)
+        }
+        
+        // Add hour boundaries for basal program transitions
+        var nextHour = (startTime / 3600_000L + 1) * 3600_000L  // Next whole hour
```
sprig3: I do worry a little about hard-coding this to hourly. I guess it accurately reflects the current reality, though! 

Just one of those things that down the line (if half hour basal segments for Omnipod became a thing) would need to be changed, but maybe those doing the change might miss.

vanelsberg:
@nl-ruud **Proposed way to testing this PR:**

To see if AAPS is affected first _test this whiteout the PR_, preferably for 2 or more days. _Then merge this PR and retest_ with current AAPS (this can be current dev or current 3.4 release).

**Test:**
1. Check the pod history on the DASH tab.
Note "Total insulin delivered" at two points in time for a specific day: At 00:00h and 24:00h; the difference is what the Pod reports as insulin delivered for that day period.

2. Check AAPS TDD statistics
For that same day (2), check on what is reported as the sum of insulin (first column at the right of the date). 

_Validation_:
"Total insulin delivered" from (1) should closely match what is reported as TDD. Expect a difference no more than 0.**0**5U

Optional, check the AAPS logfiles, like filtering on the terms below:
_grep -E 'PUMP_BASAL|deliverBasalCorrection|calculateBolusPulseIncrease|integrateExpectedDelivery' *_

nl-ruud:
> Expect a difference no more than 0.5U

Actually no more than 0.05U. If a bigger difference is perceived then I’d love to see the AAPS logs of that 24 hour period.

Also, the [docs](https://androidaps.readthedocs.io/en/latest/CompatiblePumps/OmnipodDASH.html#omnipod-dash-known-aaps-constraints-issues) currently state “When using SMB, limit the interval to 5 minutes minimum to avoid this issue.”. Even though that does not avoid the issue (as stated there), it may reduce the severity. So reset that setting to default (i believe it’s 3 minutes) before you test if you followed that direction from the docs. Especially if you use a sensor like FSL that updates every minute.

vanelsberg:
Have been testing/reviewing this PR for the past 4 days (PR merged with current 3.4 dev):
Code looks good. 

Measuring insulin drift comparing what AAPS reports on statistics/TDD with actual "insulin delivered" as reported on the DASH pod history. Differences measured are in expected range and tolerance of 0.00..0.05U 👍

_Note_:
Did need to do some approximations on actual insulin delivered as reported by Pod history for running TBR's at 00:00. For example: TBR running from 23:55h to 00:05 of 0.5U/h, adjust insulin delivered at 23:55 with 0.25 units

results

Now testing for 4 days (similar circumstances, identical Profile) for AAPS without PR4499.
Will share final test and code review results.

xitation:
I've cherry picked the commits into Master (3.4.0.0) and installed updated apk with fix applied tonight. 

I saw a drift of about 2U of insulin before patch. Will post back findings tomorrow with patch installed. 

image

xitation:
So far seems to be ok, however it's been beeping when making corrections. I have the Bolus alert / beep enabled only. Annoyed my wife overnight :)

Perhaps could look at using SMB/TBR or one of the other alert beeps for this instead of the Bolus one, I suspect a number of people would leave the Bolus beep enabled. 

image

olorinmaia:
Thanks for looking into this issue and providing a promising fix.

I will also assist with testing. Will log TDD for Dash and AAPS calculated like this:
image

I will first run without this PR to see for myself the difference between Dash TDD and AAPS TDD, I haven't had time to check before now.

I will after a day or two apply PR by cherry-picking commits onto my master build and report back with result. My son only got around 15-20TDD, depending on activity, should I expect to see a big total diff without this PR?

rickvh3:
With the fix, I notice that the loop has become more consistent and shows less unpredictable behavior. I’ve been using this since January 23 (14 days now), and I’m satisfied with the result.

nl-ruud:
> So far seems to be ok, however it's been beeping when making corrections. I have the Bolus alert / beep enabled only. Annoyed my wife overnight :)

If you cherry pick 254dfe3e610cb4a2ffc541db1cf1206285c31939 then the beeps should be gone. Will wait for the others to complete testing and then add a final commit for the PR.

vanelsberg:
Confirming beeps for correction bolus. Noticed this only twice during testing: 1 _unexpected beep_ (ignored it) and _unexpected bolus notification for 0.05U on Watch_. The last one triggered some questions (forgot all about this when posting test results on TDD deviations).

Currently starting to test DASH without PR4499 under similar circumstances. After 4 days testing I'l post results here.
 
> .... If you cherry pick [254dfe3] then the beeps should be gone.

@nl-ruud You may want to check if your code change not only skips the **beep** but also the **notification**?
Once you added this to the PR I'll retest for "no beeps" and no "notification" and post final test results.

olorinmaia:
Posting results from logging total insulin delivery for two days, one without PR and one with PR. No issues so far. 

We got all beeps disabled. No issues when they are disabled either.

image

xitation:
> > So far seems to be ok, however it's been beeping when making corrections. I have the Bolus alert / beep enabled only. Annoyed my wife overnight :)
> 
> If you cherry pick [254dfe3](https://github.com/nightscout/AndroidAPS/commit/254dfe3e610cb4a2ffc541db1cf1206285c31939) then the beeps should be gone. Will wait for the others to complete testing and then add a final commit for the PR.

Seems to have resolved beeping, thanks!

In terms of test data here are my results so far:

image

belevine:
I have tested this PR for several days now on v3.4.  Before building with the PR, I observed a drift of 1.5-2.5 units per day.  After the PR, I'm getting no drift or 0.05 units (which I suspect may be more due to a challenge in determining the exact Dash delivered amounts at midnight if the only timestamps are 5-10 mins before or after).

I also have manually bolused while testing the PR and found this also was reported correctly.  The AAPS statistics included the manual bolus while the Dash history did not.

nl-ruud:
I just submitted a (final?) update to the PR that addresses two issues that were reported by @xitation, which are:

- no more beeps
- reduced the cooldown period from 5 minutes to 2 minutes to avoid drift reset

vanelsberg:
**Test without this PR:**
image
⚠️Did not trust the **high drift values calculated for 08/09 and 09/02**: doublecheck my math, but they are correct. Nothing unusual to see in basal rates and.... then I realized I did manual bolus at the time by PEN ("only registering"). To my surprise in AAPS treatments a "register only" bolus is indistinguishable from a normal "meal" type, So I cannot check statistics. No logfiles either as they were already deleted.

As far sa I can remember these manual boluses very likely did amount to 4U. Actual drift values would match expected values?

**Test with PR4490:**
image
Except for the bolus beeps (which is now fixed) I did not encounter any problems with PR4499. My tests as reported above confirm reported insulin drift for current DASH driver which is effectively fixed with this PR.

Reviewing code for PR4499 with fix I did not see any problems.

vanelsberg:
> I just submitted a (final?) update to the PR

Will retest this for a few days.

olorinmaia:
Applied latest commit but so far only 0,05+- difference between pump and AAPS TDD. I think this PR is ready for merge, but will keep monitoring / logging. CC: @MilosKozak 

image

vanelsberg:
> I think this PR is ready for merge...

100% Agree.

@MilosKozak Ik think this is an important bugfix, affects users especially when on lower TDD (like for kids on DASH)
After merge and testing on dev, we should consider making this a bugfix on 3.4 release?

vanelsberg:
My final review / food for thought: (feel free to correct me on this!):

(Edit) Post was not relevant for this PR. Moved to separate issue.

crimsonkage:
Been testing this for a couple days, it works great.
Some background is that I'm a 1 minute looper using Auto ISF. 

Because of the frequency of the loops I was really susceptible to this. 
My basal is 24 units a day, but I was missing at least 6 of that. (25% off)
It appears that my profile was boosted by 20% to compensate for the missing basal and it causes some instability in my profile. 

I'm still dialing it back down and it appears that my profile is matching what my profile was before starting on AAPS. 

A++++ needed fix for 1 minute loopers.

No instability or crashes or pump errors introduced that I can see.

crimsonkage:
One other comment on this. 

Everyone using a dash will need to be careful and update their profile when installing an AAPS version with this fix included. 

My profile basal has dropped over 20% at this point. 
(1 minute looper so caused a bigger drift)

I think this will need to be mentioned pretty high up the patch notes / release announcement to get attention.

nl-ruud:
> Everyone using a dash will need to be careful and update their profile when installing an AAPS version with this fix included.

I think this statement is somewhat too strong. I can imagine the basal drift kind of slipped into users profiles over time, especially when using tools like autotune. However, the underlying issue would still be that the profile itself was inaccurate. If the profile had originally been established using proper basal and/or ISF testing with the loop disabled, it should remain valid.

> I think this will need to be mentioned pretty high up the patch notes / release announcement to get attention.

That said, I do agree that the release notes should explicitly mention that basal drift may have (un)intentionally propagated into user profiles. In such cases, users may need to reassess and correct their profile after updating.

sprig3:
> I think this statement is somewhat too strong. I can imagine the basal drift kind of slipped into users profiles over time, especially when using tools like autotune. However, the underlying issue would still be that the profile itself was inaccurate. If the profile had originally been established using proper basal and/or ISF testing with the loop disabled, it should remain valid.
> 


I don't know about others, but I raise my profile by 20% when I loop just to counteract this issue. So, knowing I no longer have to do that is important. (But I bet folks will adjust back quickly.)

nl-ruud:
@sprig3 you did read the second part of my comment as well, did you?

xitation:
> > Everyone using a dash will need to be careful and update their profile when installing an AAPS version with this fix included.
> 
> I think this statement is somewhat too strong. I can imagine the basal drift kind of slipped into users profiles over time, especially when using tools like autotune. However, the underlying issue would still be that the profile itself was inaccurate. If the profile had originally been established using proper basal and/or ISF testing with the loop disabled, it should remain valid.
> 
> > I think this will need to be mentioned pretty high up the patch notes / release announcement to get attention.
> 
> That said, I do agree that the release notes should explicitly mention that basal drift may have (un)intentionally propagated into user profiles. In such cases, users may need to reassess and correct their profile after updating.

Yeh I agree there is a chance for people to not understand the ramifications of this patch without sufficient warning of that this may have had them do to profiles etc. 

I think it's just a good idea to have sufficient warnings around Basil profile maybe being incorrect etc.

nl-ruud:
After running this fix for several weeks, I realized that basal drift actually originates when the active temp basal is cancelled. Previously, I was checking for a drift compensation need on every status update. This could introduce an edge case where a status response arrives just before the basal is delivered, potentially causing an early compensation of 0.05U. To address this, I’ve moved the compensation evaluation to immediately follow the (driver internal) temp basal cancellation. This aligns the check with the source of the drift, simplifies logic and eliminates potential timing issues, since at that point we know the temp basal has already been cancelled.

vanelsberg:
> After running this fix for several weeks, I realized that basal drift actually originates ....

Will review test this, but no earlier then at the end of next week. Currently on Medtrum...

2223333333:
> I realized that basal drift actually originates when the active temp basal is cancelled.

Perhaps even when temp basal 0 is enabled and standard basal is disabled, drift may occur.

nl-ruud:
> Perhaps even when temp basal 0 is enabled and standard basal is disabled, drift may occur.

When temp basal is 0, no drift can occur (actually existing drift can - and will - decrease if this is the case for hours due to how the Omnipod works). When there is no active (basal) profile, you cannot activate the pump: it requires a basal profile.

But even if this would all not be the case: the bolus pulses, basal pulses and drift are updated every status response: no change there. Just the decision to _deliver_ a compensation is only made when we can know for sure a compensation is needed; the latest commit just eliminates a potential race condition.

xitation:
I've merged latest commits, built apk and upgraded. Hold tight while I collect data for next few days.

nl-ruud:
I created a (disposable..) [Python script](https://gist.github.com/nl-ruud/c4ce8034bfe43bb838fd78539339aff2) that parses the AAPS logs and generates a graph to illustrate how this PR handled the basal drift. Reporting for a week:
- A negative accumulated error indicates that the pod delivered less insulin than expected.
- The green bars represent the basal compensation boluses delivered by the PR.
- When AAPS is zero-temping (dark red line at the bottom of the graph), no compensation is delivered if there is a deficit (e.g., ~10:00 to ~15:00 in the first graph), for safety reasons.
- On a pod change (diamond in the graph), the accumulated error resets.








The reconciliation for these dates:

| Date       | Delivered (Pump) | TDD (AAPS) | Deviation | Corrections |
|:-----------|-----------------:|-----------:|----------:|------------:|
| 2026-02-26 |            15.20 |       15.2 |      0.00 |          23 |
| 2026-02-27 |            15.30 |       15.3 |      0.00 |          15 |
| 2026-02-28 |            15.05 |       15.0 |      0.05 |          11 |
| 2026-03-01 |            17.75 |       17.7 |      0.05 |          16 |
| 2026-03-02 |            14.60 |       14.7 |     -0.10 |          15 |
| 2026-03-03 |            16.70 |       16.7 |      0.00 |          22 |
| 2026-03-04 |            14.95 |       14.9 |      0.05 |          33|

Note: if you want to try this script yourself, make sure all pump related logging is enabled in the log settings!

nl-ruud:
Rebased onto dev. No conflicts; updated file paths to match upstream moves.

xitation:
Ran your Python Script produced the following:

[AndroidAPS_LOG_1772412285131_err_2026-03-02.html](https://github.com/user-attachments/files/25670561/AndroidAPS_LOG_1772412285131_err_2026-03-02.html)

xitation:
Latest data extracts:

image

xitation:
Hey all, 

@nl-ruud rightly pointed out I had forgotten to minus the priming a new pod amount from my values in the previous data posted.

I've fixed this for the data I still have and below is the correct view of my data since cherry picking commit ece3ecc on the 2026/02/27 - 10:30AM.

image

RSC-RJCgit:
My percent difference ob about 24 units was more like 12% or more, on 1 minute aISF

1st 2 days modified more like 3.5% and 5.5% of 20u tdd
( next 2 days  closer to 2.2 %of  19.1 tdd then 15.75 tdd ),much better for 1 minute looping ( really 1 minute bgl, about 2 minute looping set on 1 minute ) than about 15% of 21u pre- fix.

nl-ruud:
To merge this PR into dev, and after discussing with @vanelsberg, I added a semaphore file to enable basal drift compensation. You need an empty file called `omnipod_drift_compensation` in the `extra` subfolder of your phone’s AAPS directory for the compensation to be active. Commits are squashed into three parts:
- Detect basal drift.
- Automatically compensate basal drift. Logic extended: drift can occur on temp basal cancel, cancel + set, or set without cancel. The last case was previously not considered, which could delay compensation.
- Make automatic compensation conditional on the semaphore file. This allows merging while enabling broader testing; the guarding can be removed later.

I believe this PR is ready to merge. Please react with 👍 if you agree or 👎 if you have concerns, and share reasoning if you disagree.

Release notes:
- Basal drift may have (un)intentionally propagated into user profiles; users may need to reassess their profile after updating.
- Basal deficit compensation is delivered as a 0.05U bolus and will appear as such in the GUI and Watch app.

xitation:
I know that this PR will help a bunch of people I've had conversations with. Especially in the AutoISF fork/branch community where minor things like this can have a pretty big impact on a system with a lot of power to adjust doses. Also for anyone with little people on low doses this could be a real big issue for them too.

Appreciate you taking the time to put this together @nl-ruud!

From my testing it's been working well for me and I vote it's ready for merge.

nl-ruud:
Rebased onto dev; no conflicts or changes.

xitation:
Cherry picked the 3 commits into the latest 3.4.1.0 Master and it builds ok.

1st commit name - Quantify the basil drift
git cherry-pick 35189a90b087fa976e6b4f9951e60480bf423f75
2nd commit name - Resolve the basil drift
git cherry-pick 3a19017f8f2d65d27b4efe25e60b91f625ae6ba5
3nd commit name - Added semaphore file for enabling compensation
git cherry-pick 2574f3564ffe549b48c735945424abb97233465 

Merging the issue-4158-dev-fix branch into a non Dev Master ended up building a Dev build.

vanelsberg:
For testers:

**Merging a PR from Android Studio:**
Make sure you are on the latest AAPS dev or release commits. Then from the AS command line run:

      git fetch origin pull/4499/head:dev-PR4499
      git merge dev-PR4499 -m "Merging PR4490"
      
**Merging using AAPS-CI:** (preferred)

aaps-ci4499

vanelsberg:
Took some time before I could test (was running Medtrum), my test results on deviation for the past 3 days.
Note that for 10-03-2026 deviation was at 0,10 but was compensated to 0,05 18 mins later (because of a long running 0-TBR. So based on these results I'd say: looks good!

Code review: afaikt no issues
Drift compensation fix: as expected, no issues

Personal evaluation:
This PR fixes an important issue with the DASH driver that has positive impact on users BG regulation while on TBR's
@MilosKozak Recommend merging with dev for broader test coverage.

image

nl-ruud:
@MilosKozak This PR has been tested over the past couple of months with very positive feedback, everyone [agrees](https://github.com/nightscout/AndroidAPS/pull/4499#issuecomment-4013899964) it’s ready to merge. The issue it addresses might seem minor at first glance, but the changes make a real difference in practice. It’s now guarded by a semaphore file, allowing broader testing before it becomes the default.

crimsonkage:
Got another data point that this PR helps. 
I helped someone with brittle and their basal is under 10 a day. 
Their profile was good, but loop was unstable. 
Installed this PR and it Immediately stabilized their loop. 
(Except they had their basal too high to compensate for missing basal from bug 10%)

No bugs or issues present.

BrowserBuildCG:
Working nicely.

BrowserBuildCG:
> ## Omnipod Dash basal drift resolution
> 
> This pull request introduces a new "basal correction" feature for the Omnipod Dash pump integration. The main goal is to detect under-delivery of basal insulin and automatically trigger a small correction bolus when needed. The implementation tracks delivered basal and bolus pulses, calculates expected vs. actual insulin delivery, and manages correction logic with safety checks and cooldowns. 
> 
> ## Recap of the issue
> 
> The Omnipod Dash pump exhibits behavior that causes it to deliver less basal insulin than AAPS expects (issue #4158). This is effectively a hardware limitation of the pump.
> 
> The Dash uses an internal timer to determine when a basal pulse of 0.05 U is delivered. Once the timer interval elapses, the pulse is delivered. However, this timer is restarted whenever a basal rate change occurs (and possibly also when a bolus or SMB is delivered).
> 
> When used in combination with looping, this leads to structural under-delivery of basal insulin, as the algorithm adjusts the basal rate frequently. While the Medtrum pump compensates for this behavior internally, the Dash does not, and neither does the current Dash driver implementation in AAPS.
> 
> The issue is most apparent during the night. During daytime operation, SMBs often result in a basal rate of 0, which masks the effect. In observed usage, this results in approximately 10% of the expected Total Daily Dose (TDD) not being delivered over a 24-hour period. Additionally, glucose targets are often not reached overnight, particularly after meals with prolonged glucose impact (e.g. pasta).
> 
> ## Approach
> 
> This pull request introduces automatic basal drift compensation by correcting the pump lag with small correction boluses. A correction bolus of 0.05 U is delivered as soon as the detected basal deficit reaches 0.025 U. Several safety constraints are applied. For example, corrections are not delivered during a temporary basal of 0 unless that temporary basal is the direct result of a recently delivered bolus.
> 
> With this approach, the deviation between expected and delivered basal insulin is kept within a bounded range of approximately −0.050 U to +0.025 U.
> 
> With this fix applied:
> - The TDD reported by AAPS (under *Statistics*) matches the total amount actually delivered by the pump (Dash menu → *Pod management* → *History*).
> - Overnight glucose targets are consistently reached.
> 
> Without this fix, a clear discrepancy between reported and delivered TDD can be observed.
> 
> ## Main Changes
> 
> ### Basal Correction Feature
> 
> - Added a new custom command, `CommandDeliverBasalCorrection`, and integrated it into the command queue and handling logic in `OmnipodDashPumpPlugin` to trigger a basal correction bolus when under-delivery is detected.
> - Implemented the `deliverBasalCorrection` method to handle the actual delivery of the correction bolus, including safety checks (reservoir level, bolus in progress, cooldowns, etc.).
> 
> ### Basal Delivery Tracking and Drift Calculation
> 
> - Added logic in `OmnipodDashPodStateManagerImpl` to track basal and bolus pulses separately, compute actual delivered basal, and calculate "drift" (difference between expected and actual basal delivery). This includes new fields in the pod state and methods for integrating expected delivery over time. 
> - The `needsBasalCorrection` method determines when a correction is needed based on drift, cooldowns, and other safety criteria.
> 
> ### Pod State Update and Logging Refactor
> 
> - Refactored pod state update logic to centralize updates in a new `updatePodState` method, which also updates basal tracking fields and logs basal delivery/drift for debugging. This method is now used in both default and alarm status response handlers and removes code duplication from these methods. 
> 
> Logging example:
> 
> ```text
> PUMP_BASAL act=8,10U (tot=33,25U bol=25,15U) exp=9,42U err=-1,32U dErr=-0,02U
> ```
> 
> Legend for the log fields:
> 
> | Field | Description |
> |------|------------|
> | `act` | Actual basal units delivered (`tot` − `bol`) |
> | `tot` | Total units delivered |
> | `bol` | Bolus units delivered |
> | `exp` | Expected basal units delivered |
> | `err` | Error / cumulative basal drift (`act` - `exp`) |
> | `dErr` | Delta of `err` on this pump status response |
> 
> ### Interface and State Enhancements
> 
> - Updated the `OmnipodDashPodStateManager` interface and its implementation to support new fields and methods for basal correction tracking (`lastBasalCorrectionTime`, `basalCorrectionInProgress`, `needsBasalCorrection`). 
> 
> These changes together enable the system to automatically detect and correct small basal under-deliveries, improving insulin delivery reliability and safety.
> 
> ## Tests Performed
> 
> To verify the fix, we compared the total insulin delivered at 00:00 at the start of a day with the total at 00:00 at the end of the same day. 
> 
> Before this fix, the difference consistently resulted in a lower number than the TDD reported by AAPS, confirming under-delivery. After applying the fix, this discrepancy no longer occurs: AAPS TDD matches the pump’s delivered total.
> 

This fix for Dash Basal Correction is a big step closer to a perfecting an accurate delivery of insulin by the Omnipod Dash pump. The AAPS calculations to keep BG in tighter range is now closer than before this fix. RUDD had made the AAPS system better. Nice work. This should be merged into the AAPS Master.

nl-ruud:
@BrowserBuildCG Thanks for the feedback — and great to hear you're enthusiastic about this! 😀

Small nit: would you mind removing the quote of the full PR description from your comment? It’ll make the thread a bit easier to scan.

RSC-RJCgit:
Using the newest " basal drift" has smoothed out control heaps.. and using this PR we just got a 100% TIR week for the first time i can remember; (as well .. other factors have helped)

MilosKozak:
## PR Review: Omnipod Dash Basal Drift Compensation

Well-motivated PR addressing a real and impactful hardware limitation (~10% TDD under-delivery). The safety guards are reasonable and the semaphore-file gating makes it low-risk for non-opt-in users. Here are my findings:

### Strengths

- **Conservative safety design**: cooldown timer, drift bounds reset at ±0.10U, skip when TBR=0 (unless recent bolus), skip during activation/suspend/kaput
- **Good refactoring**: extracted `updatePodState()` eliminates duplication between `updateFromDefaultStatusResponse` and `updateFromAlarmStatusResponse`
- **Feature-gated** behind a semaphore file — off by default, consistent with existing patterns (`enable_autotune`, etc.)
- **Minimal correction size** (0.05U = 1 pulse) — low risk per correction

### Concerns

#### 1. `integrateExpectedDelivery` timezone handling is fragile

The basal program segment boundary calculation manually offsets by `podState.timeZoneOffset`:
```kotlin
val timeZoneOffset = podState.timeZoneOffset ?: 0
val dayStartLocal = ((startTime + timeZoneOffset) / 86400_000L) * 86400_000L - timeZoneOffset
```
Meanwhile, `BasalProgram.rateAt()` uses `Calendar.getInstance()` (device local time). If the pod timezone and device timezone diverge (which `sameTimeZone` already tracks as a known issue), the segment boundaries in `integrateExpectedDelivery` won't match the rates returned by `rateAt()`. The PR comments show this was already a problem once (commit reverted and re-fixed) — this area warrants extra scrutiny and test coverage.

#### 2. Short overflow risk in pulse arithmetic

`basalPulsesDelivered` is computed as `(total - bolus).toShort()`. Kotlin `Short` wraps at 32767. While typical 3-day pod usage stays well within bounds, `bolusPulsesDelivered` is incremented cumulatively via `calculateBolusPulseIncrease` and cast via `.toShort()` — any intermediate `Int` arithmetic silently truncated to `Short` could wrap over many status updates.

#### 3. Correction bolus recorded as `BS.Type.NORMAL` / `BolusType.DEFAULT`

The correction bolus is logged as a normal bolus in history. The loop algorithm will correctly account for it in IOB, but there's no way to distinguish basal corrections from user boluses in history/reports. A dedicated bolus type or annotation would improve traceability and debugging.

#### 4. Race condition with `bolusDeliveryInProgress`

In `deliverBasalCorrection()`, the shared `bolusDeliveryInProgress` flag is set, but if a user bolus is requested concurrently via the queue, the interleaving depends on command queue serialization. The `deliveryStatus?.bolusDeliveringActive()` check helps, but there's a TOCTOU window between the check and the actual bolus command being sent.

#### 5. Drift reset threshold may mask real under-delivery

```kotlin
if (drift >= POD_PULSE_BOLUS_UNITS * 2 || drift <= -POD_PULSE_BOLUS_UNITS * 2)
```
This resets at ±0.10U, but correction triggers at -0.025U, so the effective correction window is -0.025U to -0.10U. If a single status update shows a jump past -0.10U (e.g., delayed status response), the drift is reset rather than corrected — potentially masking real under-delivery.

#### 6. No unit tests

There are no tests for `needsBasalCorrection()`, `integrateExpectedDelivery()`, `calculateBolusPulseIncrease()`, or the drift calculation logic. These are complex numerical algorithms with edge cases (timezone transitions, pod restarts, concurrent boluses). For code that autonomously delivers insulin, thorough test coverage seems essential.

#### 7. Interface placed in `common` but only Dash uses it

`needsBasalCorrection()`, `lastBasalCorrectionTime`, and `basalCorrectionInProgress` are added to the `OmnipodDashPodStateManager` interface in the `common` module. If Eros isn't planned for this feature, it forces Eros implementations to stub these methods unnecessarily.

### Minor nits

- `statusDescription = "BASAL COMPENSATION BOLUS"` is all-caps, unlike other command descriptions
- Inconsistent log precision (`"%.3f"` vs `"%.4f"`) scattered across drift logging
- `mapIndexed` in `integrateExpectedDelivery` captures `index` only for logging — could use `forEachIndexed` + accumulator to avoid intermediate list allocations

### Summary

This is a well-designed approach to a real problem. Before merging, I'd recommend:
1. **Add unit tests** for the core drift calculation and correction logic
2. **Verify timezone handling** with explicit test cases for timezone changes mid-pod
3. **Consider a dedicated bolus type** for correction boluses for traceability
4. **Document the ±0.10U reset threshold** rationale and whether it should allow larger corrections

nl-ruud:
@MilosKozak Thanks for your detailed feedback — much appreciated. I’ve reviewed all points carefully. I’ll add testcases (which address points 1 and 6). For the other items, I’ve considered the feedback and decided that the current approach is intentional. I’d appreciate your thoughts on whether this seems reasonable, or if I might be overlooking something.

> #### 2. Short overflow risk in pulse arithmetic

I used Short here to stay consistent with totalPulsesDelivered. basalPulsesDelivered and bolusPulsesDelivered are aligned with that type. Even at the maximum value (32767 pulses × 0.05U = 1638.35U), we’re well within pod limits (200U), and both values reset on pod change.

I understand the concern about silent truncation from Int → Short, but since totalPulsesDelivered is already a Short, any overflow would originate there first. From that perspective, keeping these fields as Short is consistent.

Fully eliminating this risk would require changing the underlying type of totalPulsesDelivered, rather than addressing it only here.

> #### 3. Correction bolus recorded as `BS.Type.NORMAL` / `BolusType.DEFAULT`

I agree with this point. Properly addressing it would require a broader change to the driver internals, so the current implementation is intentional. At the moment, lastBolus carries a BS.Type, but conceptually it should be a BolusType, only converted to BS.Type when communicating back to AAPS.

If you agree with that direction, I’m happy to take it on — but I’d propose handling it in a separate PR to keep this one focused. It would for example also mean a change to the underlying database structure and probably need a data conversion upon upgrade. I will need some help there...

> #### 4. Race condition with `bolusDeliveryInProgress`

I agree this isn’t ideal. I currently see bolusDeliveryInProgress as a workaround. I expect this concern would be resolved as part of addressing point 3, once basal corrections are given their own dedicated BolusType.

> #### 5. Drift reset threshold may mask real under-delivery

The primary goal here is to prevent overdelivery at all costs. In uncertain situations, I prefer resetting rather than compensating to avoid potential over-correction. For example, when enabling drift compensation after it has been off for a while, the drift can be large — resetting first prevents overdelivery.

Even if this occasionally means missing a small compensation, the practical impact is negligible.

It’s also worth noting that I haven’t observed this scenario in any real-world test logs. I’m aware the thresholds are set quite tightly, but this was intentional. They can be increased if needed; so far, however, all tests — even with 1-minute loops using AutoISF, where I would have expected to see issues due to high basals and the 2-minute cooldown — have not triggered this scenario.

> #### 7. Interface placed in common but only Dash uses it

This file was recently moved to common in preparation for starting work on Omnipod 5 support (#4574). It's even called `OmnipodDashPodStateManager.kt` - so not that common after all 😀.

> ### Minor nits

My observations:
- statusDescription follows the ALL-CAPS convention used by other CustomCommands in this codebase (e.g. "SUSPEND DELIVERY" but also “ACCEPT TEMP BASAL" and "BOLUS #.## E").
- The differing log precision (”%.2f” vs “%.4f”) is intentional — the extra precision helped during debugging to spot small drift effects. I’d be happy to align it for consistency if you feel strongly about it, but my preference would be to keep the extra precision.
- I used `mapIndexed` as it felt like the most idiomatic option here. `windowed(2)` already allocates an intermediate list, and since boundaries is very small, the overhead should be negligible. Alternatives like `forEachIndexed` + accumulator would add verbosity without much gain — but open to suggestions if you see a cleaner approach.

> ### Summary

Thanks again — this is very helpful feedback. I’ll:
	1.	Add unit tests for drift calculation and correction logic
	2.	Update and verify timezone handling with explicit test cases
	3.	Leave the bolus-type refactor for a follow-up PR
	4.	Add inline documentation for the ±0.10U reset threshold and rationale

nl-ruud:
@MilosKozak I added unit tests for the drift calculation and correction logic, updated the timezone handling in `integrateExpectedDelivery` (with accompanying test cases), and documented the 0.10U threshold rationale.

Note: #4574 moved the Dash driver to common. The existing test cases are still under dash, but since common is now the location, I placed the new tests there: testing the new internal methods would otherwise add unnecessary complexity.

nl-ruud:
For the record: I may have tested the Atlas pods before without realizing it, but I have now explicitly verified that the basal drift is not specific to the SAW pods—it is also present in the Atlas pods.

nl-ruud:
# DST Verification

As there was a DST transition today, I verified the behavior of `integrateExpectedDelivery` across the change.

## Case 1: Phone time (DST jump)

```text
AndroidAPS._2026-03-29_00-00-00_.0:01:52:28.911 [RxCachedThreadScheduler-367] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():276]: integrateExpectedDelivery: period 5.549s
AndroidAPS._2026-03-29_00-00-00_.0:01:52:28.912 [RxCachedThreadScheduler-367] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():336]:   segment 1/1: 5.5489999999999995s @ 0.25U/h = 0,0004U
AndroidAPS._2026-03-29_00-00-00_.0:01:52:28.913 [RxCachedThreadScheduler-367] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():342]:   total integrated delivery: 0,0004U
AndroidAPS._2026-03-29_00-00-00_.0:03:02:23.214 [RxCachedThreadScheduler-367] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():276]: integrateExpectedDelivery: period 594.303s
AndroidAPS._2026-03-29_00-00-00_.0:03:02:23.216 [RxCachedThreadScheduler-367] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():336]:   segment 1/2: 451.08900000000006s @ 0.25U/h = 0,0313U
AndroidAPS._2026-03-29_00-00-00_.0:03:02:23.216 [RxCachedThreadScheduler-367] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():336]:   segment 2/2: 143.214s @ 0.2U/h = 0,0080U
AndroidAPS._2026-03-29_00-00-00_.0:03:02:23.217 [RxCachedThreadScheduler-367] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():342]:   total integrated delivery: 0,0393U
```

### Observation

`integrateExpectedDelivery` correctly accounts for the DST jump on the phone:
- Interval 01:52:28.9 → 03:02:23.2 = 594.3s
- (Expected; incorrect would be ~4194.3s)

It also correctly splits the interval at the basal change at 02:00, applying:
- 0.25 U/h before
- 0.2 U/h after

## Case 2: Pump time (post-DST sync)

```text
AndroidAPS._2026-03-29_00-00-00_.1:05:24:30.736 [RxCachedThreadScheduler-385] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():276]: integrateExpectedDelivery: period 1198.068s
AndroidAPS._2026-03-29_00-00-00_.1:05:24:30.737 [RxCachedThreadScheduler-385] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():336]:   segment 1/1: 1198.068s @ 0.2U/h = 0,0666U
AndroidAPS._2026-03-29_00-00-00_.1:05:24:30.738 [RxCachedThreadScheduler-385] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():342]:   total integrated delivery: 0,0666U
AndroidAPS._2026-03-29_00-00-00_.1:05:41:13.249 [main] D/PUMPQUEUE: [CommandQueueImplementation.add():209]: Adding: CommandCustomCommand - HANDLE TIME CHANGE
AndroidAPS._2026-03-29_00-00-00_.1:05:41:14.655 [DefaultDispatcher-worker-6] D/PUMPQUEUE: [QueueWorker.doWorkAndLog():140]: performing HANDLE TIME CHANGE
AndroidAPS._2026-03-29_00-00-00_.1:05:41:15.109 [RxCachedThreadScheduler-398] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():276]: integrateExpectedDelivery: period 1004.373s
AndroidAPS._2026-03-29_00-00-00_.1:05:41:15.110 [RxCachedThreadScheduler-398] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():336]:   segment 1/1: 1004.3729999999999s @ 0.2U/h = 0,0558U
AndroidAPS._2026-03-29_00-00-00_.1:05:41:15.111 [RxCachedThreadScheduler-398] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():342]:   total integrated delivery: 0,0558U
```

### Observation

The interval **05:24:30.7 → 05:41:15.1 = 1004.4s** is handled correctly, matching real elapsed time.

## Case 3: Profile transition after DST

```text
AndroidAPS._2026-03-29_00-00-00_.1:06:07:26.716 [RxCachedThreadScheduler-396] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():276]: integrateExpectedDelivery: period 475.138s
AndroidAPS._2026-03-29_00-00-00_.1:06:07:26.717 [RxCachedThreadScheduler-396] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():336]:   segment 1/2: 28.423s @ 0.0U/h = 0,0000U
AndroidAPS._2026-03-29_00-00-00_.1:06:07:26.717 [RxCachedThreadScheduler-396] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():336]:   segment 2/2: 446.71500000000003s @ 0.0U/h = 0,0000U
AndroidAPS._2026-03-29_00-00-00_.1:06:07:26.718 [RxCachedThreadScheduler-396] D/PUMP: [OmnipodDashPodStateManagerImpl.integrateExpectedDelivery$common_fullRelease():342]:   total integrated delivery: 0,0000U
```

### Observation

The basal profile change occurs at **06:00**, which is correct after the DST adjustment  (an incorrect implementation would have shifted this to 07:00.)

## Conclusion

-   DST handling is correct for both:
    -   Phone time
    -   Pump time
-   Basal transitions remain aligned with expectations after the DST change

---

## Limitation

There was no basal/profile transition **before** the pump timezone update in this dataset, so that specific edge case is not covered here.

PR #4509, opened 2026-01-24
Adjust the date display in Statistics to use locale-based formatting …

Angus-repo:
@MilosKozak 
On the statistics page, the date format is currently determined by the phone’s time setting: using dd/MM for 24-hour format and MM/dd for 12-hour format. However, this does not always align with the actual date conventions used in different regions.

I have changed it so that MM/dd or dd/MM is determined based on the locale settings in the language (I10n) files.
Please review and let me know if this approach is acceptable.

### DIFF
--- a/core/interfaces/src/main/res/values/strings.xml
+++ b/core/interfaces/src/main/res/values/strings.xml
@@ -48,6 +48,7 @@
     <string name="tomorrow">Tomorrow</string>
     <string name="today">Today</string>
     <string name="yesterday">Yesterday</string>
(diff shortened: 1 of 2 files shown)


MilosKozak:
I'm not sure about this PR. My concerns:
1) correct beavior will depend on correct translations
2) PreferenceManager will be replaced by DataStore soon
3) try to utilize LocaleDependentSetting interface instead

Angus-repo:
Please close the PR for now, and we can reopen it once there is a suitable approach for point 1.

PR #4510, opened 2026-01-26
feat: Expand insulin pump compatibility by serial number prefix

hhfcvmars:
I have confirmed with their developers that this configuration supports all pump models, and the serial numbers comply with their naming conventions. 

Add serial number-based support check for insulin pumps. Only pumps with serial numbers starting with '0', '1', '3', 'A', or 'D' require firmware version validation. All other pump models are now supported by default.

I submitted this once before, but found that the latest dev branch doesn't include this content.

### DIFF
--- a/pump/equil/src/main/kotlin/app/aaps/pump/equil/manager/command/CmdDevicesOldGet.kt
+++ b/pump/equil/src/main/kotlin/app/aaps/pump/equil/manager/command/CmdDevicesOldGet.kt
@@ -174,7 +174,14 @@ class CmdDevicesOldGet(
         }
     }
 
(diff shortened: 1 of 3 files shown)


hhfcvmars:
Additionally, their patch-style waterproof pump is about to receive CE certification, and we hope the community can support and merge this soon.

Philoul:
Target branch for this PR is not the right one (never target on `l10n-dev` only used to synchronize translations with crowding).
Target branch for PRs should always be `dev` branch.

Philoul:
Target branch for this PR is not the right one (never target on `l10n-dev` only used to synchronize translations with crowding).
Target branch for PRs should always be `dev` branch.

hhfcvmars:
ok ,thanks

PR #4541, opened 2026-02-11
Fix KSP build

Philoul:
Each time I modify code during development, I have to build twice:
- first time I get a wrong KSP error (KSP failed with exit code: PROCESSING_ERROR)
- second time it works without any code modification

with this modification, no need to build twice, it works directly the first time...


olorinmaia:
Could you add this to dev also? I think it happens there as well :)

Philoul:
I still have to measure the overall build duration to compare with and without this parameter 😉

PR #4544, opened 2026-02-12
Fix site rotation manage pump setting not being used

samfundev:
Looks like the code accidentally used the CGM setting instead of the pump setting.


Philoul:
Good catch 👍

PR #4582, opened 2026-03-07
Medtrum: Improve BLE connection recovery using handler thread

HuismanErik:
This PR resolves "Pump unreachable" issues where the Medtrum driver stays stuck in the `CONNECTING` state, often following a GATT status 133 error. While other drivers (like Dash) recover correctly on the same hardware, the Medtrum implementation can fail to trigger `mCallback?.onBLEDisconnected()`. This causes the `QueueWorker` to stall and skip new connection attempts.

The root cause is performing a `close()` and state updates directly on the Bluetooth callback (Binder) thread, which can lead to deadlocks or silent connection stalls. Wrapping the GATT cleanup and state updates in a Handler.post() ensures these operations are executed asynchronously. This allows the onConnectionStateChange thread to finalize its lifecycle without being blocked.

In my testing, this change eliminated the "Pump unreachable" connection issues on my Xiaomi device. Additional testing by other Medtrum users would be appreciated to confirm correct behavior across different Android devices.

### DIFF
--- a/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/services/BLEComm.kt
+++ b/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/services/BLEComm.kt
@@ -187,7 +187,6 @@ class BLEComm @Inject internal constructor(
         } else {
             aapsLogger.debug(LTag.PUMPBTCOMM, "Not connected, closing gatt")
             close()
-            isConnected = false
             mCallback?.onBLEDisconnected()
         }
     }
@@ -198,6 +197,8 @@ class BLEComm @Inject internal constructor(
         mBluetoothGatt?.close()
         SystemClock.sleep(100)
         mBluetoothGatt = null
+        isConnected = false
+        isConnecting = false
     }
 
     /** Scan callback  */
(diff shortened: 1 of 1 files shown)


MilosKozak:
any testers?

pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/services/BLEComm.kt
```diff
@@ -387,11 +387,14 @@ class BLEComm @Inject internal constructor(
...
             }
-            close()
-            isConnected = false
-            isConnecting = false
-            mCallback?.onBLEDisconnected()
-            aapsLogger.debug(LTag.PUMPBTCOMM, "Device was disconnected " + gatt.device.name) //Device was disconnected
+            // CLosing asynchronously to avoid deadlocks or silent failures in the Android BLE callback thread during disconnect.
+            handler.post {
+                close()
+                isConnected = false
```
jbr7rr: Setting these 2 flags async can cause race conditions

jbr7rr:
Before applying this fix, did you test with scanning enabled? This setting: 
https://androidaps.readthedocs.io/en/latest/CompatiblePumps/MedtrumNano.html#scan-on-connection-error

This usually takes care of hardware specific connection issues

jbr7rr:
Before applying this fix, did you test with scanning enabled? This setting: 
https://androidaps.readthedocs.io/en/latest/CompatiblePumps/MedtrumNano.html#scan-on-connection-error

This usually takes care of hardware specific connection issues

HuismanErik:
Addressed the race condition raised by @jbr7rr: moved the flag resets into close() itself, which is already `@Synchronized`, so state is always updated with the GATT cleanup.

Regarding "scan on connection error": I tested this but it didn't help. The issue seems to be related to a stall on the Binder thread on some specific devices, which prevents onBLEDisconnected() from ever firing — scanning more aggressively didn't fix that.

HuismanErik:
@MilosKozak  Could you share why this was closed? This fix has been running stable on my device for 4+ months. Happy to address any remaining concerns.

PR #4604, opened 2026-03-12
Move BLE code for Omnipod DASH bledriver into interfaces

jwoglom:
This change moves the Omnipod BLE stack (for Omnipod DASH and, in the future, Omnipod 5) to use interfaces, separating business logic from direct calls to the raw Android BLE stack.

This keeps all platform BLE code behind interfaces so alternative implementations (e.g. different Bluetooth libraries) can be added later without touching core logic, while also making the code more testable.

The interfaces (BleCharacteristicIO, CmdBleIO, DataBleIO, PodScanner, BleConnection, BleConnectionFactory, BleDeviceManager) live under comm/interfaces/ with subfolders (io, scan, session, device). All Android BLE code is moved into comm/legacy/ using the same structure. 

As future work, I plan to develop an implementation of the BLE interface for DASH which uses [BLESSED](https://github.com/weliem/blessed-kotlin) -- a wrapper on top of raw Android BLE which may improve reliability for the DASH driver.

### Testing

I do not have DASH pods currently, so I cannot end-to-end test. However everything builds and the underlying code has just been copied over into new places, so there _should_ be no difference in runtime behavior. Before merging, I hope we can get confirmation from those with DASH pods that these code changes don't affect anything.


je-l:
Started a new Dash pod couple hours ago with this and the activation process, temp basal, bolus are working as usual. I will comment here if I see anything weird.

Device is Galaxy A52.

PR #4606, opened 2026-03-12
Fix hoursago format

Philoul:
String only used in Insight and Medtronic fragment...
Update of hourAgo with `inWholeHours` (Long instead of Double) has been done after my migration to Medtrum so I didn't saw this bug earlier...

### DIFF
--- a/core/interfaces/src/main/res/values/strings.xml
+++ b/core/interfaces/src/main/res/values/strings.xml
@@ -12,7 +12,7 @@
     <string name="minago">%1$dm ago</string>
     <string name="secago">%1$ds ago</string>
     <string name="minago_long">%1$d minutes ago</string>
-    <string name="hoursago">%1$.1fh ago</string>
+    <string name="hoursago">%1$dh ago</string>
     <string name="days_ago">%1$.1f days ago</string>
     <string name="days_ago_round">%1$.0f days ago</string>
     <plurals name="plurals_day_hour_ago">


andyrozman:
Shouldn't you rather fix the root cause? Not change precision of displayed hours? 

    override fun hourAgo(time: Long, rh: ResourceHelper): String {
        val duration = (now() - time).milliseconds
        val hours = duration.toDouble(DurationUnit.HOURS)   // applied fix
        return rh.gs(R.string.hoursago, hours)
    }

I sent this fix to Milos sometime ago. This method always displayed time in decimals, but now (bug first popped out in november or something) someone applied fix to show it in full hours, which changes implementation for everybody else... So if someone needs to have time displayed in whole hours, they should create new method hourAgoInWholeHours, if they needed it that way, but I think that this problem was introduced by some newbie (Gothraw) who didn't know what he was doing...

Philoul:
Ok with your comment...

I just discovered recently the inconsistency between value and string resource... (Insight pump restarted for tests in compose and hourago is used...)

andyrozman:
We are using this method also in tandem so we discovered in there (relatively soon after it was introduced - I am doing updates usually every month, to keep up with dev) and it was also reported for Medtronic few weeks ago... For me showing that bolus happened 1.4 h ago is better than 1 h ago.

Philoul:
This string is only used in Medtronic and Insight driver...
I think everything will probably be standardized soon for all pump drivers and show informations the same way 😉

andyrozman:
I hope standardisation will allow for some flexibility, while all drivers display 60% of stuff on all of them, other 40% are usually pump specific... Compose is wonderful with what you can easily do and also very flexible (all Tandem dialogs are already compose based, with exception of main fragment of course)...

MilosKozak:
so what to do with this PR?

andyrozman:
Revert the change to strings and fix it properly... See above code...

PR #4633, opened 2026-03-24
Add Instara CGM support.

leohunt:
Add files to support a new kind of CGM: Instara, with the following functions: 
1. Receiving glucose value data from Instara CGM.
2. History data consistency check, with user setting option(default as enabled).


olorinmaia:
Hi. Change target branch to dev.

MilosKozak:
Thank you for the contribution. However, I have concerns about extending the `GlucoseValue` database schema with new columns (`sgvId`, `sgvMark`) and indexes. Schema changes have wide-reaching consequences across the project (migration, schema validation, all modules touching GV).

**Suggested approach:**

1. **Use existing `InterfaceIDs.pumpId`** to store the Instara `sgvId`. This field is already a `Long?`, already indexed, and used as HW record id.

2. **Store `sgvMark` in `Preferences`** rather than in the database. It is a single value per device (not per glucose reading), used only by the periodic stale-check worker. It does not belong on `GlucoseValue`.

3. Add `INSTARA` to `InterfaceIDs.PumpType` enum (no migration needed).

This eliminates the v31→v32 migration, the new columns, the new indexes, the dedup SQL in migration, and the `TeljaneScopeRow` projection class — while preserving all functionality.

MilosKozak:
2nd) why Teljane vs INSTARA?

MilosKozak:
3rd) you wrote Patched Teljane app. If it's not an app from app store, create PR to wiki too, how to deal with this CGM

leohunt:
> 3rd) you wrote Patched Teljane app. If it's not an app from app store, create PR to wiki too, how to deal with this CGM

Thanks for pointing it out. Actually it is an app from Google play store. I think it's already released in Europe and it has plan to release in more countries and regions later. So it's not a patched app. I'll update the comments accordingly. 
Link: https://play.google.com/store/apps/details?id=com.teljane.instara

Besides, I would like to add "Instara" into the list of Compatible CGMs in wiki, after this feature was included in the next release of AAPS. Please let me know if I should update the wiki then.

leohunt:
> 2nd) why Teljane vs INSTARA?

This is more like company name vs application name, and the name "Instara" came in later in the development phase. I'll update the code and comments to "Instara" as this is the product name they want, to make it as consistent as possible. However, it will still have some parts refering to "Teljane", like the action name which can't be updated to preserve functionality.

leohunt:
> Thank you for the contribution. However, I have concerns about extending the `GlucoseValue` database schema with new columns (`sgvId`, `sgvMark`) and indexes. Schema changes have wide-reaching consequences across the project (migration, schema validation, all modules touching GV).
> 
> **Suggested approach:**
> 
> 1. **Use existing `InterfaceIDs.pumpId`** to store the Instara `sgvId`. This field is already a `Long?`, already indexed, and used as HW record id.
> 2. **Store `sgvMark` in `Preferences`** rather than in the database. It is a single value per device (not per glucose reading), used only by the periodic stale-check worker. It does not belong on `GlucoseValue`.
> 3. Add `INSTARA` to `InterfaceIDs.PumpType` enum (no migration needed).
> 
> This eliminates the v31→v32 migration, the new columns, the new indexes, the dedup SQL in migration, and the `TeljaneScopeRow` projection class — while preserving all functionality.

Thanks for the suggestion. I completely agree with the idea of avoiding schema changes. However, 

1. The `sgvId` is more like glucoseValueId, with its own defining logic. Are you suggesting that we "borrow" the existing `InterfaceIDs.pumpId` file, but use it to store `sgvId` information instead, for the stale-check worker function? 
2. The `sgvMark` is a single value per device. But it needs to be read from the `GlucoseValue` record and might change across different devices even within the same pluginType Instara (I'll double check with the manufacture). In this case, is storing it into `Preferences` the best option, in order to keep the periodic stale-check worker? 
3. Sure I can add "Instara" into `InterfaceIDs.pumpId`. But, Instara is a CGM device, not a pump (at least not working as a pump at this moment). Perhaps adding it into `InterfaceIDs.pumpId` will cause some confusion?

MilosKozak:
`pumpId` name was selected because we expected it will be used for pumps only. But generealy it's hardware id of selected record. So no problem to use it

leohunt:
> 3\. INSTARA



> `pumpId` name was selected because we expected it will be used for pumps only. But generealy it's hardware id of selected record. So no problem to use it

Thanks for the reply. In this case, 
1. I'll use the `pumpId` to store `sgvId`. 
2. Save `devicePrefix`, `sgvMark`, `sgvStart` into prefrence xml. As this information is related to the device, not glucose record. When the user changes to a new device, this part of information will be updated. 
3. I tested it can work without adding "INSTARA" into `InterfaceIDs.pumpId`. May I ask if there is a specific reason that we need to add INSTARA to InterfaceIDs.PumpType enum?

In all, I'll update the code to provide same function without DB schema changes.

MilosKozak:
> 3. I tested it can work without adding "INSTARA" into `InterfaceIDs.pumpId`. May I ask if there is a specific reason that we need to add INSTARA to InterfaceIDs.PumpType enum?
Maybe not if used only for CGM. You have to test

leohunt:
> > 3. I tested it can work without adding "INSTARA" into `InterfaceIDs.pumpId`. May I ask if there is a specific reason that we need to add INSTARA to InterfaceIDs.PumpType enum?
> >    Maybe not if used only for CGM. You have to test

Got it. Thanks for the reply.

leohunt:
Will submit a new PR with changes from suggestions and recommendations. Thanks!

PR #4635, opened 2026-03-26
fix(equil): adjust resistance threshold by firmware version

hhfcvmars:
1. increase BLE command spacing from 150ms to 500ms to improve the success rate of consecutive multi-packet transmissions
2. centralize resistance threshold logic in EquilManager
3. use firmware version and serial prefix to choose legacy or high threshold
4. reuse shared serial prefix checks in device support detection


MilosKozak:
@hhfcvmars I reverted this change in dev3 branch because it breaks filling on current pumps

MilosKozak:
create new PR if needed

PR #4645, opened 2026-03-28
Select ICfg in InsulinDialog (RecordOnly)

Philoul:
First proposal using `SelectInsulin` function
- Not fully satisfied by the result: There is no really safety issue here because insulin is always in IU whatever the concentration... So a simple DropDown is probably better... (the only risk is wrong peak/dia in IOB calculation if a wrong insulin has been selected...)


MilosKozak:
filtering by concenration is extra step i think. 
My proposal on top of this in insulin branch https://github.com/nightscout/AndroidAPS/tree/insulin
* split path for recording pump  and pen insuling
* pen path ... offer only U100 because that's what user see on pen

Issue #4613, opened 2026-03-16
[FEATURE REQUEST] NFC Commands similar to SMS Communicator


juehv:
I would like to propose an NFC-based commands plugin as an alternative to SMS commands.

 I already have a working implementation, heavily inspired by the existing SMS communication plugin, but adapted for local command triggering via NFC tags. The goal is to allow selected commands to be issued by scanning a configured NFC tag instead of sending an SMS.

In the implementation, I already considered the main security aspects:
- local, short-range interaction instead of remote SMS delivery
- explicit opt-in through a dedicated plugin
- restricted command handling based on a predefined command format (from SMS Communicator)
- use only with slef-written known NFC tags (leveraging JWT)
- keeping the same general command model and safeguards as close as practical to the SMS plugin

Before opening a PR, I want to confirm that this feature direction is acceptable for AndroidAPS.

If there are no major objections, I can open a PR for the `NfcCommunicator` plugin in the next few days.

juehv:
I created a Pull-Request here:
https://github.com/nightscout/AndroidAPS/pull/4655

Besides the unit tests, I tested the feature two weeks and it works fine for me. Due to safety and security, only tags written with the same phone work and only if the phone is unlocked.

juehv:
I added a branch for everybody to try the plugin based on AAPS 3.4.2.1 --> https://github.com/juehv/AndroidAPS/tree/patch-nfc

juehv:
See Pull-Request.

PR #4655, opened 2026-03-30
Implemented new plugin "NFC Commands"

juehv:
The PR is related to 
https://github.com/nightscout/AndroidAPS/issues/4613


MilosKozak:
From my point of view, this is very nice research project. But it's targeting just a few people, maybe. Nothing with mass usage. I keep this open for discussion.

juehv:
Thanks for the quick response. I am not entirely sure what you mean by “research.” I understood it as suggesting that this feature has limited practical value, and I would disagree with that assessment.

For me, this plugin provides a real improvement in daily use. Being able to trigger a command by tapping an NFC tag at places like the shower or kitchen table is much more convenient than navigating through multiple menus each time.

More broadly, AndroidAPS already includes a number of optional features that are only relevant to smaller groups of users. That is part of its strength: users can configure it to fit their own needs and workflows. I see the NFC Commands plugin in the same category.

From an engineering perspective, the implementation is cleanly separated through the existing plugin system and follows the same command model and safeguards as closely as practical. In my view, it is a safe and maintainable addition.

Unfortunately, I do not have the time to actively promote the plugin outside the normal contribution process. If it is not accepted into the main repository, and later documented in Read the Docs, most users will likely never discover it, since many do not regularly follow issue discussions.

If there are specific technical, safety, or maintenance concerns, I would be happy to discuss them and improve the implementation where needed.

MilosKozak:
That's why keeping this open for discussion. If there will be more people interested, I can change my mind ...

lightswitch2159:
A shower NFC tag to trigger a 30 min disconnect would be great, I often just don't trigger a disconnect due to it being annoying to do, or not thinking about it. Simply hitting a NFC tag in my bathroom would be _huge._

xitation:
I use NFC tags with Home Assistant already, the ability to trigger things in AAPS would be handy with tags around the house. Start the breakfast/meal pre-bolus or temp target with a tap would be ace, rather than using time based that becomes hard if you sleep in etc.

juehv:
I improved the UX a bit. Also, I added a branch for everybody to try the plugin based on AAPS 3.4.2.1 --> https://github.com/juehv/AndroidAPS/tree/patch-nfc

Philoul:
I have hundred of old FSL sensors (I had FSL v1 with a miaomiao transmitter a long time ago 😉)

All FSL sensors includes NFC communication.

Does your plugin work with this kind of NFC tag?
- if yes you could have thousands of potential users for your plugin (it could be a simple way to re-use old sensors, FSL 2 also use NFC for activation...)

juehv:
> I have hundred of old FSL sensors (I had FSL v1 with a miaomiao transmitter a long time ago 😉)
> 
> All FSL sensors includes NFC communication.
> 
> Does your plugin work with this kind of NFC tag?
> 
>     * if yes you could have thousands of potential users for your plugin (it could be a simple way to re-use old sensors, FSL 2 also use NFC for activation...)

This is a very interesting idea!
I don't know what kind of NFC tag the FSL 1/2/3 use. From the top of my head I remember that FSL1 was quite exotic at that time. I quickly tried on my FSL 3, but it's not working out of the box. Also I don't know if they are write-able at all (as they are not intended for that ... I could imagine that Abbot made them read-only). 

I will definitely dig into that. But on the other hand, the set of 30 NTAGS in sticker size is around 5 EUR... So the question for a user is more "Does it relax my daily routine or not". And for me it's a game changer: I have one at my shower to disconnect pump, one at my kitchen table to set target eating soon and the same in my office on the door. Since I have them, I not only have to do less menu clicking, but also with the tags in my sight I forget it less often.

Philoul:
I tried to install on my dev config to test your PR but it still requires a few updates:
- a few merge conflicts on latest dev (easy to fix)
- but preference management should be reworked
- User Interface no more use fragment but compose code


For read only NFC tags, if you have a unique ID available for each tag, do you need write access?
- what you store within the tag could be stored in preferences (i.e. json string with tag ID and associated commands) 🤔 (just personal thoughts 😉)

Philoul:
I dive a bit on NFC tags (I build my first transmitter with an ESP32 for BT communication associated with a BM019 hardware for NFC communication 7 years ago...)
https://github.com/Philoul/LimitTer-ESP32/tree/master

Communication protocol of NFC 1 and 2 is ISO/IEC 15693
- NFC 1 probably had write access (for most of them) but are no more used
- NFC 2 has been fixed by Abbot to remove all write capabilities (and crypto layer added)

It could probably work if we limit communication to only read the UID of the FSL sensor and nothing else... (then everything should be stored on phone side associated with the UID)...

juehv:
> I dive a bit on NFC tags (I build my first transmitter with an ESP32 for BT communication associated with a BM019 hardware for NFC communication 7 years ago...) https://github.com/Philoul/LimitTer-ESP32/tree/master
> 
> Communication protocol of NFC 1 and 2 is ISO/IEC 15693
> 
>     * NFC 1 probably had write access (for most of them) but are no more used
> 
>     * NFC 2 has been fixed by Abbot to remove all write capabilities (and crypto layer added)
> 
> 
> It could probably work if we limit communication to only read the UID of the FSL sensor and nothing else... (then everything should be stored on phone side associated with the UID)...

I agree this should work fine, but this is quite a bit less secure. At the moment I write a JWT that includes the ID of the token on the token. It has a limited lifetime which gets automatically renewed every now and then. That means I'm always 100% sure the token is placed and written by me, and no other person could change the command that is executed.

For me this feels quite secure (and safe in a consequence). I think it won't be too complicated to add a read-only option but it's hard to communicate non technical people the difference... Or do you think that's non of an issue?

juehv:
> I tried to install on my dev config to test your PR but it still requires a few updates:
> 
>     * a few merge conflicts on latest dev (easy to fix)
> 
>     * but preference management should be reworked
> 
>     * User Interface no more use fragment but compose code
> 
> 
> For read only NFC tags, if you have a unique ID available for each tag, do you need write access?
> 
>     * what you store within the tag could be stored in preferences (i.e. json string with tag ID and associated commands) 🤔 (just personal thoughts 😉)

Ah yes, I saw the GUI is changing quite a bit. I developed the plugin before the last merge.

Baseline question is, will the plugin be merged or not. Because if I'm the only user I'd rather target the master branch, as I'm using the plugin every day.

Philoul:
> Baseline question is, will the plugin be merged or not. Because if I'm the only user I'd rather target the master branch, as I'm using the plugin every day.

All new features will now be targeted in v4.0 and following versions.
On current master (TBC by @MilosKozak ), I think only bug fix will be included.

You can of course continue to use it in your local version with v3, but your PR should be updated with V4 for a potential merge.


> That means I'm always 100% sure the token is placed and written by me, and no other person could change the command that is executed.

Not sure to understand correctly the way it works currently, and security impact (better or worse) with a read only tag (if we keep the idea to re-use old FSL sensor) 🤯.

In your current solution, where are the commands associated to a tag? Within the tag or within AAPS?
- My idea was to just read the NFC tag UID. That way it's 100% secure (to modify something a hacker should hack your phone, but nothing can be done on the tag 🤔)... Am I wrong ?

juehv:
> All new features will now be targeted in v4.0 and following versions. On current master (TBC by @MilosKozak ), I think only bug fix will be included.
> 
> You can of course continue to use it in your local version with v3, but your PR should be updated with V4 for a potential merge.

Ok thanks for clarification.


> Not sure to understand correctly the way it works currently, and security impact (better or worse) with a read only tag (if we keep the idea to re-use old FSL sensor) 🤯.
> 
> In your current solution, where are the commands associated to a tag? Within the tag or within AAPS?
> 
>     * My idea was to just read the NFC tag UID. That way it's 100% secure (to modify something a hacker should hack your phone, but nothing can be done on the tag 🤔)... Am I wrong ?

Currently the command is written on the tag as part of the JWT token. However, the phone manages a redundant database of the commands on the tags (but only for reference).

Changing to only reading the tag id and ignoring any content requires the command chain to be on the phone. If we consider the phone as "safe", both designs are equally secure (according to a security expert colleague of me). 

If you think that provides a value to someone I would spend a weekend to change the implementation (along with some other items on my todo list). Would be great if you can check if that works well with the FSL1 tags.

Philoul:
@juehv I just installed "NFC Tools" in my phone to scan my different old FSL sensors 😯(I'm really a rooky on this topic 😂)
- with NFC Tools, I can read all my old FSL sensors, but I can't write anything on them
- they all have a unique serial number

I now understand a bit better the standard way to use NFC tags (writing commands on tag), but information stored within NFC tags should not be "sensitive" on a safety/security point of view. For me it's not the case concerning an AAPS command (deliver bolus, stop loop, etc are in my mind very sensitive so they shouldn't be stored within tags).

I think that keeping every sensitive information within AAPS is probably easier (code simplified) and much more secure:
- no need to implement a security layer concerning tags (password management for tags, read /write information within tag, crypto layer, lock tag as read only 🤔...)
- a simple list of map of tag UID and associated commands can be stored (as a json string) within AAPS preferences
- when you have several diabetic persons in the same family, each one can have their own commands for the same tag

Just my thoughts 😉

juehv:
> @juehv I just installed "NFC Tools" in my phone to scan my different old FSL sensors 😯(I'm really a rooky on this topic 😂)
> 
>     * with NFC Tools, I can read all my old FSL sensors, but I can't write anything on them
> 
>     * they all have a unique serial number
> 
> 
> I now understand a bit better the standard way to use NFC tags (writing commands on tag), but information stored within NFC tags should not be "sensitive" on a safety/security point of view. For me it's not the case concerning an AAPS command (deliver bolus, stop loop, etc are in my mind very sensitive so they shouldn't be stored within tags).
> 
> I think that keeping every sensitive information within AAPS is probably easier (code simplified) and much more secure:
> 
>     * no need to implement a security layer concerning tags (password management for tags, read /write information within tag, crypto layer, lock tag as read only 🤔...)
> 
>     * a simple list of map of tag UID and associated commands can be stored (as a json string) within AAPS preferences
> 
>     * when you have several diabetic persons in the same family, each one can have their own commands for the same tag
> 
> 
> Just my thoughts 😉

Ok then let's do it. This weekend is already full but next week I have the time to do it. Although I still have to figure out how the new UI is working ... but I'll manage :D

I'll update the existing pull request as soon as I'm done.

xitation:
Yeh I agree, from a security perspective the only think on the tag should be a unique identifier that AAPS can read to trigger an action / automation. All the logic on this should be within AAPS i'd say, but configurable on what the string/code/hex value is that is read from the tag. Then you could really use any tag that already has a readable value on it to trigger that automation without the need to write anything to a tag (unless you wanted too for some reason).

Ideally when setting up the automation you could tap the tag to read the value into the trigger field within that automation to make easy of use better?

juehv:
@Philoul I worked this weekend on the NFC Plugin and run into a potential issue:

What if FSL is used as glucose source and a new tag is scanned?

Android's passive NFC dispatch follows a strict priority chain:
NDEF_DISCOVERED > TECH_DISCOVERED > TAG_DISCOVERED.
Registering ACTION_TAG_DISCOVERED is safe, it only fires when no higher-priority app claimed the tag, so it cannot break CGM sensor reading by apps like LibreLink (which win via TECH_DISCOVERED).

Proposed approach (three layers):
1. Keep ACTION_NDEF_DISCOVERED + vnd.app.aaps.command for tags written by AAPS. The NDEF payload no longer carries the command — it only carries a tag ID that maps to a command stored on the phone. This preserves the tight dispatch (only AAPS-written tags trigger the activity) while moving the command definition off the physical tag.
2. Add ACTION_TAG_DISCOVERED as a fallback to support arbitrary tags (blank tags, finished Libre sensors, etc.) identified purely by UID. Caveat: if another app has a matching TECH_DISCOVERED filter for that tag type (e.g. LibreLink for NfcV sensors), it wins and AAPS does not receive the scan. This is intentional, we prefer that over breaking other services.
3. Foreground dispatch in the main UI (opt-in setting): when the user enables it and AAPS is in the foreground, the app registers NfcAdapter.enableForegroundDispatch() which has the highest priority and intercepts any tag before LibreLink. This allows scanning NfcV tags (finished Libre sensors) even with LibreLink installed, at the cost of requiring AAPS to be the active screen (and modification of main activity) -> less convenient, but functional.

What are your thoughts on this?

Philoul:
I just tried your plugin on my dev phone:
Impossible to scan one of my old FSL to register it:
- I move slowly the old libre on the back of my phone, but nothing is recognized...
- If I hit the android button to switch to another app, I immediatly see an Android Popup to select either eDropletNFC, LibreLink or xDrip+ (AAPS is not proposed among the different apps)

Philoul:
When I disabled NFC command from xDrip, it is no more proposed within the different Apps
- The setting Enable NFC within AAPS doesn't change anything (I of course set it enabled), but AAPS is not proposed and NFC is not recognized...

See below a screenshot of NFC type from NFC Tools...

juehv:
Yeah .. that's I did not implement the support for non NDEF tags yet :sweat_smile: 
I made the post to get your opinion if you find that an acceptable behavior.

But the UI migration is done and the UID only feature is in place for NDEF tags. So multiple user could use the same tag with different actions.

Philoul:
Second test with a spare tag I have (from an old Arduino kit...)
- Record worked as expected
- I understood the setting (not here to record AAPS within Android as an NFC potential app, but here to allow or not command execution)

I have several little comments (UI improvement, code, ...) but I was able to test it and it worked as expected.

FSL users should probably also test it: Normally NFC is only used to start the sensor and not later, but I don't know if we can disable NFC to let AAPS use older tag without conflicts... And it could vary according to user configuration (juggluco, xDrip+ and LibreLink, ...)

juehv:
@Philoul 
Ok I think it's done! Maybe you can give it another try, I implemented it as described above and did some UI polishing. Let me know what you think :)

Philoul:
Sorry for later answer (I forgot to send it 5 days ago 😯)

Tested with latest updates (be careful, `LocalProfileManager` has been renamed `ProfileRepository` so I first had to fix code to be able to build and test with latest dev 😉)

- it still doesn't work with my old FSL sensor.AAPS is not proposed as a compatible App for this tag: when I try to register tag is not visible, and when I click on Android button to switch to another app, then I immediately see a popup to propose me eDropletNFC or LibreLink apps to read the tag). Probably something is missing in manifest 🤔.

My other comment concerning code
- you put your plugin in `main` module. It should probably be more within `sync` module in my mind (@MilosKozak ?)
- within My Tags tab, the edit icon only allow you rename the tag, but not to edit commands (now we have all commands within the phone, don't you think it could be nice to be able to edit the command attached to this or this tag 🤔)
- user interface to add Commands is better, but should still be improved (we can probably re-use or be similar to Automation action selection)
- user interface for value adjustments (Carbs, Bolus, Profile Switch...) should probably be harmonized with the other screens (i.e. for bolus the +5, +10, +20 buttons and editable area).
- in My Tags tab, in each tag tile you show the tag UID (minimum is a little label, but I'm not sure this information is relevant for end user).

Don't spend to much time on UI (it's enough to test the feature and UI can be improved later 😉)

juehv:
Haha happens to me all the time :sweat_smile: 

> you put your plugin in main module. It should probably be more within sync module in my mind (@MilosKozak ?)

When I started SMS Communicator was in main and therefore I placed it there too. I can move it wherever you want it, I don't have any constrains with that.

> user interface to add Commands is better, but should still be improved (we can probably re-use or be similar to Automation action selection)
user interface for value adjustments (Carbs, Bolus, Profile Switch...) should probably be harmonized with the other screens (i.e. for bolus the +5, +10, +20 buttons and editable area).

I build the user interface following the current UI standards and adding short buttons (+5 for Bolus etc) for a seldom used interface feels strange (and violates current standards). But I think UI discussion is for a later time (as you said)

>  in My Tags tab, in each tag tile you show the tag UID (minimum is a little label, but I'm not sure this information is relevant for end user).

True, but does it hurt? I like to know it.

> within My Tags tab, the edit icon only allow you rename the tag, but not to edit commands (now we have all commands within the phone, don't you think it could be nice to be able to edit the command attached to this or this tag 🤔)

I don't see the point. You can just create a new one and assign the tag. This will delete the current command chain.

> it still doesn't work with my old FSL sensor.AAPS is not proposed as a compatible App for this tag: when I try to register tag is not visible, and when I click on Android button to switch to another app, then I immediately see a popup to propose me eDropletNFC or LibreLink apps to read the tag). Probably something is missing in manifest 🤔.

I try to get my hands on an old FSL1 tag and will investigate as soon as I have one. For the meantime all other relevant tags are working fine (NDEF as de-facto standard and the Mifare variants which are the super cheap sticker things). From my perspective the usage of FSL 1 tags is just a nice gimmick but if someone wants to use the plugin, buying a pack of tags for 3$ is the better deal.

Philoul:
Additionally, build doesn't work any more with latest dev (most commandQueue are in suspend mode...

juehv:
Is it worth fixing it? Or does it make sense to wait until the architectural changes are finished?

Philoul:
It's generally easier fix build issues regularly.
Nobody will test a branch with build issues.
And early tester of dev branch are with latest version.
- the more you wait the more it's difficult to fix everything.

juehv:
I understand your point, but I see it a bit differently. As a professional developer, I’m not concerned about whether the changes are easy or difficult. Whatever changes are required will be manageable.

That said, I may not have been clear enough about my intention in my previous question, so let me clarify.

I’m very happy to make a meaningful contribution to the project. I’m also completely fine with following the coding conventions and architectural decisions you have established, since this is your software.

What I would like to avoid is having to update the pull request every couple of days because of architectural changes, renamings, or other breaking changes elsewhere in the project. Each time I update the pull request, I make sure that it compiles, that the tests pass, or at least that I have not broken anyone else’s tests, and that my own tests pass. I also manually test all plugin features. That takes a fair amount of time. From my perspective, doing this repeatedly while the surrounding architecture is still changing creates overhead that may become obsolete shortly afterward.

So my actual question was: are there any further breaking changes planned? If so, I would prefer to wait until they have been pushed before updating the pull request again.

Regarding testing, I respectfully disagree with your argument. Anyone who wants to try the feature can already do so by using either my dev branch or my master branch patch. I therefore do not see much value in integrating it into the current dev state just for testing purposes.

vanelsberg:
> FSL sensors

I have some to. But I do not use them (to big).
But just go searching AliExpress: NFC tags is all formats (tags, stickers, ..) that are perfect for this. Costs are minimal.

vanelsberg:
> Regarding testing....

Just found 2 patch branches on your AAPS repo: are these working patches for latest AAPS V3 release versions? 
If so I may give it a try: wicht branch?

juehv:
> > Regarding testing....
> 
> Just found 2 patch branches on your AAPS repo: are these working patches for latest AAPS V3 release versions? If so I may give it a try: wicht branch?

Here is the link:
https://github.com/juehv/AndroidAPS/tree/patch-nfc

It is based on the AAPS 3.4.2.1 (most probably I will update it to 3.4.2.2 this weekend)

vanelsberg:
> It is based on the AAPS 3.4.2.1

Tnx!
Will have a look.

Unless you have reasons to build on master release version, maybe base it on the dev3 branch?

juehv:
The other branch is based on the dev branch and I will update it to current version when I find a longer time slot to take care about the conflicts. --> https://github.com/juehv/AndroidAPS/tree/patch-new-plugin-nfc_commands

Philoul:
@juehv 
FYI, I made an exercice to be able to continue the tests on my dev configuration...
- As I explained earlier, several early tester of AAPS V4 synchronized several time per week their version with latest dev, and several of them on a real loop configuration (connected to the body). But unfortunatly very few of them will be able to test a branch with conflicts and build issues, fix issues, etc.
- even those using a dedicated dev configuration with a test phone don't install outdated version of dev branch...
- On my side I spend a bit more time to continue testing on my dev phone (and even for testing, I always synchronised my development configuration with latest dev branch).

I share with you what I did on my side (be carefull, some modifications has been done directly, but I also used Gemini to help me, especially where I'm not skilled enough to analyze and check the results, so consider the working branch below as a POC for testing, and not as a robust code)...
- I merged latest dev to your working branch to fix conflicts
- I updated the command functions within plugin to be able to build the app
- I added NFcV compatibility for testing with my hundred of Free Style Libre patch (100% vibe coding) (it works on my dev phone with my old FSL sensors)
- I included an additional setting to propose AAPS as a potential app for tags scanned with AAPS is in background (if disabled, the default app is launched directly, but if enabled, Android propose to launch AAPS or xDrip or LibreLink or any other app) (100% vibe coding)

For me what I did is probably not optimized and not very robust (several crashes before having something working as expected...).
- It's just an exercice on my side so I keep this working branch in my repo and share it with you if it can help you (or not...), so feel free to do what you want with it!
https://github.com/Philoul/AndroidAPS/tree/new-plugin-nfc_commands_fix-build

juehv:
@Philoul 
I got my hands on a FSL 1 tag. With that I managed to fix the issue. It is also working with FSL3 out of the box (seems to be the same tag standard).

Also I merged with current dev branch and fixed the issues you described above. Hope this reduces friction.

vanelsberg:
> @Philoul I got my hands on a FSL 1 tag. With that I managed to fix the issue. It is also working with FSL3 out of the box (seems to be the same tag standard).

Just installed @Philoul's (Tnx for porting this to v4!) [dev build](https://github.com/Philoul/AndroidAPS/tree/new-plugin-nfc_commands_fix-build) on my test phone.

Initial impression: this is really an awesome feature @juehv ! 

Tested with some tags I have here to set/unset Eating soon TT/Profile:
-r/o FSL2: Works OK!
-Writable key-chain tag: Works OK!
-Writable NFC-Card tag: Works OK!

Would love to have this on the current dev branch (AAPSv4) to continue testing on a daily basis and finalize UI and features.

vanelsberg:
@juehv @Philoul - Would be nice if we could a draft PR on dev to test/merge this with latest dev (v4).
This way it can be easily merged using AAPS-CI for early testers?

Philoul:
@juehv I will propose you a PR on your working branch to update the User Interface.

Concerning Profile Switch, the command must be updated: we cannot use the profile index as a reliable key to manage a profile switch, because if you delete a profile, index can change so the wrong profile will be applied, so for me the unique key to be included within the command must be the Profile Name.

vanelsberg:
> Also I merged with current dev branch and fixed the issues you described above....

Have not seen a new or updated draft PR on dev for this. Any progress?
Eager to have us start development/testing of your new plugin for AAPS v4 ☺️

juehv:
@vanelsberg 
I merged  @Philoul 's ui pull request and tested everything. Hope that helps to merge it to v4 test branches.

vanelsberg:
> ... merged @Philoul 's ui pull request and tested everything

Top! Will have a look at it tomorrow to see if I can merge this with latest dev (v4) code.
If it is good enough you could do your first PR for AAPS v4 for other dev's to test 🙂

Philoul:
For @MilosKozak 

Main updates done on this PR:
- plugin architecture is now very close to Automation or SMS plugin concerning actions management (each action included into a dedicated class)
- command string replaced by a json string (more flexible, especially for parameters like profile name, UI with icons...)
- uel added with a dedicated NFC icon for source
- User Interface improved to be very close to Automation UI.
- actions list is unchanged (only TBR absolute and % is now filtered according to pump).
- I added a mandatory confirmation popup before command exec in my commit, but @juehv replaced it by an optional confirmation popup... I'm afraid this "optional" side is a blocking point for a potential merge 😱 (currently nowhere else within AAPS a bolus, profile switch, TT, etc. can be applied or delivered without an explicit confirmation for safety and end user responsability).

I found this proposal a really nice feature 👍.
Even if writable tags can be found at a very low price, probably few users will buy them, but the possibility to reuse old FSL sensors increase much more the number of potential users (probably about half AAPS users have FSL 🤔 so these sensors can have a second life after usage with this plugin).

Remaining topics:
- plugin is still in main module (maybe better in sync module for module consistency?)
- actions list can be adjusted a bit (Temp Profile Switch, Custom TT, Scene actions are missing, maybe a few others should be removed) but it's minor and very easy to update

In my mind the most important point is the confirmation popup.
- I proposed it for all commands (to be consistent with AAPS UI)
- @juehv propose a setting to bypass for all commands
- other possible solution (mandatory for a few commands like boluses, optional for others) ?

juehv:
> For @MilosKozak
> 
> Main updates done on this PR:
> 
>     * plugin architecture is now very close to Automation or SMS plugin concerning actions management (each action included into a dedicated class)
> 
>     * command string replaced by a json string (more flexible, especially for parameters like profile name, UI with icons...)
> 
>     * uel added with a dedicated NFC icon for source
> 
>     * User Interface improved to be very close to Automation UI.
> 
>     * actions list is unchanged (only TBR absolute and % is now filtered according to pump).
> 
>     * I added a mandatory confirmation popup before command exec in my commit, but @juehv replaced it by an optional confirmation popup... I'm afraid this "optional" side is a blocking point for a potential merge 😱 (currently nowhere else within AAPS a bolus, profile switch, TT, etc. can be applied or delivered without an explicit confirmation for safety and end user responsability).
> 
> 
> I found this proposal a really nice feature 👍. Even if writable tags can be found at a very low price, probably few users will buy them, but the possibility to reuse old FSL sensors increase much more the number of potential users (probably about half AAPS users have FSL 🤔 so these sensors can have a second life after usage with this plugin).
> 
> Remaining topics:
> 
>     * plugin is still in main module (maybe better in sync module for module consistency?)
> 
>     * actions list can be adjusted a bit (Temp Profile Switch, Custom TT, Scene actions are missing, maybe a few others should be removed) but it's minor and very easy to update
> 
> 
> In my mind the most important point is the confirmation popup.
> 
>     * I proposed it for all commands (to be consistent with AAPS UI)
> 
>     * @juehv propose a setting to bypass for all commands
> 
>     * other possible solution (mandatory for a few commands like boluses, optional for others) ?

For clarification, two points from my side:

1. **FSL sensors as NFC tags are already working with the current code.** I tested it with FSL 2 and 3 sensors. Both are working fine. Moreover also the dirt cheap Mifare Classic tags are supported now.

2. **About the confirmation dialog:** I understand the safety concern, and that is why I left the confirmation dialog in place and enabled it by default. However, I strongly disagree that it must be mandatory in all cases.

Unlocking the phone and scanning a physical NFC tag is already an intentional user action. It requires several deliberate steps, physical access to the phone, and the tag must be scanned on purpose. This is not something that can realistically happen accidentally. Because of that, an additional mandatory confirmation dialog does not add meaningful safety in my opinion. It mainly adds friction and results in a worse user experience.

For me, the important point is that the confirmation dialog can stay enabled by default, but there must be an option to disable it. Without this option, the feature becomes unnecessarily cumbersome for my use case, and I would need to maintain a separate branch, which I would really like to avoid.

vanelsberg:
> For me, the important point is that the confirmation dialog can stay enabled by default, but there must be an option to disable it

Agree on that. Need for confirmation should depend on task to perform.
From a usability standpoint using NFC otherwise has no advantage over just opening AAPS and execute the task directly.

vanelsberg:
- Some action could be removed, other added.
- Additional, once this PR is in place it would be interesting to see if we could add functionality to the Wear integration to scan NFC tags directly from the watch, without the need to take out the phone.

Philoul:
> This is not something that can realistically happen accidentally. Because of that, an additional mandatory confirmation dialog does not add meaningful safety in my opinion

You are talking about you, but it's impossible to be 100% sure for all users in all situations.

Just an example: during my tests, one tag defined for a breakfast management was on my desk, invisible below a sheet of paper (important amount of carbs and important associated bolus as command actions). I suddenly discovered looking at AAPS overview a huge huge amount of COB and a huge huge amount of IOB 😱. Hopefully it was my dev phone and not my loop phone otherwise I would not have been here to answer you!
Why? Just because several boluses and carbs (not only one, but 3 in my case) has been automatically recorded and delivered by the tag that was below the paper. And it was definitely unexpected by me! No alerts, no popup, just delivered and recorded silently within the app! It's for me a strong "No-Go" (hundred complains against UI are much better than at least one complain because of coma or worse!)

Nowhere else within AAPS it's possible to deliver a bolus automatically without an explicit popup, and I think it should be the same here without any bypass.

I opened in my answer a third bullet point to make the bridge with another question recently asked on aaps4-testing discord channel concerning AAPS User Interface:
- can we authorize less clicks for "less risky actions" and sometimes bypass the confirmation popup?
- or should we keep the global UI consistency and keep confirmation popup everywhere like it is currently?

But this question has a wider impact on AAPS User Interface and cannot be managed locally in one plugin here in an inconsistent way with the overall app...

Just my thoughts 🤔😉

juehv:
> You are talking about you, but it's impossible to be 100% sure for all users in all situations.
> 
> Just an example: during my tests, one tag defined for a breakfast management was on my desk, invisible below a sheet of paper (important amount of carbs and important associated bolus as command actions). I suddenly discovered looking at AAPS overview a huge huge amount of COB and a huge huge amount of IOB 😱. Hopefully it was my dev phone and not my loop phone otherwise I would not have been here to answer you! Why? Just because several boluses and carbs (not only one, but 3 in my case) has been automatically recorded and delivered by the tag that was below the paper. And it was definitely unexpected by me! No alerts, no popup, just delivered and recorded silently within the app! It's for me a strong "No-Go" (hundred complains against UI are much better than at least one complain because of coma or worse!)
> 
> Nowhere else within AAPS it's possible to deliver a bolus automatically without an explicit popup, and I think it should be the same here without any bypass.

I understand your safety concerns, and I agree that this kind of functionality needs to be treated very carefully.

That said, your example seems to describe a very specific and unusual situation rather than the normal intended use of NFC tags. If a user says, “I know where my tags are, I use them deliberately, and I want to disable the confirmation popup,” then the responsibility is different from an unintended background automation.

I am not suggesting that the popup should be removed for everyone or disabled by default. My point is that it should be possible as an explicit opt-in option, with a clear and strong warning when enabling it. Users who consider this too risky can simply leave the confirmation enabled.

Also, there is already feedback when an NFC action is triggered: the phone vibrates, shows a toast, and opens the app. And importantly, this only works with an unlocked phone. So while accidental scans are possible, this is still not the same category as a remote or fully automatic action running without any local user interaction.

For that reason, I do not think NFC tag scanning should be treated the same as real automatic or remote bolus options. If the project policy is “no bolus action without an explicit confirmation popup under any circumstances,” then that is a clear safety policy and should be stated as such. But I would not justify it by calling NFC scanning an automatic action, because it still requires physical, local interaction with an unlocked phone.

So my preferred solution would be: keep the popup enabled by default, show a strong warning before allowing it to be disabled, and let advanced users explicitly opt in if they accept that risk.

Philoul:
@juehv No chance to convince me on this topic.
In safety, the question is not if a possible human error will occur or not, it will!
If a parameter exists to disable the popup it will be disabled, and if a human error is possible, then we will have them. So the real question is more how many times per day or per week one of the hundred or thousand users will have disabled the popup, scan a wrong tag or put his phone too close to a recorded tag and an unexpected command will be launched, with the corresponding consequences.

I don't know how many times you have clicked on the "Cancel" button of the current existing confirmation popups in AAPS, but on my side it's several times per week and probably a few hundred times per year.

AAPS should never include possible bypass on safety topics. If a user disagree and have the skills to modify the code, it's under his own responsability, but for a plugin included into AAPS, I'm still convinced the confirmation popup is mandatory without any bypass for end users.

I still think your NFC plugin is really a nice feature and would like to maximize the chance to have it included within AAPS app 😉.

juehv:
> @juehv No chance to convince me on this topic. In safety, the question is not if a possible human error will occur or not, it will! If a parameter exists to disable the popup it will be disabled, and if a human error is possible, then we will have them. So the real question is more how many times per day or per week one of the hundred or thousand users will have disabled the popup, scan a wrong tag or put his phone too close to a recorded tag and an unexpected command will be launched, with the corresponding consequences.
> 
> I don't know how many times you have clicked on the "Cancel" button of the current existing confirmation popups in AAPS, but on my side it's several times per week and probably a few hundred times per year.
> 
> AAPS should never include possible bypass on safety topics. If a user disagree and have the skills to modify the code, it's under his own responsability, but for a plugin included into AAPS, I'm still convinced the confirmation popup is mandatory without any bypass for end users.
> 
> I still think your NFC plugin is really a nice feature and would like to maximize the chance to have it included within AAPS app 😉.

Thank you for the clear statement.

It looks like we at least found some common ground: the NFC plugin is not an automation. It is a conscious user action. The user intentionally unlocks the phone, brings it very close to a tag, and triggers a defined action.

That said, I have to be very direct: I still do not find your argument convincing.

What you describe feels to me like pushing the discussion into a theoretical worst-case “commercial closed loop” territory. AAPS users are not random consumers blindly pressing buttons. This is a community of people who know that diabetes is a serious disease and that treatment decisions require care. And the potential user group for this plugin is even narrower: tech-affine users within an already tech-affine community, using technology intentionally to make daily diabetes management easier.

Comparing this to clicking “Cancel” on existing popups does not really fit, in my opinion. A much more fitting comparison would be entering `05` instead of `0.5` in the bolus dialog, then not properly reading the resulting confirmation popup because of warning fatigue, and then having a very bad day. If we strictly follow your argument, then AAPS itself becomes unacceptable, because it is inherently powerful and therefore inherently risky software.

To be clear about the likelihood of the NFC error scenario: this can only happen if someone disables the confirmation, writes a tag, puts it somewhere where they cannot clearly recognize it, forgets about it, unlocks the phone, keeps the screen active, brings the phone very close to that tag, ignores the vibration and AAPS opening, and only then the unwanted action is triggered. To me, this is an extremely unlikely chain of events, and I do not see it as a strong factual argument for forcing a confirmation popup in every case.

My position is also very clear: I built this plugin to reduce unnecessary user interaction and make daily use easier. I thought it would be nice to share it with the community and avoid maintaining a separate branch. I did not build it to create a complex scan process followed by yet another mandatory interaction that contributes to warning fatigue and turns the plugin into something I personally would not want to use.

So my final statement is this: if the confirmation popup, or any similar regular non-essential warning or user interaction, is mandatory and cannot be disabled by the user, then the plugin loses its main value for me. At that point, it is probably not worth the time to develop, test, or maintain it.

I still think the plugin is a useful and nice feature, and I would really like to maximize its chance of being included in AAPS. But the main purpose is to reduce friction, not to move the same friction to a different place.

vanelsberg:
> So my final statement is this: if the confirmation popup, or any similar regular non-essential warning or user interaction, is mandatory and cannot be disabled by the user,

Understand your reasoning. But as this is about safety (agree on @Philoul 's concerns on this):
How about only making mandatory confirmation popup depending on the task to perform? Some tasks really need user confirmation, others involving no immediate risk could do optionally without?

vanelsberg:
@juehv @Philoul As an alternative: how about limiting the execution of NFC commands to only once during a set timeframe?

F.i: First NFC command does not require confirmation. Any next execution of the same command would require user confirmation when its most recent execution was within the timeframe set?

juehv:
> @juehv @Philoul As an alternative: how about limiting the execution of NFC commands to only once during a set timeframe?

I think this does not help in either direction: in my world it is still annoying, in @Philoul 's world it is still unsafe.

> Understand your reasoning. But as this is about safety (agree on @Philoul 's concerns on this):
How about only making mandatory confirmation popup depending on the task to perform? 

For me that is not good enough. I use a 2U Pre-Bolus when I exit my office on the way to lunch. I don't want to interact with UI there, otherwise I could directly use the UI.
Safety is not the "excuse" for everything. We all use the autonomous bolusing feature of AAPS (SMB) that is completely automated. This is inherently unsafe and I find it strange to "trust" the user in the one feature and don't in the other...


But talking about compromises: I am a big hater of the objectives... However, maybe we can add an objective that must be fulfilled before deactivating the popups. That give the user a training to take extra care while using NFC tags. Is this s.th. you can live with @Philoul ?

vanelsberg:
> We all use the autonomous bolusing feature of AAPS (SMB) that is completely automated.

SMB is not "autonomous bolusing" as you put it. It's another way to do basal insulin and folows the oRef algo safety gards.

Also in my opinion, not having user confirmation on certain NFC commands is a safety issue. Without proper measures I do not think this PR will ever be accepted in release code...

juehv:
> SMB is not "autonomous bolusing" as you put it. It's another way to do basal insulin and folows the oRef algo safety gards.

I’m very familiar with the algorithm, as I was active in the dev community back then. In practical terms, it is de facto autonomous bolusing; I don’t think there is another accurate way to describe it. It may feel similar to basal behavior, but it is not the same.

That said, this is a separate discussion. My main point is that there are more serious safety concerns in AAPS than allowing user interaction without repeated pop-ups.

> Also in my opinion, not having user confirmation on certain NFC commands is a safety issue. Without proper measures I do not think this PR will ever be accepted in release code..

Ok this is good to know. I think the different opinions are clearly stated now. Is there a way to get a final decision?

Philoul:
> For me that is not good enough. I use a 2U Pre-Bolus when I exit my office on the way to lunch. I don't want to interact with UI there, otherwise I could directly use the UI.

I understand your point but maybe a bit too "binary" in conclusion:
- unlock phone and then scan: mandatory in both cases
- validate in the confirmation popup : one click

> Otherwise I could directly use the UI

- open AAPS (automated with the NFC scan ? Or one click)
- open Insuline (one click)
- select the amount (open keyboard or do a few clicks on buttons shortcuts according to the amount)
- validate the screen (one click)
- validate in confirmation popup (one click)
🥵

I understand you prefer 0 click, but you cannot compare directly both solutions and explain they are equivalent: even with confirmation popup we provide a huge simplification concerning the number of clicks and manual actions to be done to deliver the bolus...

> Is there a way to get a final decision?

 @MilosKozak any guidance?

vanelsberg:
> I understand you prefer 0 click,

_On discussing the need for user conformation, how I see this:_

First of all:
**Minimum requirements**:
- User confirmation is required enabled when not all objectives are done.
- Confirmation should adhere current standards applied for AAPS in general. Any exceptions on this should be based on explicit t design decisions.

Apart from the above I can see several (combinations of) options:

1. User confirmation is required only when enabled by user in settings.
2. User confirmation is always required. No option to disable.
3. Only actions "having a low risk", being just "inconvenient" could optionally have confirmation disabled (e.g: Eating soon, Activity, Hypo).
4. User can disable confirmation for individual tags.
5. After executing an action with confirmation disabled, confirmation for the next action is always required within a certain time frame (fi.1 hour, to prevent accidently firing a tag event multiple times).
6. Only have the option to disable confirmation when running developer mode with related semaphore file.
7. .....others I did not think of?

I can only see option 2 (and maybe 3 + 5) as viable. All others would involve unacceptable risks to the end user.
For the initial version of this feature, I think we should go for only option 2, always requiring confirmation.

(Personal note:
I cannot stress on safety enough: To me (and I think am not alone in this) AAPS operations safety is of the highest importance. Foremost to prevent possible health risks for users. In addition, Opensource AID is more and more accepted by healthcare professionals as an alternative to commercial, FAA approved, AID systems. Any accident that could be seen as a result of "unsafe" could be extremely damaging to the trust in Opensource AID in general)

juehv:
Thanks everyone for the discussion and for taking the time to explain the safety concerns and project expectations. I appreciate the feedback, even if we do not agree on the trade-off.

For my personal use case, disabling confirmation is an essential part of the plugin’s value. A version that always requires manual confirmation would not solve the workflow I built this for, so I do not want to continue this path in my free time.

If disabling confirmation is not acceptable for upstream AAPS, I understand that decision. In that case, I will not continue this PR for upstream inclusion. Others are of course welcome to take over the idea or reuse the code for a confirmation-based variant if that better fits the project’s safety requirements.

Thanks again for reviewing the PR and for the constructive input.

Issue #4676, opened 2026-04-03
Unable to recover Medtrum activation process in FILLED state


jbr7rr:
The fix for https://github.com/nightscout/AndroidAPS/issues/4594 blocks the most used case for the retry activation, i.e. user is filling the patch with pump based attached and interrupts the process in AAPS (phone call or whatever reason)

The deeper issue is that there is a failure mode with Medtrum patches were it goes back to FILLED state (basicly it resets itself) We should handle this in all cases, also outside of the activation process. (i.e. throw an alert/alarm when this happens). 

It would be good to block reactivation of a patch session when once it's passed the PRIMING stage, regardless of which state it is in (either NONE, FILLED or whatever) and always require patch deactivation first in those cases

For this I will propose a PR in the short term to rework the activation process and handle this failure mode.

See also:
https://github.com/nightscout/AndroidAPS/issues/4333
While we cannot address that exact issue (pump dies completly and a connection can not be established, it ties a bit into that as well and can maybe be addressed at the same time

jbr7rr:
Todo: dev4 merge/port (working on it)

PR #4694, opened 2026-04-06
Dev3 medtrum handle patch reset

jbr7rr:
https://github.com/nightscout/AndroidAPS/issues/4676

PR to dev3
Will port to dev4 asap after merge, and will double check dev4 activation process (some things are different)

## Changes
- Give error when patch goes from primed state to FILLED or IDLE state (both local and NS)
- Block retry activation on primed patches
- Deactivate patch: Skip connection when in fault state instead force deactivation
- Activation/deactivation: Handle errors with OKDialog returning to overview, instead of toast and staying in activation process

## Manual tests

- patch reset (state set to FILLED or IDLE) during activation, after prime:
    - EXPECTED: Gives error (pop up dialog) and further activation blocked
    - EXPECTED: Retry-activation also not possible (not presented to user at all)
- Patch reset during priming
    - EXPECTED: Gives error and further activation blocked
    - EXPECTED: Retry-activation also not possible (not presented to user at all)
- Reset during normal operation
    - EXPECTED: Gives error and pressing change patch goes straight to deactivation screen
- Start retry activation when after primed but patch reset (when state has not been read back)
    - Note: This can be tested by exiting the activation screen (press cancel), then reset patch and press "Change Patch" in overview
    - Note2: To speed up, you can export the settings when in PRIMED state when doing another test, and import when you do this test
    - EXPECTED: retry-activation should throw error (pop up dialog), and retry is blocked
- Force close app when in activation process pre-prime but patch has not been primed yet (fresh activation)
    - EXPECTED: retry-activation is allowed, and goes to correct screen (fill patch)
- Force close app when in activation process during priming
    - EXPECTED: retry-activation is allowed, and goes to correct screen (priming progress)
- Force close app when in activation process after priming
    - EXPECTED: retry-activation is allowed, and goes to correct screen (filling completed)

**Test instructions:**
To force reset to FILLED/IDLE state: Remove pumpbase, wait for 15 seconds, reattach pumpbase. (confirm 4 beeps when reattaching pumpbase)
Depending on the reservoir level this will move the state back to FILLED or IDLE

Tested OK with 300U by myself
Requesting at least one tester with Nano (200U), my own 200U nano is broken, and I want another person to test it.

### DIFF
--- a/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/MedtrumPlugin.kt
+++ b/pump/medtrum/src/main/kotlin/app/aaps/pump/medtrum/MedtrumPlugin.kt
@@ -56,6 +56,7 @@ import app.aaps.core.validators.preferences.AdaptiveSwitchPreference
 import app.aaps.pump.medtrum.comm.enums.MedtrumPumpState
 import app.aaps.pump.medtrum.comm.enums.ModelType
 import app.aaps.pump.medtrum.keys.MedtrumBooleanKey
+import app.aaps.pump.medtrum.keys.MedtrumBooleanNonKey
 import app.aaps.pump.medtrum.keys.MedtrumDoubleNonKey
 import app.aaps.pump.medtrum.keys.MedtrumIntKey
 import app.aaps.pump.medtrum.keys.MedtrumIntNonKey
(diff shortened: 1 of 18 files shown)


MilosKozak:
tested on Nano.
But merge to dev will be real challenge

PR #4696, opened 2026-04-07
Insulin Management Improvements

Philoul:
- Update UI (active insulin in yellow) after Insulin activation
- Select running insulin by default on load
- ProfileManagement click on running Insulin open InsulinManager on Running ICfg

### DIFF
--- a/app/src/main/kotlin/app/aaps/ComposeMainActivity.kt
+++ b/app/src/main/kotlin/app/aaps/ComposeMainActivity.kt
@@ -724,7 +724,8 @@ class ComposeMainActivity : AppCompatActivity() {
                         withProtection(ProtectionCheck.Protection.BOLUS) {
                             navController.navigate(AppRoute.ProfileActivation.createRoute(index))
                         }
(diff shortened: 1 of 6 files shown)


MilosKozak:
A few observations:

1. **Duplicated logic in `observeProfileChanges()`** — The `EventLocalProfileChanged` and `EPS` observers have identical bodies. Should be extracted to a shared function.

2. **`loadData` lost `targetIndex` after `addNewInsulin()`** — The `targetIndex` parameter was removed, but after `addNewInsulin()`, `loadData(reload = false)` is called. With `reload = false`, it uses `uiState.value.currentCardIndex` — but after adding a new insulin the current card index may not point to the newly added one. Previously it explicitly passed `insulinManager.currentInsulinIndex`. This looks like a regression — the newly added insulin won't be selected.

3. **Minor style**: Extra spaces in `it.copy( activeInsulinLabel = ... )`.

Philoul:
Fixed (oups forgot to push)

PR #4697, opened 2026-04-08
Pump-Concentration

Philoul:
Update Pump Overview with some missing ch (Virtual, Dash, Eros, Dana, Equil, EOPatch, Medtronic)

One question concerning the formating included into translated strings:
- if I replace the format : `%1$.2fU/h` by `%1$s`, I think translated strings (with previous format) should be replaced by the english one until a new translation is done in Crowdin?

If not we have a risk of crash... (I can update my PR with another key to prevent this)


MilosKozak:
you must create new string if param is different otherwise it crash

MilosKozak:
`<string name="reservoir_value">%1$.0f / %2$d U</string>`
this is current definition
you are using it like
`rh.gs(app.aaps.core.ui.R.string.reservoir_value, ch.fromPump(PumpInsulin(pump.dailyTotalUnits)), ch.fromPump(PumpInsulin(pump.maxDailyTotalUnits.toDouble()))),`

and I checked only this one example ...

Philoul:
Here this string is not only used for reservoir level, but also for TDD (with a max TDD).
-  I hesitate to provide the full string for non U100 `xxx U (yyy CU) / XXX U (YYY CU)` or just TDD in IU values

Additional question: how far we standardize information provided in pump overview
I.e. for TBR I found 
- TBR rate only
- TBR rate + remaining time
- starting time + TBR rate + remaining time
- ...

I found customization of information on several other "standard" items like reservoir level (with or without max), TDD (with or without max TDD), extended boluses (rate, delivered amount? timing?...)
- here I tried to be as close as possible to the origin, but we can probably simplify much more the readability of the code in pump ViewModels if we standardize more the presentation 🤔 (whatever the concentration)

MilosKozak:
I'd standartize it everywhere. easier to maintain
and it cannot be too long to fit on screen .... or split to 2 lines ... you must test, what looks good

Philoul:
I started to standardize common pump information (first with lastBolus including ago)
- I included Insight and Combo even if not converted (will fix conflicts with compose once both Insight on combo are merged in dev)

Philoul:
With my latest commit here are the proposed harmonisation pour Reservoir Level and TDD:
- Reservoir level: proposal is to remove max level (Dana, Diaconn and Medtronic pump only provide max reservoir level). "Over xxx U left" is managed in ViewModel for Dash, Eros and Eopatch pumps.
- TDD: proposal is to add a new row for "Max daily units" for the pumps managing this value (Dana and Diaconn only). Both value "TDD / MaxTDD U" on the same row is only ok for U100 but too long for other concentation with "xxx U (yyy CU)"

Issue #4698, opened 2026-04-08
Max IOB limits not informing user


Descent098:
When adding in a max IOB the hardcoded limits specified [here](https://androidaps.readthedocs.io/en/latest/DailyLifeWithAaps/KeyAapsFeatures.html#open-aps-features-overview-of-hard-coded-limits) are not triggering a message when exceeded.

**Reproduction steps**
1. Be in loop mode (for me it SMB)
2. Go to preferences --> OpenAPS SMB --> `Maximum total IOB OpenAPS can't go over [U]`
3. Set this value to one higher than the [table](https://androidaps.readthedocs.io/en/latest/DailyLifeWithAaps/KeyAapsFeatures.html#open-aps-features-overview-of-hard-coded-limits)
4. Value is silently turned down to the limit without warning user (As confirmed by going to the OPENAPS SMB page and looking at `max_iob`)

So for example if you set it to 38 while your patient type is `Insulin Resistant Adult`, it will be turned down to 30 without any notification. Importantly it still **visually shows up** as 38 in the settings, even though it's being turned down to 30.

<img width="720" height="1416" alt="Image" src="https://github.com/user-attachments/assets/7239eb60-71e3-453c-8897-57618759b01f" />

<img width="720" height="1416" alt="Image" src="https://github.com/user-attachments/assets/344a80cc-92e9-413e-b3b3-ee73194f2bfd" />

**Extra info**
Patient type was set to `Insulin Resistant Adult` if that helps

Build version: `fe9789d-2026.04.07`

I also created a PR to link to those limits from the summary #4699 4699 to come shortly

Descent098:
I unfortunately am not familiar enough with the codebase to fix the primary issue. 

But, after a bit of playing around it seems like potentially some good places might be the max value here overriding everything:

- https://github.com/nightscout/AndroidAPS/blob/master/core/keys/src/main/kotlin/app/aaps/core/keys/DoubleKey.kt#L30

or something in here:
- https://github.com/nightscout/AndroidAPS/blob/master/plugins/aps/src/main/kotlin/app/aaps/plugins/aps/openAPSSMB/OpenAPSSMBPlugin.kt#L426

I'm just not familiar enough with either to know what/how to update this to do a warning, or at least visually show the corrected value

MilosKozak:
will be updated in AAPS4

PR #4699, opened 2026-04-08
Added hard limit docs to the max_iob summary

Descent098:
Including the documentation about hard limits set based on patient type to the IOB summary information when modifying the value

PR to go along with #4698


MilosKozak:
will be updated in v4

PR #4709, opened 2026-04-12
Wear: Enable Running mode tile for AAPSClient

olorinmaia:
Running mode tile was silently blocked for AAPSClient. This removes the guard and adapts the available modes: 
image


MilosKozak:
does it copy behavior from phone?

olorinmaia:
> does it copy behavior from phone?

Yes. I sent you a message on Discord regarding behaviour of reconnect pump on AAPSClient phone on AAPS4 vs AAPS 3.x - I can adjust either phone or wear according to what is intentional.

PR #4732, opened 2026-04-18
Concentration string

Philoul:


MilosKozak:
Reviewed the new exam questions. The gating logic (Objective 2 → concentration feature) looks fine. Main concerns are with the three new exam strings.

## Missing context for the reader

All three questions assume the reader knows the history — that before this feature, users worked around the lack of concentration support by creating a "50% profile" and halving all amounts. None of the questions names this as the legacy workaround, so:

- **Q1** (`u200_with_profile`) describes the legacy workaround without saying so — newer users who never did this will wonder why anyone would
- **Q2** (`switch_u200_profile`) describes the migration path from legacy workaround → new feature, but neither is named
- **Q3** (`switch_u100_u200`) describes fresh use of the new feature, but without contrast

Suggestion: add a one-sentence preamble per question naming the scenario, e.g.:
- Q1: *"Before AAPS supported insulin concentrations, users ran U200 by setting a 50% profile to halve all values. In that setup, which answers are true?"*
- Q2: *"You have been using the 50% profile workaround with U200 and want to switch to the new Insulin Concentration feature. Which answers are true?"*
- Q3: *"You currently use U100 and plan to switch to another concentration using the new Insulin Concentration feature. What should you do?"*

## Grammar / wording

- `u200_with_profile_reports`: "becomes wrongs" → "become wrong"
- `u200_with_profile_pens2`: "what the amount selected on the pen to deliver the bolus" — broken grammar, hard to parse. Suggest: *"I have to convert between the AAPS-calculated amount and the amount I select on the pen for every external bolus."*
- `switch_u200_profile_ps`: "twice more aggressive" → "twice as aggressive"
- `switch_u100_u200`: "included into AAPS" → "included in AAPS"
- `switch_u100_u200_change`: "Insulin into the reservoir" → "insulin in the reservoir"

## Ambiguous options

1. `switch_u100_u200_pen` is marked FALSE, but correctness depends on the pen's concentration. A user who switches the pump to U200 but still uses a U100 pen *does* still need conversion. Suggest specifying *"if the pen matches the new concentration"*.
2. `switch_u100_u200_tube` bundles two unrelated claims (tube pumps + external PDM). Consider splitting.
3. In Q1, `reports` and `mental` options test essentially the same concept (values need conversion for the health team). A user may pick one and move on.

## Recommendation

Worth fixing on this PR before merge rather than iterating later — once strings land on `dev`, Crowdin picks them up and every English edit afterward costs translators rework. The grammar pass is small, and adding the preambles would meaningfully improve comprehension for users who never used the legacy workaround.

PR #4733, opened 2026-04-18
Fix: NumberInputRow updates on keystroke and while field is focused

olorinmaia:
- Calculation now updates on every keystroke instead of requiring Enter or focus loss
 - +/- buttons and quick-add buttons now update the text field even when it has focus


MilosKozak:
it's not so simple how it looks like. this would have many unwanted side effects

PR #4735, opened 2026-04-19
Wear: Add Blood Glucose (large) complication

olorinmaia:
New complication showing the BG value as large as possible, for users who find the standard BG complication text too small. Some users have reported that certain languages also aren't able to display the delta correctly due to total text length.
Trades delta for more space, keeping trend arrow and auto-updating age in the title slot.                                                                                                                      
                                                                                                                                                                                                                                                                                                                                             
  - Text: BG value only — maximizes display size                                                                                                                                                                                                                                                                                             
  - Title: trend arrow + auto-updating age ("3m ↗") — no delta
  - Tap opens BG graph (same as standard BG complication)                                                                                                                                                                                                                                                                                    
  - Appears as "Blood Glucose (large)" in the complication picker    
  
Screenshots:
<img width="1045" height="263" alt="image" src="https://github.com/user-attachments/assets/be418f0c-b5a1-41a7-8698-10967cdd2911" />
<img width="450" height="450" alt="image" src="https://github.com/user-attachments/assets/b1c532b2-1c44-4786-87f5-b1eb6f0e3c20" />

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -194,6 +194,24 @@
         </service>
 
         <service
+            android:name=".complications.SgvLargeComplication"
+            android:exported="true"
+            android:icon="@drawable/ic_sgv"
+            android:label="@string/complication_blood_glucose_large"
+            android:permission="com.google.android.wearable.permission.BIND_COMPLICATION_PROVIDER">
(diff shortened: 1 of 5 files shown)


t1dude:
I have tested this PR and it works as expected, with improved readability of the BG value.

PR #4743, opened 2026-04-22
Add BG graph Wear OS tile

olorinmaia:
Adds a new Wear OS tile (`BgGraphTileService`) showing the same BG history                                                                                                                                                                                                                                                                 
  graph as `BgGraphActivity` as a swipe-accessible tile.

  - Full-screen bitmap tile: BG value + trend arrow + delta + age + IOB/COB/BR/target header,
    followed by the 3h BG graph (shared `renderBgGraph` renderer)
  - Tap anywhere opens `BgGraphActivity` for live updates and window cycling
  - Auto-updates on every new BG reading via `ComplicationDataRepository` flow
  - Pre-renders bitmap in `onCreate` so tile display is instant; 5s timeout
    fallback to black placeholder prevents system timeout on slow DataStore reads
  - PNG encoding for lossless color quality
  - Also removes redundant `AAPS(...)` prefix from all tile picker labels —
    all Wear OS platforms show the app name/icon alongside tile names

Screenshots:
<img width="1577" height="317" alt="image" src="https://github.com/user-attachments/assets/772d1022-9625-483a-b085-2db845df616f" />


t1dude:
Tested and works well! I really welcome this feature. I'd look into the tap action on the tile, maybe open AAPS main menu and not the graph itself?

t1dude:
> I'd look into the tap action on the tile, maybe open AAPS main menu and not the graph itself?

Nevermind, I see now that it isn't possible to change time interval in the tile itself. So it makes sense that tapping on the tile opens the graph in AAPS, letting me tap to change time interval. Swiping from there opens the AAPS mail menu. 

All good!

olorinmaia:
> > I'd look into the tap action on the tile, maybe open AAPS main menu and not the graph itself?
> 
> Nevermind, I see now that it isn't possible to change time interval in the tile itself. So it makes sense that tapping on the tile opens the graph in AAPS, letting me tap to change time interval. Swiping from there opens the AAPS mail menu.
> 
> All good!

I could possibly add tile settings for user to change what tap to action does.
Available options could be:
BgGraphActivity (Default)
MainMenuActivity
LoopStatusActivity
... anything else?

For my own part I got complications that take me to the main menu by tapping but this isn't the case for everyone I would assume so letting the user decide what tap to action should be isn't a bad idea for improving this new tile :) 

I'm actually planning migrating tile settings to compose, so this is something I can add in that PR. Tile settings can't be set  or migrated to the phone app.

PR #4759, opened 2026-04-24
CGM: Add Instara CGM support.

leohunt:
Background
Instara CGM Link:
https://play.google.com/store/apps/details?id=com.teljane.instara

Main changes
Add files to support a new kind of CGM: Instara, with the following functions:

Receiving glucose value data from Instara CGM.
History data consistency check, with user setting option(default as enabled).
InstaraTrendArrowUtility to help calculate trend arrow related information.

Development Notes
As suggested, use the pumpId to store sgvId, with notes to avoid confusion.
Save devicePrefix, sgvMark, sgvStart into preference xml. For the same device, only insert once. When the user changes to a new device, this part of information will be updated.
+v4 changes: Refactoring methods in PersistenceLayer to make it more general.

Test
Currently a few users are using this version and it works as expected. We'll continue the test.

### DIFF
--- a/core/keys/src/main/kotlin/app/aaps/core/keys/BooleanKey.kt
+++ b/core/keys/src/main/kotlin/app/aaps/core/keys/BooleanKey.kt
@@ -55,8 +55,6 @@ enum class BooleanKey(
     BgSourceUploadToNs("dexcomg5_nsupload", true, R.string.pref_title_bg_source_upload_to_ns, defaultedBySM = true, hideParentScreenIfHidden = true),
     BgSourceCreateSensorChange("dexcom_lognssensorchange", true, R.string.pref_title_bg_source_create_sensor_change, R.string.pref_summary_bg_source_create_sensor_change, defaultedBySM = true),
 
-    InstaraHistoryRequestEnabled("instara_history_request_setting", true, R.string.pref_title_instara_history_request, R.string.pref_summary_instara_history_request, showInNsClientMode = false),
-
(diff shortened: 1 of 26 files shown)


leohunt:
@MilosKozak Would you please review this PR? Thanks!

core/keys/src/main/kotlin/app/aaps/core/keys/BooleanKey.kt
```diff
@@ -55,6 +55,8 @@ enum class BooleanKey(
     BgSourceUploadToNs("dexcomg5_nsupload", true, R.string.pref_title_bg_source_upload_to_ns, defaultedBySM = true, hideParentScreenIfHidden = true),
     BgSourceCreateSensorChange("dexcom_lognssensorchange", true, R.string.pref_title_bg_source_create_sensor_change, R.string.pref_summary_bg_source_create_sensor_change, defaultedBySM = true),
 
+    InstaraHistoryRequestEnabled("instara_history_request_setting", true, R.string.pref_title_instara_history_request, R.string.pref_summary_instara_history_request, showInNsClientMode = false),
```
MilosKozak: Looks good. Just make these 2 preferences plugin specific. Look at this pattern:
`ownPreferences = listOf(VirtualBooleanNonPreferenceKey::class.java),`

leohunt: Thanks for the suggestion! I've moved Instara specific related setting and meta to two corresponding key files.

PR #4766, opened 2026-04-25
Scenes Category Translation + additional activities

Philoul:
Include additional activities (tennis, football, ...)
Replace hardcoded categories by strings

(will continue with icon labels to allow translations)


Philoul:
Icon label added for translation...

PR #4775, opened 2026-04-29
Add adaptive smoothing plugin as selectable smoothing option.

MTR93600:
Integrate adaptive smoothing into plugin registration, BG quality checks, and smoothing interfaces with dedicated unit tests, while keeping dashboard-specific features out of dev scope.


MilosKozak:
Thanks for the work — the UKF math is solid (sigma-point geometry, λ/γ, predict/update separation all check out). I read the full diff and noticed a few things worth addressing before merge:

**Likely build-breaking**
- `AdaptiveSmoothingPluginTest.kt` uses `whenever { … } doReturn …`. mockito-kotlin 6.3.0 doesn't have a lambda form — it's `whenever(call).thenReturn(value)` or `whenever(call) doReturn value`. `calculateIobFromBolus()` is `suspend`, so it also needs `runBlocking { … }` (see `TizenPluginTest.kt:53` for the canonical pattern in this repo).

**Concurrency / lifecycle**
- `LoadBgDataWorker.smoothData()` calls `smooth()` inside `synchronized(dataLock)`. The new `runBlocking { iobCobCalculator.calculateIobFromBolus() … }` therefore runs suspending DB code while the lock is held. Moving from per-point to once-per-call reduces but doesn't eliminate the ANR risk. Cleaner: compute IOB before `smoothData` and pass it in (would need a `Smoothing` overload or context object).
- `subscribeToSensorChanges()` / `loadLastSensorChange()` create ad-hoc `CoroutineScope(Dispatchers.IO)` from `init { }`. As a `@Singleton`, this fires at app start and can race with plugin/DB initialization. Consider gating on `Config.awaitInitialized()` or binding to a lifecycle scope.
- `shouldResetLearning` returns `true` when `currentTimestamp - lastProcessedTimestamp < 0`. NSClient/follower catch-up routinely delivers older readings — this would spuriously reset learned `R`.

**Dead code / cleanup**
- "BACKWARD PASS (RTS SMOOTHER)" block is a stub — declares `xSmooth` and does nothing. Either implement or remove `forwardStates` + `FilterState`.
- `aapsSchedulers` is injected but never referenced.
- `core:keys` dep is added but unused; the plugin uses raw string keys (`"ukf_learned_r"`, …) with `SP` directly, bypassing the typed-key convention.
- Unused imports: `CompositeDisposable`, `plusAssign`, `TimeUnit`, `kotlin.math.{exp,pow,min}`, deprecated `kotlinx.coroutines.flow.collect`.
- `sensorSessionId++` is never read; `if (gv.trendArrow == null)` in `sanitizeOutput` is unreachable (field is non-null).
- `Pzz < 1e-6` early-return in `update()` leaves `x`/`P` at pre-prediction values; should fall back to `xPred`/`PPred`.

**Interface surface**
- `Smoothing.preferDashboardGlucoseFromGlucoseStatus()` is added but has no consumer in this PR. Since the description says dashboard wiring is intentionally out of scope, consider keeping it out of `core:interfaces` until the consumer lands (or document it as phase-1 forward-compat).

MTR93600:
let me know if it's ok with the last commit

ga-zelle:
When checking for overnight compression low you use "isNight". Its definition of sleep hours seems very rigid and certainly would not work for me. Also it refers to a preference which may only be availble in AIMI.

piecycle:
This code plagiarises my work. It should be removed and replaced with my original.

https://github.com/piecycle/tsunami/blob/devTsuV35/plugins/smoothing/src/main/kotlin/app/aaps/plugins/smoothing/UnscentedKalmanFilterPlugin.kt

piecycle:
The filter is an unscented kalman filter and should also be called that. Calling it "adaptive smoothing plugin" is just an attempt to hide the origin of this code.

PR #4784, opened 2026-05-01
Fix dead "Don't Kill My App?" link for unknown manufacturers

PieterDeBruijn:
The About dialog built its target URL as
https://dontkillmyapp.com/<Build.MANUFACTURER>, which 404s for any manufacturer that does not have a dedicated page on dontkillmyapp.com (e.g. Fairphone, HMD, custom ROMs, emulator).

Link to the homepage with the ?app=AAPS query parameter instead, as recommended by DKMA's own API docs. The homepage lets the user pick their device and always resolves.

### DIFF
--- a/app/src/main/kotlin/app/aaps/MainActivity.kt
+++ b/app/src/main/kotlin/app/aaps/MainActivity.kt
@@ -214,7 +214,7 @@ class MainActivity : DaggerAppCompatActivityWithResult() {
                                 startActivity(
                                     Intent(
                                         Intent.ACTION_VIEW,
-                                        Uri.parse("https://dontkillmyapp.com/" + Build.MANUFACTURER.lowercase().replace(" ", "-"))
+                                        Uri.parse("https://dontkillmyapp.com/?app=AAPS")
                                     )
                                 )
                             }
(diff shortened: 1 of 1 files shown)


MilosKozak:
I'm not sure if it's a win to make it more convenient to 5% of users and less convenient to 95%

MilosKozak:
sorry it's even worse because it offloads maintaining the list to me

Philoul:
Hum, I agree with @MilosKozak, probably more than 95% of the users are with known brands (sorry for you...)
- Do we have an easy way to detect 404 error (before showing the target page to end user), and only show the generic page in this situation ? (if yes, it could be a way to keep direct page for > 95% of current users, and improve for the < 5% remaining ones)

vanelsberg:
Do not like linking to the AAPS documentation as feels as RTFM... :-|

> * Do we have an easy way to detect 404 error 

Like @Philoul suggestion, if doable, technically I'd prefer checking for 404. However, this assumes user has an internet connection!

Other solution would be "whitelisting" <Build.MANUFACTURER>. If it is not in the list set <Build.MANUFACTURER> to "empty string" and link to the [main page](https://dontkillmyapp.com/) for user to decide?. Less coding but would take a bit more maintenance on the whitelist...?

Also Agree with @MilosKozak and I think there are other priorities currently.
**That said:**
Why making this more complex then needed:
Just link to the main site. User is perfectly capable to select his/her phone or go to general!?

PR #4452, opened 2025-12-30
Improve OpenAPS reason display logic for AAPSClient

olorinmaia:
Split out from: https://github.com/nightscout/AndroidAPS/pull/4448

**Problem**
Duplicate "Enacted" and "Suggested" messages shown when they're the same
Stale enacted data in AAPSClient OpenAPS pill.

**Solution**
Show "Loop" label when enacted/suggested match
Hide stale enacted data

**Testing**
✅ TBR only
✅ SMB only
✅ TBR + SMB
✅ No changes loop
✅ Stale enacted data

**Result:** Cleaner, more accurate loop status display.

Some pictures to illustrate. See also: https://github.com/nightscout/AndroidAPS/issues/3706

**BEFORE**
| Duplicated data | Stale enacted data | Stale misleading enacted data |
|--------|--------|--------|
| image | image | image | 

**AFTER PR**
| Loop result | AAPSClient OpenAPS pill |
|--------|--------|
| image | image |
| image | image |
| image | image |

Issue #3706, opened 2025-01-06
[3.3] AAPSClient displays several OpenAps statuses


emilisev:
Related to #3571 and #3620, opening a new issue for clarity.

The bug : AAPSClient displays several data coming from deviceStatus.

![Screenshot_20250106-104712](https://github.com/user-attachments/assets/b327ac9a-7e0a-4984-bf4c-564b05d76a6c)

Content of device status collection : 

![2025-01-06 10_51_20-AndroidAPSdocs-emilisev](https://github.com/user-attachments/assets/1970ffa1-8408-4b92-8670-9d45c4c3c7e8)

So, my guess is that it displays the last entry + last entry which had the openaps.enacted field.

Attached AAPS log, timestamps are `09:41:20` and `09:21:24`
[AndroidAPS.log](https://github.com/user-attachments/files/18317848/AndroidAPS.log)

MilosKozak:
That's perfectly OK.
First is last enacted, second is last sugested

emilisev:
The double display is not ok, though ?!... And never happened before 3.3...

MilosKozak:
It was always so.
Suggested is genereated on every APS run
Enacted only when change is applied to pump 
There can be both records of the same time (if there is was a change requested and and change was enacted by pump)
or different time (if last APS run was not resulting in change in TBR/SMB)

emilisev:
I have checked display with 3.2 and confirm that with my current data, I have the same duplicate display.
Then I checked my logs and deviceStatus collection from 3.2 and I have NO entry with enacted field, explaining why I never saw this before... 
I'm trying to make sense of what you say, but it doesn't, for me. The result on AAPSClient side is that we display both current and outdated data, with no difference whatsover in the display. If the "enacted" entry was displayed differently to show what changed, I could understand, but here it's just the same display and no explanation. Even if the display was adapted for enacted entry, is there really a good reason to show that here ? I mean, all things enacted by the pump are already present in the main graph. What other relevant info is there to show ?

MilosKozak:
https://github.com/nightscout/AndroidAPS/commit/930ac1783ad02155f022590a141a6c8626beb1fe

MilosKozak:
Suggested is important to see that APS is running and what is the result
Enacted what really happened
both is moportant

emilisev:
[...]

PR #4805, opened 2026-05-09
AAPSClient: Improve the NSClient status row (Pump / OpenAPS / Uploader chips) in the overview screen

olorinmaia:
See also closed PR https://github.com/nightscout/AndroidAPS/pull/4452 and issue: https://github.com/nightscout/AndroidAPS/issues/3706

  - Fix row breaking into two lines on translated languages: equal-width chip slots with maxLines=1/ellipsis instead of free wrapping
  - Fix flavor tint not covering card rounded corners (padding moved inside header row so card clip applies correctly)
  - Remove extra space between charging symbol and battery percentage
  - OpenAPS expanded detail: deduplicate enacted/suggested entries.                                                                                                                                                                                                                                                                          
    - The loop always produces a suggested result, and enacts it if the pump needs a new temp basal / smb. Both entries previously showed at the same time even when they represented the same decision. 
    - New logic: if suggested and enacted timestamps are within 60s of each other (same loop cycle), show enacted only - confirming the action was executed. 
    - If suggested is newer than enacted (loop ran but no pump action was needed, e.g. continuing current temp basal), show  suggested only. Never show both simultaneously.
    
Screenshots: **(SEE UPDATED SCREENSHOTS IN COMMENTS BELOW)**
| Before | After |
|--------|--------|
| <img width="397" height="291" alt="image" src="https://github.com/user-attachments/assets/c2d30ed4-aee6-4382-9256-4fffb061a234" />| <img width="391" height="245" alt="image" src="https://github.com/user-attachments/assets/aa3773dc-5626-4ec0-a121-c50adf2911a3" />|
| <img width="389" height="949" alt="image" src="https://github.com/user-attachments/assets/dc07228c-640d-4081-b4cc-0dc932a73536" />| <img width="389" height="949" alt="image" src="https://github.com/user-attachments/assets/9b135e41-e9b1-4db9-a83c-be19bfd59609" />|

### DIFF
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/overview/graph/OverviewDataCache.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/overview/graph/OverviewDataCache.kt
@@ -124,6 +124,7 @@ interface OverviewDataCache {
     // NSClient status (pump/openAPS/uploader from Nightscout) — only for AAPSCLIENT builds
     // =========================================================================
     val nsClientStatusFlow: StateFlow<AapsClientStatusData>
+    fun refreshNsClientStatus()
 
     fun reset()
 }
--- a/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/overview/graph/OverviewDataCache.kt
+++ b/core/interfaces/src/main/kotlin/app/aaps/core/interfaces/overview/graph/OverviewDataCache.kt
@@ -124,7 +124,6 @@ interface OverviewDataCache {
     // NSClient status (pump/openAPS/uploader from Nightscout) — only for AAPSCLIENT builds
     // =========================================================================
     val nsClientStatusFlow: StateFlow<AapsClientStatusData>
-    fun refreshNsClientStatus()
 
     fun reset()
 }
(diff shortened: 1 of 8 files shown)


MilosKozak:
why is onExpand called refreshNsClientStatus?
data should be held in viemodel

olorinmaia:
> why is onExpand called refreshNsClientStatus? data should be held in viemodel

Reverted this change due to your fix in ef476a98

MilosKozak:
I'm not convinced it's a good solution. On small displays nothing will be visible

olorinmaia:
> I'm not convinced it's a good solution. On small displays nothing will be visible

I got a "small" in width phone z flip that I use, and the most important info is shown, like "Pump: 2m...", cutting text of "ago" and this type of text is not a problem imo. I don't have any smaller phone to test on. If you expand the row you will see it all. But I think on even smaller phones, you should be able to see the "x min", "x sec" , it's the "ago" that hopefully will be clipped.

olorinmaia:
@MilosKozak 
If I zoom to max on my z flip phone it looks like this, imo no problem:
| Before | After |
|--------|--------|
| image | image |

MilosKozak:
run it on small phone. you'll see only labels, no data

olorinmaia:
Improved. One row if possible, if not splits evenly and centers into two rows:
English:
image


Norwegian:
image

PR #4806, opened 2026-05-09
AAPS CI automatic cleanup (for now: Branch CI only)

vanelsberg:
**For now this PR is used from branch-ci only. Cleanup logic can be easily linked from other workflows like AAPS-CI and PR-ci** 

This PR cleans up the number of CI runs when it exceeds the maximum runs set. Implementation uses a threshold value to prevent running the cleanup process on every build.

- Added cleanup-workflow-runs.yml 
- In branch-ci it is linked at the end of the workflow with parameters _keep_runs=30, keep_threshold=10_. This way the total number of branch_ci in the list is shrinked to 30 once it exceeds 40 runs.

### DIFF
--- a/.github/workflows/branch-ci.yml
+++ b/.github/workflows/branch-ci.yml
@@ -294,3 +294,18 @@ jobs:
           upload_to_gdrive "aaps-wear-${VERSION}${VERSION_SUFFIX}.apk" "aaps-wear-${VERSION}${VERSION_SUFFIX}.apk"
 
           echo "🎉 APKs successfully uploaded to Google Drive!"
+
+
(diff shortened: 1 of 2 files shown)


MilosKozak:
this will probably work only after merged to master.
you should target dev3

vanelsberg:
> this will probably work only after merged to master. you should target dev3

Thought we should test this on dev for dev's first? But you are right as it is not related to V4.

I do not expect any problems. Will target this PR for dev3 tomorrow (and I guess this wil then also get to current dev eventually...).

vanelsberg:
Replaced by PR targeting dev3

Issue #4542, opened 2026-02-12
AAPS continuously attempts to connect when disconnected from Dash Pods


belevine:
AAPS 3.4.0.0

After disconnecting from a Dash pod, AAPS will continuously state that it is trying to connect despite new Pod activation not being started.  This is usually not noticed when immediately starting a new pod.  While trying to connect, AAPS drains battery at a high rate.

This was observed on an AutoISF variant, which should not have any changes to Dash code.

Attached log was during a period of not having a pod active.  

[AndroidAPS._2026-02-09_09.log](https://github.com/user-attachments/files/25248249/AndroidAPS._2026-02-09_09.log)

xitation:
I see this same behaviour too on the normal AAPS 3.4.0.0 Master.

MilosKozak:
try with this fox: https://github.com/nightscout/AndroidAPS/commit/9d3bdce2a573c8bf49e693d6a045d72b95a9598d

xitation:
I'll cherry pick this and give it a test, let you know how it goes.

belevine:
> I'll cherry pick this and give it a test, let you know how it goes.

Curious if it works for you.  It did not for me, still have the same issue.

PR #4852, opened 2026-05-24
Fix: Omnipod Dash infinite connecting loop after pod deactivation

### LINKED ISSUE #4848: AAPS continuously attempts to connect to Omnipod after deactivating
Reporting this issue still exists: https://github.com/nightscout/AndroidAPS/issues/4542

A14, AAPS 3.4.2.2.  AAPS continuously tries to connect to a pod after deactivating an active pod. 

Image

nl-ruud:
## Fix: Omnipod continuously shows "Connecting" after pod deactivation

Fixes #4848, #4542

### Problem

After pod deactivation, AAPS gets stuck in an infinite "Connecting for X s" loop and never returns to a normal idle state. Battery drain increases significantly during this period. Commit 9d3bdce (since 3.4.1.0) made `isConnected()` return `true` when no pod is present to suppress spurious BLE connect attempts, but inadvertently caused `QueueWorker` to spin forever: `isBusy()` returns `true` at `NOT_STARTED`, so the queue-empty exit is never reached and a WorkManager wake lock is held indefinitely.

### Fix

Exclude `ActivationProgress.NOT_STARTED` from the busy range in `OmnipodDashPumpPlugin.isBusy()` — consistent with how the Medtrum and Eopatch drivers report this state. APS commands and activation are unaffected (guarded by `isInitialized()` and a separate code path respectively).

### Testing

Verified on a physical device — no more infinite connecting loop after deactivation.

### Target

Bugfix for `dev3` (AAPS 3.x). Applies as a clean cherry-pick to `dev` (AAPS 4.x) as well.

### DIFF
--- a/pump/omnipod/dash/src/main/kotlin/app/aaps/pump/omnipod/dash/OmnipodDashPumpPlugin.kt
+++ b/pump/omnipod/dash/src/main/kotlin/app/aaps/pump/omnipod/dash/OmnipodDashPumpPlugin.kt
@@ -237,7 +237,8 @@ class OmnipodDashPumpPlugin @Inject constructor(
 
     override fun isBusy(): Boolean {
         // prevents the queue from executing commands
-        return podStateManager.activationProgress.isBefore(ActivationProgress.COMPLETED)
+        return podStateManager.activationProgress != ActivationProgress.NOT_STARTED &&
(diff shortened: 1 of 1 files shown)


belevine:
Fix worked for me on A14, AAPS 3.4.2.2.  Thank you

PR #4863, opened 2026-05-28
Add xDrip to notification parser plugin

cascer1:
This PR adds support for xDrip notification parsing.

xDrip notifications contain a variant selector character after the trend arrow. So, this PR introduces those to the `SYMBOL_RANGES` list in the notification parser.


Philoul:
Hum

This is more a workaround solution because of recent issues with xDrip sync, but in my mind it cannot be a target solution included within AAPS...
And additionally xDrip is not a sensor brand, so for traceability we don't know which sensor is behind xdrip...

MilosKozak:
why? we have xdrip xDrip source plugin

PR #4873, opened 2026-06-01
Notify when an invalid profile blocks Nightscout profile sync

### LINKED ISSUE #4872: A single invalid local profile silently stops all profiles from syncing to Nightscout
### Summary

If one local profile is invalid — for example a newly added profile that hasn't been fully filled in (shown in red in the Profile editor) — AndroidAPS silently stops uploading the **entire** profile store to Nightscout, not just the invalid profile. Every other, fully-valid profile also stops syncing. There is no notification, no NSClient log entry, and nothing in the UI to indicate sync has stopped, so it can go unnoticed for a very long time — Nightscout just keeps showing the last profile it received.

### Why this matters (real-world impact)

Because the freeze is silent, Nightscout can keep showing an old profile for a long time while the values actually running in AAPS have changed. In our case this had a direct clinical-safety consequence: my child's endo team was reviewing the profile in Nightscout and making percentage adjustments to basal, ISF and carb ratio — working from the stale Nightscout values, not realizing those no longer matched the settings AAPS was actually using. When the displayed profile and the live profile silently diverge, dose adjustments get made against the wrong baseline. A simple notification that sync is blocked would have surfaced the problem immediately.

### Steps to reproduce

1. Have one or more valid local profiles syncing to Nightscout normally.
2. In the Local Profile editor, add a new profile and leave it incomplete (some values at 0 / shown in red). It can't be saved, but it stays in the profile list, and you can still switch to and use the existing valid profiles.
3. Make any change that would normally trigger a profile upload.

### Actual behaviour

- No profiles are uploaded to Nightscout — including the valid ones that were syncing before.
- There is no warning, notification, or log message. The profile-switch dialog even hides the invalid profile from its list, so nothing visibly points to the cause.
- The condition persists indefinitely; sync does not recover until the invalid profile is fixed or removed.

### Expected behaviour

At a minimum, the user should be clearly told that profile sync to Nightscout is blocked and why, so they can fix it. (Ideally the valid profiles would still upload, but even just surfacing the problem would prevent a silent, long-term sync outage.)

### Where it happens

In `DataSyncSelectorV3.processChangedProfileStore()` the upload is gated by:

```kotlin
if (!profileStore.allProfilesValid) return
```

`allProfilesValid` is true only when *every* profile in the store is valid, and the store is uploaded as a single document — so one invalid profile blocks the whole upload. The early return is silent and the last-synced marker is never advanced, so the check fails again on every sync cycle. The same gate exists in `DataSyncSelectorV1` and the xDrip data-sync path.

### Affected versions

Reproducible on 3.4.x and present on current `dev`.

### Ideas for fixing it

A few options, depending on how far you want to take it:

- **Quick win (happy to PR this now):** just make it loud. Pop a notification and an NSClient-log line when this gate stops the upload — something like *"A profile is invalid, so none of your profiles are being uploaded to Nightscout. Open the Profile editor and fix the one shown in red."* — and clear it automatically once everything's valid again. It doesn't change what actually gets uploaded, so it's low-risk; it just means you find out instead of it failing quietly.

- **A bit further:** don't let one bad profile drag down the good ones. Upload the valid profiles and skip only the broken one, and call out which profile is the problem. (Someone would need to sort out how that plays with `defaultProfile` and what Nightscout ends up showing.)

- **Bigger picture:** catch the bad profile earlier so it never just sits there silently blocking sync — e.g. check profiles on import/restore (probably the most common way one sneaks in), don't let a half-finished profile live in the store, and have the profile-switch dialog actually show an invalid profile instead of quietly hiding it.

ColebyPearson:
## What this does

Surfaces a notification when an invalid local profile silently blocks the Nightscout profile-store upload.

When any local profile is invalid, `allProfilesValid` is `false` and `DataSyncSelectorV3.processChangedProfileStore()` skips uploading the **whole** profile store (it's uploaded as a single document). Today that early `return` is completely silent — no log, no notification, and `ProfileStoreLastSyncedId` is never advanced — so none of the profiles, including the valid ones, reach Nightscout and the user has no way to know. It can persist indefinitely.

This PR adds observability on that already-failing path:

- Posts a `NORMAL` notification (`PROFILE_NOT_SYNCED_INVALID`): *"A profile is invalid, so none of your profiles are being uploaded to Nightscout. Open the Profile editor and fix the one shown in red."*
- Writes an `NSCLIENT` debug log line.
- Dismisses the notification automatically once every profile is valid again.

It does **not** change what is uploaded or when — purely a "make the silent failure visible" change, so there's no loop/insulin impact.

Scope: this covers the default NSClientV3 path. `DataSyncSelectorV1` and the xDrip data-sync path have the identical gate and can be a small follow-up.

## Why

Closes #4872 — the issue describes the real-world impact (a stale profile shown in Nightscout while the live AAPS settings had changed, leading to therapy adjustments against the wrong baseline).

## Test plan

New unit tests in `DataSyncSelectorV3Test`:
- invalid store → posts the notification and does **not** upload.
- valid store → dismisses the notification.

`./gradlew :plugins:sync:testFullDebugUnitTest --tests "*DataSyncSelectorV3Test"` passes locally.

## Files touched

- `core/interfaces/.../notifications/NotificationId.kt` — new `PROFILE_NOT_SYNCED_INVALID` id.
- `plugins/sync/.../res/values/strings.xml` — new string.
- `plugins/sync/.../nsclientV3/DataSyncSelectorV3.kt` — post/dismiss at the gate.
- `plugins/sync/.../test/.../nsclientV3/DataSyncSelectorV3Test.kt` — tests.


ColebyPearson:
Self-review / notes for reviewers:

- **Risk: minimal.** The change only runs on the existing `allProfilesValid == false` early-return branch (plus a dismiss on the valid fall-through). It doesn't change what is uploaded, when, or any loop/insulin behaviour — purely a notification + log + auto-dismiss.
- **No spam:** the notification id de-dupes in `NotificationStore`, so re-evaluating every sync cycle yields a single card; `dismiss` is a no-op when the id is absent, so calling it on the valid path each cycle is harmless.
- **Severity:** chose `NORMAL` to match the existing profile/sync notifications (`INVALID_PROFILE_NOT_ACCEPTED`, `OLD_NS`). Happy to switch to `URGENT` given the silent-data-loss nature if you'd prefer.
- **Verified:** two unit tests added (invalid → posts notification + does not upload; valid → dismisses); `:plugins:sync:testFullDebugUnitTest --tests "*DataSyncSelectorV3Test"` passes locally.
- **Follow-ups (intentionally not in this PR):** the same gate exists in `DataSyncSelectorV1` and the xDrip data-sync path; and naming the specific offending profile would need cross-profile validity plumbing.

Not an approval — author self-review for the audit trail.

MilosKozak:
This was replaced by https://github.com/nightscout/AndroidAPS/commit/9ccd1e0ea32410fd0e64f3b1be2aa335d6ec1ac9
Hopefully it will solve the issue, not ex-post warning

PR #4894, opened 2026-06-10
BolusWizard source for uel

Philoul:
Improve source management within BolusWizard


MilosKozak:
this is right now changing in ns_settings

Philoul:
Ok, will check 😉

PR #4916, opened 2026-06-20
Fix glucose unit display labels to use correct casing (mmol/L, mg/dL)

olorinmaia:
- Add `GlucoseUnit.displayLabel` property returning "mmol/L" and "mg/dL" — the correct notation for display
  - `GlucoseUnit.asText` ("mmol" / "mg/dl") is unchanged — it is Nightscout's wire format
  - Replace all UI display usages of `.asText` with `.displayLabel` in dialogs and profile screens

**Affected screens**
  - Temp target editor input field
  - Bolus wizard dialog
  - Calibration dialog
  - Care dialog
  - Profile editor (ISF and target unit labels)
  - Profile helper / compare screen
  - Profile viewer
  - Autotune screen and warnings

**Example before/after**
| Before | After |
|--------|--------|
| image | image |


MilosKozak:
No to this.  Code like this:
```
"${units.displayLabel} ${rh.gs(R.string.profile_per_unit)}"
```
must be eliminated, not improved, because it's not possible to translate it correctly in RTL langs

olorinmaia:
> No to this. Code like this:
> 
> ```
> "${units.displayLabel} ${rh.gs(R.string.profile_per_unit)}"
> ```
> 
> must be eliminated, not improved, because it's not possible to translate it correctly in RTL langs

Thanks for feedback. I agree and have made adjustments. Please review again.

PR #4920, opened 2026-06-22
Wear: Scene tile improvements, AAPSClient relay fix, and duration formatting

olorinmaia:
**AAPSClient wear fix (pre-existing bug)**

  - Scene activation and End from watch were broken on AAPSClient: both called `SceneAutomationApi` directly with no master relay and no `RemoteDelivered`, causing the watch to time out

**Watch Scene tile UX**
  - Scene icon tinted purple (`ic_scene_purple`) matching phone Scenes element color
  - "End" button replaces "STOP" (not caps), red icon (`ic_cancel_red`), routes through
    `AcceptActivity` swipe confirmation via `ActionSceneStopPreCheck` — no accidental
    one-tap scene end
  - Tile picker preview image updated

**Confirmation dialog improvements**
  - Scene and Profile Switch confirmation duration now uses `formatMinutesAsDuration`
    ("1 h 30 min" instead of "90 min"), aligned with TT and running mode
  - Profile Switch: duration line hidden when 0 (indefinite switch)
  - Fix issue with double space in "Profile:  150%" when no profile is selected in scene.

Tested scene flow on AAPS and AAPSClient phone/wear side OK.

Some pictures before/after:

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image |


MilosKozak:
Scenes don't use BatchExecutor but have own path

olorinmaia:
PR Updated and re-tested OK.

PR #4922, opened 2026-06-22
Fix scene TT preset dropdown showing English instead of localized names

olorinmaia:
+ add comment for "back"-string to help translating correctly in Crowdin.
 
 
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/core/ui/src/main/res/values/strings.xml
+++ b/core/ui/src/main/res/values/strings.xml
@@ -253,7 +253,7 @@
     <string name="configbuilder_nightscoutversion_label">Nightscout version:</string>
     <string name="not_available_full">Not available</string>
     <string name="engineering_mode_enabled">Engineering mode enabled</string>
-    <string name="back">Back</string>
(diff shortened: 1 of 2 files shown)


core/ui/src/main/res/values/strings.xml
```diff
@@ -253,7 +253,7 @@
     <string name="configbuilder_nightscoutversion_label">Nightscout version:</string>
     <string name="not_available_full">Not available</string>
     <string name="engineering_mode_enabled">Engineering mode enabled</string>
-    <string name="back">Back</string>
+    <string name="back" comment="Label for a navigation back button or arrow — go to the previous screen. Not the back of a body or object.">Back</string>
```
MilosKozak: comment not necessary here.

olorinmaia: I added comment to crowdin. I notice a few languages have translated wrong.

PR #4929, opened 2026-06-25
Wear: bolus wizard correction — adjust calculated dose directly on the watch

olorinmaia:
Related to a personal request from the caregivers of my son and #4462 - tagging @Philoul due to our discussion on Discord :)

**28/6-2026; UPDATED PICTURES AFTER COMMIT** de18e354

image

#### Demo of new functionality
https://github.com/user-attachments/assets/5298f577-c6f9-456d-8daf-6e115c262e7a

#### What this does

The bolus wizard on the phone has a direct correction field where you can fine-tune the calculated dose before confirming. This PR brings the same capability to the watch.

After the wizard result appears on the watch, **− and + buttons** let you adjust the dose up or down in pump-step increments — without going back to recalculate with a different percentage or carb count.

The adjusted amount is shown live on the result page and carried through to the confirm page, so what you see is exactly what gets delivered. Bolus calculator log in Treatments also reflects the corrections done by watch.

#### Safety

  - **Max bolus constraint enforced**: the + button is disabled once the adjusted dose reaches the pump's max bolus limit.
  - **Cannot go below zero**: the − button is disabled at 0 U. When IOB already exceeds the calculated dose (net negative), the full IOB offset must be recovered before the dose turns positive — matching phone wizard behaviour exactly.
  - **`applyBolusConstraints()` re-runs at delivery time** on the master as a final safety net, regardless of what the watch sent.
  - Works identically for **Master-paired watches** and **Client-paired watches** (correction travels through the NSClient wire protocol).

 #### Testing
  Tested bolus wizard / quickwizard wizard preset on both Master and Client wear with pump Omnipod Dash — all scenarios green: direct correction, negative correction, IOB > calculated dose, max bolus boundary, AAPSClient relay, and carbs-only confirm.
  
Screenshots:
  
image
image
image


Philoul:
Didn't review the code, but on UI point of view implementation is nice 👍.

The only (little) comment (but not sure if there is enough space to improve or not) is the correction amount 🤔
- when we start to adjust with + or - button, we don't see "live" the current correction, only if it's positive (green background behind "+") or negative (red background behind "-")
- but we need to expand the detailed calculation to see the Corr amount...

Do you think it could be possible to add a line with Corr amount between total bolus and carbs amount when Corr is not 0 ?

olorinmaia:
> Didn't review the code, but on UI point of view implementation is nice 👍.
> 
> The only (little) comment (but not sure if there is enough space to improve or not) is the correction amount 🤔
> 
> * when we start to adjust with + or - button, we don't see "live" the current correction, only if it's positive (green background behind "+") or negative (red background behind "-")
> * but we need to expand the detailed calculation to see the Corr amount...
> 
> Do you think it could be possible to add a line with Corr amount between total bolus and carbs amount when Corr is not 0 ?

Thanks for feedback! I actually tried adding a new row for correction amount but it's not enough space, especially for smaller watches. There is more rows added if eCarbs are present (QuickWizard). So I went for the minimalistic solution by adding colors to the + and - buttons. 

Right now I see these possible alternatives. 

We could consider removing the coloring of +/- buttons and instead show "(Corr. +0.05)" if next to "Calculation" or "Corr. +0.05" with smaller font above "Total insulin" (if space). Here we could color the correction value green or red to visualize better or keep it plain and keep coloring of +/- buttons. 

What do you prefer? For me the current solution is sufficient, but since the phone shows you how much you are adjusting without having to look at calculations, I understand the need to show this on watch also. :)

image

Philoul:
I like the colored +/- (so I think we should keep them)

Above the total amount it's a bit strange, so I prefer below
- don't you think it could be in the same row than the Carbs amount, just below the total ?

I think we can only put the value into bracket (it will appear and change on each click on + or - button, so no need to have the "Corr." label).
`20 g carbs (+0.20 U)` or `20 g carbs (-1.15 U)`
- Carbs color is in orange, the correction value can be with in the light "insulin blue" (the same way we have the corr value in the top bar of wizard dialog into the phone within the summary of the calculation), and you keep in addition the colored red or green button.
- let's see if it's ok with a screenshot...

olorinmaia:
I like your proposal. An issue is that wizard result now support Quick Wizard wizard preset and the Carbs line can be pretty long.

Screenshot_20260627_230436_androidaps

Any other place we can add it? What about 
`Total Insulin (+0.20 U)` or `Total Insulin (-1.15 U)`

olorinmaia:
Problem with adding it after "Total insulin" is that it takes to much place, especially with translations to other languages. It gave me an idea to replace "Total insulin" with only -0.05 or +0.05 using the same color as the + and - buttons, this shows clearly how much the user is correcting the total insulin amount in each direction. When swiping to confirm the total insulin also once again shows clearly for user what will be dosed.

Philoul:
An easy way to avoid translation problem is probably to limit information to translate...
Everywhere in AAPS, Carbs is Orange, and Insulin is light blue 🤔

So I think we can remove "Carbs" for immediate amount.

I would keep Correction amount in light blue (I don't think we have somewhere else in AAPS a colored insulin amount in red/green)

So I would keep for Carb row:
`30 g (+1.15 U)` (with or without brackets)
Orange and light blue

Proposal for eCarbs row, replace 
`eCarb: 30g / 2h (+60min)`
by
`10:21: 30 g / 2h`
(With time and duration in white, and only `30 g` in orange)

- No more variation according to translation so 2 row max
- Less "Text" so cleaner confirmation popup

🤔

olorinmaia:
PR updated with several improvements in de18e35 based on feedback from @Philoul ! Thanks!

See PR description for updated screenshots

Philoul:
Latest version really nice 😉
👍

PR #4954, opened 2026-07-05
Wear: let users pick which scenes show on the Scenes tile

olorinmaia:
The Scenes tile only showed the first 4 scenes in phone order, with no way to choose which ones once more than 4 are saved. 

This adds a tile settings screen with 4 slots, each Automatic (fills with the next unused scene), None (deliberately empty), or a specific scene — so a deleted/disabled scene's slot shows "Unavailable" instead of a raw id, without losing the selection if it comes back.

Tested OK on AAPS and AAPSClient wear.

Screenshots:
image

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -609,6 +609,9 @@
             </intent-filter>
 
             <meta-data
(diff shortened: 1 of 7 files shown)


vanelsberg:
Nice! 👍

PR #4968, opened 2026-07-11
new plugin "NFC Commands"

Philoul:
This plugin is an updated version of @juehv's proposal included within PR #4655
I opened this new PR, at least to keep this topic "for discussion" (I personnaly found this proposal as a very nice feature, especially with the possibility to reuse my hundred of old FSL sensors...)

Reminder of the feature:
- User can scan a NFC tag to automate AAPS Command launch (it simplifies a lot user interface for commands done regularly like Pump Disconnection / Connection for showers, breakfast bolus, sport training...)
- It works for most NFC Tags: low cost read/write tags you can buy on internet, but also works with FreeStyle old sensors (read only) that can be reused
- It requires to have the phone unlocked to allow NFC scan

All the NFC communication has been developed by @juehv (unchanged)
- Commands structure updated (the previous "String commands" replaced by a json string)
- Architecture aligned on SMS or Automation architecture with all commands included within a dedicated "Action Classes" within an `actions`sub-folder
- User Entry Logger added
- UI reworked, now very close to Automation UI with command icons, parameters, command selection, confirmation Popup... (I know @juehl the confirmation popup was a blocking point for you... but it's mandatory for a potential merge)
- Commands and Tag's name can be edited / modified (without the need to rescan the corresponding tag)
- All hardcoded text has been replaced by translated strings (with as much as I could I reused existing strings)
- Parameters are now generic (Insulin, Carbs, duration, percent, ...), so if we want to add new actions, most parameters already exists and can be reused without coding
- Plugin moved from `general` to `sync` module
- NfcCommandsPlugin, NfcBuildScreen, etc. are now 100% generic (Category and Action list are manaaged within `NfcCommandCode` File and 100% of the action's code is centralized within each Action Class)
- A few new commands has been added (BolusWizard, Run Scene, Temp Target Manual) and a few others hidden (but code is still here) in System Category (`AapsClientRestartAction`and `RestartAction`). Probably several other actions could be hidden or removed (especially within LOOP Category)
- NFC_COMMANDS.md file and tests has been updated
- All the work has been rebased on latest dev (yesterday), with all @juehv commits into the history (I simplified my work within a few commits only to have a clean history)

Note, I cherry-picked within this branch the commit included within PR #4966, it was needed for BolusWizard source management in uel.log (NFC instead of Wizard Dialog)

@juehv, @vanelsberg, @xitation and others, feel free to test and comment


vanelsberg:
Nice! Merges with current def. Works as expected.👍
As reported, one issue to fix concerning Bolus Wizard (mg/dl <-> mmol/L?)

Philoul:
@vanelsberg BolusWizard issue (mmol units) fixed with https://github.com/nightscout/AndroidAPS/pull/4968/commits/d73622516fd344f2861c499833489cc4877f7bcd

vanelsberg:
> BolusWizard issue (mmol units) fixed with [d736225]

Confirming: fixed! 😃

vanelsberg:
PR merges with current dev V4 without problems👍
Currently testing. Have not found any real issues works as expected. Mostly looking on how to use this new feature.

je-l:
Just tried this, seems to work as expected :+1: 

I didn't realize old sensors could be reused like this. Another tag I tested was a large credit-card sized tag inside a book from library, works just like Libre 3+.

Philoul:
Tags used during my tests

vanelsberg:
A personal example on how  I use this:

_Tag1_: I always start the day with a sort of standard breakfast (fixed amount of carbs, meal bolus, TT). This to prevent my BG spiking as in the morning I am very sensitive for carbs. Enabling SMS for a full day does not work for me because of extreme changes in insulin sensitivity during the day (And before you ask: yes, my profile is perfectly tuned).

Instead of tapping several command into AAPS I scan my "breakfast" tag:

- Enter 10gr of carbs,
- Enable SMB for 90 minutes,
- Execute a meal bolus of 2U,
- Set temporary profile of 110%,

I just scan the tag, make coffee and have my breakfast. Love it 

_Tag2_: When taking a snack (also always about the same amount of carbs):

- Enter 6 gr of carbs,
- Start Bolus Wizard (compensating for current carbs, IOB and BG when needed),

_Tag3_: When going for a walk, go shopping or do a small physical task I have another NFC tag:

-Set Temp target,
- Set temporary profile of 90% for 90 mins,

Still experimenting on other meaningful usage, for instance tasks more effectively replacing "eating soon"...

I think, maybe in future versions (?), we could look into option to scan a tag using the watch (my GW6 can scan NFC). This way there would not even be a need to take out the phone...

vanelsberg:
Requirement is to unlock the phone's screen and confirm the scan.
This is a necessity for security reasons. I would not want to start a NFC scan by accident - without knowing 🙈

Thats why I think in future it is worth looking into Wear integration: scanning the tag from watch could be just as safe as scanning a pay terminal?

vanelsberg:
How to not mixing up different tags?

I have NFC stickers. Location (office, coffee machine, bedside, door, car, etc) of the sticker should me being able to associate its use. I also have a key-tag always with me. Easy to remember it's use. Having some other tags that are looking very different can help remembering...

vanelsberg:
Hint: If you are looking to by NFC tags: Search [Amazon](https://www.amazon.com/s?k=NFC+tags) for instance. Search for "NFC tags".

juehv:
Happy to see, that integration goes on :)

Philoul:
Still not merged, but I keep the working branch up-to-date to be ready: you fully convinced me of the added value.
- around 30% of AAPS users loop with an FSL sensor that use nfc communication only once at the start to initialize Bluetooth communication.
So in addition to the small number of "Geek" users who will buy rewritable tags online, I think the potential user base could be very large.

Personally I set up a tag (old FSL 1 sensor from 5 or 6 years ago) on my chocolate powder container for breakfast; since then, not only is using Bolus Wizard much faster thanks to scanning the tag, but more important, I no longer forget to do it, because I have that visual reminder.

I put another one in my table tennis racket case for trainings
👍

xitation:
The attaching tags to common consumed items in the fridge/pantry idea is genius. I'd use the shit out of that ;) Also a tag near bedside table for "eating soon" mode when you wake up would be handy I rekon rather than trying to automate that, seeing as I wake up at different times on different days based off my work schedule or weekend. I feel there are heaps of really interesting community ideas that could come from this feature, that would help reduce cognitive load of remembering to do things at certain times with a simple phone tap instead.

PR #4978, opened 2026-07-14
Fix deliverTreatment in PumpWithConcentrationImpl (very small boluses with U200 Insulins or more)

Philoul:
Fix proposed by @MTR93600 to handle very small boluses or SMBs with insulin U200
If amount is close to pump minimal bolus, convertion to CU divided by 2 with U200 so floor can lead to 0.0U sent to the pump
- According to pump driver it can raise an IllegalArgumentException


Philoul:
@MTR93600 I cherry-picked your commit so you should be the recorded as the author of the modification...

PR #4984, opened 2026-07-16
E2E: seam to drive pump emulators from instrumented tests

MilosKozak:
The in-tree pump emulators are already wired into the full flavor: DanaModules.provideBleTransport hands out EmulatorBleTransport instead of BleTransportImpl purely on config.isEnabled(EMULATE_*). Instrumented tests could never reach it, though -- ConfigImpl.isEnabled resolves a marker file through a SAF tree URI, and the tests never grant an AAPS directory, so every EMULATE_* flag reads false.

Rebind Config to a decorator that additionally reports whatever a test asks for, leaving everything else (including the init flow the UI observes) delegating to the real ConfigImpl.

Replaces AppModule as well as AppModule.AppBindings: AppBindings is pulled in by AppModule's includes, so replacing it alone would leave the original Config binding in the graph and collide.

First step only -- asserts the wiring (transport is the emulator, unrequested options still read false). Driving a pump through the UI comes next.


MilosKozak:
test

PR #4987, opened 2026-07-19
fix: derive EPS remaining-time display from timestamp + duration

### LINKED ISSUE #4985: ProfileFunctionImpl reads EPS.originalEnd directly for remaining-time display, producing garbage duration (e.g. "32399999998h")
# Widget / overview "remaining time" shows absurd duration (e.g. "32399999998h") for a running temporary ProfileSwitch

## Where

`implementation/src/main/kotlin/app/aaps/implementation/profile/ProfileFunctionImpl.kt`,
`getProfileName()`:

```kotlin
if (showRemainingTime && profileSwitch.originalDuration != 0L) {
    profileName += dateUtil.untilString(profileSwitch.originalEnd, rh)
}
```

This is used by both the home-screen widget (`Widget.kt` → `updateProfile()`) and the Overview
ribbon (`OverviewFragment.kt`) via `ProfileFunction.getProfileNameWithRemainingTime()`.

## What happens

`dateUtil.untilString(profileSwitch.originalEnd, rh)` treats `originalEnd` as an absolute epoch
timestamp and subtracts `now()` from it. When the stored `originalEnd` value on a given EPS record
doesn't actually match `timestamp + originalDuration`, the resulting delta is nonsensical and gets
rendered as something like `(32399999998ч 11')` instead of a sane duration.

## Why this happens

`EPS.originalEnd` (`core/data/src/main/kotlin/app/aaps/core/data/model/EPS.kt`) and
`EffectiveProfileSwitch.originalEnd` (`database/impl/.../entities/EffectiveProfileSwitch.kt`) are
both explicitly documented as:

```kotlin
var originalEnd: Long, // not used (calculated from duration)
```

Every other consumer of this field in the codebase already treats it as unreliable and derives the
real end time from `timestamp + originalDuration` instead:

- `ProfileSwitchExpiryScheduler.kt` (doc comment): *"the EffectiveProfileSwitch query selects the
  latest EPS by timestamp and ignores `originalEnd`"* — it schedules the revert-to-base event off
  `timestamp + duration`, not off `originalEnd`.
- `SceneExecutor.kt` (comment): *"...and ignores originalEnd, so the base profile only resumes once
  a NEW base-profile EPS exists."*

`ProfileFunctionImpl.getProfileName()` is the one place left that still reads `originalEnd`
directly for display purposes, contradicting the documented contract of the field.

## How the stale value gets there

`originalEnd` is only guaranteed correct at the moment of local creation
(`CommandQueueImplementation.kt`: `originalEnd = it.end` where `PS.end = timestamp + duration`).
On an **AAPSCLIENT / follower** setup, the EPS record is never created locally — it's synced verbatim
from Nightscout (`EffectiveProfileSwitchExtension.kt` / `NsIncomingDataProcessor.kt`), so whatever
`originalEnd` value was serialized to NS at creation time is trusted as-is, with no re-derivation
from `timestamp + duration` and no sanity bound. Any divergence between the two (stale historical
NS record, a different NS-writing client with a different convention, clock skew at write time,
etc.) surfaces directly as garbage in the UI, since nothing downstream re-derives or clamps it.

## Suggested fix

```kotlin
if (showRemainingTime && profileSwitch.originalDuration != 0L) {
    profileName += dateUtil.untilString(profileSwitch.timestamp + profileSwitch.originalDuration, rh)
}
```

This makes the display consistent with how `ProfileSwitchExpiryScheduler` and `SceneExecutor`
already treat this field, and removes the dependency on a field the data model itself documents as
unused.

## Confirmed present on

- `original/dev` @ `9df0b1bd4e` ("fix ProfileHelper and EPS caching")
- `original/dev3` @ `ab1f6040cc` ("Cleanup")

## Repro

Observed on a fork build (AAPSCLIENT/follower widget) while a temporary ProfileSwitch (`120%`,
fixed duration) was active: the widget's active-profile line showed
`#26 <profile name> (120%)(32399999998ч 11')` instead of a plausible remaining duration.

JetFoxy:
## Summary

Fixes #4985 — `ProfileFunctionImpl.getProfileName()` reads `EPS.originalEnd` directly to render the
remaining-time suffix on the active-profile display (Overview ribbon + home-screen widget). That
field is documented on the entity itself as `// not used (calculated from duration)`, and every
other consumer of `EPS` already treats it that way:

- `ProfileSwitchExpiryScheduler`: *"the EffectiveProfileSwitch query selects the latest EPS by
  timestamp and ignores `originalEnd`"*
- `SceneExecutor`: *"...and ignores originalEnd, so the base profile only resumes once a NEW
  base-profile EPS exists."*

When a given EPS record's stored `originalEnd` doesn't actually match
`timestamp + originalDuration`, `getProfileName()` renders a nonsensical duration (e.g.
`32399999998h` instead of a plausible remaining time).

This is most reliably triggered on **AAPSCLIENT**, since a follower never creates EPS locally
(`CommandQueueImplementation`'s creation path is gated by `if (config.AAPSCLIENT) return`) — every
EPS it has comes from Nightscout verbatim, `originalEnd` included, with nothing re-deriving or
sanity-checking it. But it's not actually AAPSCLIENT-specific: a **master** install also inserts new,
unvalidated EPS records sourced from Nightscout whenever `NsClientAcceptProfileSwitch` is enabled or
a full resync happens, so a stale/mismatched `originalEnd` on one of those records produces the same
garbage on a regular pump-owning install.

## Fix

Derive the end time from `timestamp + originalDuration` instead of trusting the stored
`originalEnd`, matching how the rest of the codebase already treats this field.

## Test plan

- [x] `:implementation:compileFullDebugKotlin` builds clean
- [ ] Maintainers/reviewers to confirm no existing test asserts on the old (buggy) `originalEnd`-based behavior

### DIFF
--- a/implementation/src/main/kotlin/app/aaps/implementation/profile/ProfileFunctionImpl.kt
+++ b/implementation/src/main/kotlin/app/aaps/implementation/profile/ProfileFunctionImpl.kt
@@ -80,7 +80,12 @@ class ProfileFunctionImpl @Inject constructor(
         if (profileSwitch != null) {
             profileName = if (customized) profileSwitch.originalCustomizedName else profileSwitch.originalProfileName
             if (showRemainingTime && profileSwitch.originalDuration != 0L) {
-                profileName += dateUtil.untilString(profileSwitch.originalEnd, rh)
+                // originalEnd is not reliable (see its declaration in EPS.kt: "not used (calculated from duration)")
+                // and other consumers of EPS (ProfileSwitchExpiryScheduler, SceneExecutor) already ignore it in
+                // favor of timestamp + duration. A stale/mismatched originalEnd on a record synced in verbatim
+                // from Nightscout (AAPSCLIENT, or NsClientAcceptProfileSwitch / full resync on master) otherwise
+                // renders a nonsensical duration here.
+                profileName += dateUtil.untilString(profileSwitch.timestamp + profileSwitch.originalDuration, rh)
             }
         }
         return profileName


JetFoxy:
Disclosure: this PR (the code, the commit message, and this description) was produced with AI assistance (Claude Code) on top of my own diagnosis of the widget bug, then reviewed and compile-checked by me before opening.

I noticed only after opening this that `CONTRIBUTING.md` on `dev` says AI-generated PRs are not welcome — see my comment on #4985. Deferring to the maintainers on whether to keep this open given the disclosure, or close it and have the fix reimplemented directly by a maintainer from the issue's analysis.

PR #5003, opened 2026-07-24
Retry Carelevo BLE session on handshake disconnect

Kisu-Choi:
## Summary

This PR addresses the issue where `Suspend pump` could fail on the first attempt when a temp basal or extended bolus was active.

## Background

During testing, increasing the inter-session delay from 1 second to 2 seconds significantly reduced the failure rate. This suggested that the issue was related to a timing-sensitive BLE reconnect problem between consecutive sessions.

## Change

Instead of increasing the fixed delay for all BLE operations, this PR keeps the existing 1 second inter-session settle time and adds a single retry when the BLE session is disconnected during the handshake stage before command execution begins.

## Rationale

This approach is intended to improve stability, particularly on slower or timing-sensitive devices, while avoiding unnecessary delay for all session-based BLE operations.


MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/commit/0e1a25b47f75ed7913c291701725129dc015a582
verify

Kisu-Choi:
Thanks, I verified it on our side, and it works well now.

PR #5008, opened 2026-07-25
Wear. Offbody detection

dkustov:
Samsung wear sends fantom heartrate and steps at off-body time or when it charging.
I'm suggesting to suppress this signals using off-body sensor or detecting charging mode

### DIFF
--- a/wear/src/main/kotlin/app/aaps/wear/heartrate/HeartRateListener.kt
+++ b/wear/src/main/kotlin/app/aaps/wear/heartrate/HeartRateListener.kt
@@ -2,10 +2,13 @@ package app.aaps.wear.heartrate
 
 import android.content.Context
 import android.content.Context.SENSOR_SERVICE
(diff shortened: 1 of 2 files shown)


dkustov:
I have troubles to connect  circleci from Russia (even using vpn). Could you please tell me whats wrong with tis PR?

MilosKozak:
I didn't found any evidence this is happening and need fixing
Any other with this issue?

Philoul:
On my side (GW4), I don't have this issue for HeartRate, but I have it for Steps (see screenshot below during my night with my watch charging...)

dkustov:
@Philoul - this is actually a separate issue. There is a bug with the step source: steps were received from the phone instead of the watch due to a bug in StepService.kt. To resolve this, we need to fetch the steps from the DB and save them with the correct source (watch, phone, etc.).

PR #5012, opened 2026-07-26
Overview graphs (Dynamic Y-axis scaling according to visible-window + nice numbers)

Philoul:
I know I touched tricky files... It's my first development with the help of Claude Code... (a strong guidance of CC was needed, and I had to learn how to work with him, but I could not do this PR without him...)

## What
Y-axis scaling for the Overview graphs (BG, IOB/BAS, and all configurable secondary graphs)
now adapts to the currently visible (scrolled/zoomed) window instead of the full 24h loaded
range — calm periods are no longer flattened by a spike elsewhere in the loaded data.

- Axis bounds and tick spacing now snap to round ("nice") numbers instead of raw decimals.
- Dual-axis graphs (two series combined, e.g. IOB+BGI) share a common zero position without
  ever clipping either curve — handles zero-floor, pivot-centered (SENS/DEV_SLOPE), and
  free-range (VAR_SENSITIVITY/HEART_RATE) combinations.
- IOB/BAS basal overlay headroom and tick labels adapt correctly to the windowed scale.

## Bug fixes
- Fixed BG graph pinch-zoom becoming unresponsive during a live gesture — axis range
  providers were being recreated on every scroll tick, forcing Vico to rebuild the chart and
  re-register gesture handling mid-gesture.
- Fixed BG viewport snapping to the wrong zoom/scroll after an auto-reset (new BG reading) —
  now recreates the scroll/zoom state objects instead of driving `VicoZoomState.zoom()` to an
  absolute target.
- Fixed COB's Y-axis getting force-symmetrized around a fake negative range (e.g. -42/+42)
  when sharing a dual-axis graph with a zero-crossing series, even though COB itself is never
  negative — zero-floor series now get their own dedicated floor-at-0 axis construction
  instead of reusing the generic symmetric zero-crossing alignment.

## Architecture notes
- `BgGraphCompose`: new `MutableYRangeProvider` — a `CartesianLayerRangeProvider` with a
  stable identity, mutated in place instead of recreated (`.fixed(...)` always returns a new
  instance, which was the root cause of the pinch-zoom bug above). Needed only for BG, the
  only interactive chart.
- `GraphUtils`: new nice-number scaling helpers (`niceScale`, `zeroFloorNiceRange`,
  `niceScaleAroundPivot`) and a passive `VisibleRangeReporter` decoration reporting the
  visible window without touching Compose state during the draw pass.
- No public API changes; all new parameters default to the previous behavior.

## Tests
Added/updated unit test coverage for the new scaling logic (nice-number rounding, zero-floor
and pivot-centered ranges, dual-axis alignment).

## Docs
Detailed writeup of the design and root causes of each bug fix added at
`_docs/graph-dynamic-scaling.md`.

### DIFF
--- a/_docs/Graph Dynamic Scaling.md
+++ b/_docs/Graph Dynamic Scaling.md
@@ -1,0 +1,194 @@
+# Overview Graphs: Dynamic Y-Axis Scaling & Nice Numbers
+
+## Motivation
+
+Overview graphs load 24h of data but typically show a 6h sliding window (zoomable down to
+30 minutes, up to 24h). Previously, each graph's Y-axis scale was computed by Vico from the
+**entire loaded 24h**, not the currently visible portion. A single spike anywhere in the 24h
+(e.g. a large IOB peak) would flatten every calm period visually, even when scrolled directly
+onto it.
+
+This change makes every graph's Y-axis recompute from the **currently visible (scrolled/zoomed)
+window** instead, and additionally snaps axis bounds/ticks to round ("nice") numbers instead of
+raw data-driven decimals.
+
+Scope: `BgGraphCompose.kt`, `SecondaryGraphCompose.kt`, `GraphsSection.kt`, `GraphUtils.kt`.
+No public API changes — all new composable parameters default to the previous behavior.
+
+---
+
+## 1. Visible-window axis scaling
+
+### Mechanism
+
+Vico (the charting library) does not expose the visible x-range at the Compose level
+(`VicoScrollState`/`VicoZoomState` give scroll pixels and a zoom factor, not a resolved time
+window). The only place this is reliably computable is at **draw time**, via
+`CartesianDrawingContext` (`layerDimensions.xSpacing`, `ranges.minX`, `ranges.xStep`, `scroll`,
+`layerBounds`) — the same context the existing "now" line decoration (`NowLine`) already reads.
+
+- **`VisibleRangeReporter`** (`GraphUtils.kt`): a passive `Decoration` that computes the visible
+  x-range on every draw pass and writes it into a plain `@Volatile` holder
+  (`VisibleRangeHolder`) — deliberately **not** a Compose `State`. Writing Compose state
+  synchronously from the draw phase was found to fight with Vico's own gesture-driven
+  scroll/zoom mutations (observed: pinch-zoom became unresponsive — see Bug Fix #1 below).
+- Each graph polls this holder from a `LaunchedEffect` (every 50ms), then debounces (30ms)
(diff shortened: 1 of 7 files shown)


vanelsberg:
@Philoul Defenitely an improvement of the graphs❤️. Well done! 

First impressions and observations:
I found I do need to adjust to the dynamic vertical scaling. Not seeing this as a real problem.
Do have vertical scaling for the main graph at 150%

1) Main Dual-axis graph:
BG curve for me now only uses the top-half of the graph (with or without enabling "activity"), target is center and below is low BG. AS BG will only sparsely use that space or (hopefully) go to 0. Maybe have a look at that.
Suggestions:
- When activity is not selected, scale on minimum/maximum BG
- When activity is selected, use dynamic scaling and set zerobase for activity (it is never negative) at BG target? Thinking this way the full graph space can be used.

2) Main Dual-axis graph: bug
When scrolling all te way to the left (no more data) dynamic scaling stops working and vertical scale is 0..15?
Also visible when pinch-zooming out to max, and then swiping left/right

At first view, haven't found any issues. Will use & validate this PR for the coming days for any important issues.
👍

Philoul:
Concerning BG graph, I kept bottom unchanged (starting from 0, even if I agree space below 40mgdl or equivalent in mmol is never used), and max is limited to upper range (180mgdl or 10mmol by default).If your BG values are all within range (a bit extended with nice scale) you will not see scaling adjustments.Can you share with me a little video (PM on discord to let me see the issue please?)

Philoul:
Hum, I understand your "locked" scale with mmol units:
Nice scale use "nice" numbers, so starting with 0, if max is 10mmol or just a bit above 10mmol, the nice scal can increase scale up to 15 mmol, way too much...
Will think to not start minimum value, but from minimum hypo (40mgdl or min range)...
This should fix the "empty space" in the bottom, and improved the max locked value...

Philoul:
@vanelsberg 
With my latest commit I replaced the hardcoded 0 mgdl (or 0 mmol) as floor of BG graph by min Range (70 mgdl or something close to 4.0 mmol).
- this fix your both issue/comments: less empty space in the bottom, and with the "nice scale" you don't directly jump from 0-10 to 0-15 mmol, but something much more progressive (4-6-8-10 -> 4-6-8-10-12 ...), so less empty space on the top.
- I checked with mmol and mgdl, both are improved 👍

vanelsberg:
@Philoul Tnx! Latest commits solved my "issues" previously mentioned.

I think this PR definitely improves fot the V4 graphs! Now running this PR on my main AAPS and monitoring for any any problems. So far all is doing just fine!

PR #5042, opened 2026-08-01
Watchface Activation for Samsung recent watch (waiting for tests to confirm it works before merge)

Philoul:
On newer "Factory Built" Wear OS 5+ Samsung watches, the on-watch picker no longer offers code-based watch faces, so CustomWatchface may be impossible to activate at all. This adds a fallback: opening CWF's settings from the AAPS menu now sends a broadcast asking Samsung SysUI to activate and open the editor for CustomWatchface, Digital or Circle

This uses a standard, Samsung-authorized entry point (exported=true, auto-granted normal permission) — the mechanism itself is legitimate and documented as callable by third-party apps, only the specific intent extra key was found by inspection rather than official docs.

Fully safe on non-Samsung watches or older Samsung builds: the receiver is checked before sending, and the fallback is the exact existing behavior (nothing sent, screen shown normally).

I'm looking for tester... (an owner of recent Galaxy Watch no more compatible)
- To activate the watchface, open AAPS app, go to Preferences > At the end you have AAPS (Custom), AAPS (Digital) and AAPS (Circle).
- Click and wait a bit, it should normally open the dedicated preferences attached to each of these Watchfaces (but before it normally should activate the corresponding watchface)

It works with my GW4, and I want to check if it also work with latest GW7, GW8 ...

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -10,6 +10,13 @@
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.BODY_SENSORS" />
     <uses-permission android:name="android.permission.ACTIVITY_RECOGNITION" />
+    <uses-permission android:name="com.google.android.wearable.permission.RECEIVE_COMPLICATION_DATA" />
+    <!-- Samsung-only, protectionLevel="normal" so it is granted automatically at install and is a
+         no-op on non-Samsung watches (the permission simply doesn't exist there). Lets us send
+         ACTION_EDIT_WATCH_FACE to Samsung SysUI's exported WatchFaceEditorBroadcastReceiver, which
+         is the only known way for an app to get the system to start a watch face editing session -
+         see ComplicationPickerSupport for why that session is required. -->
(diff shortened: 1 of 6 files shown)


Philoul:
Unfortunatly after tests by users with GW7 or Ultra uptated with latest firmaware (no more compatible with Wear Installer 2 Workaround), activation of our 3 watchfaces are also blocked so these commits doesn't work (it only work on GW4-GW6, but not needed, and probably still working on GW7 not updated with recent firmware).
It seems this code can only improve the situation for GW7 users who have blocked firmware updates to keep the WearInstaller 2 workaround still working (then watchface activation is not limited to Wear Installer 2 latest step, but could also be done from AAPS menu). 
Unfortunalty the number of users in this situation will probably be fewer and fewer...
For all the others, it doesn't work (Updated GW7 or newer) or it is not necessary (for GW4 to GW6 watchs, activation or access to preferences can be done from long press on the watchface).
@MilosKozak Should we keep this code or just simplify preference menu and remove the latest Watchface items ?

olorinmaia:
@Philoul @MilosKozak Since this fix doesn't work as we hoped for, and doesn't add anything for the watches CWF already works for, I suggest we revert this. The proper fix for CWF is probably another path :)

PR #5047, opened 2026-08-02
Add 5 complication slots within Custom Watchface

Philoul:
Add 5 complication slots within CustomWatchface:
- 5 new fields the user cas customize the same way it's done for the other field (width, height, ...) (visibility gone by default)
- 5 new preference key to show or hide complications slots: A user can decide to use complication or use the standard view defined within a Custom Watchface
- Default presentation for complication can be adjusted to fit watchface styling (Complication Title color/font, text color/font/style, icon color, ...)
- DynData and DynPref can be used to dynamically adjust Complication view according to AAPS data (BG Color...)
- Watchface preferences in now filtered to only show Preferences used within the loaded Watchface (i.e. if the selected Watchface is V1 or V2, Complications views are missing so complication preferences are hiden)

See screenshots with Analog G-Watch V3 used for my tests (I will include this zip within assets):
- this Watchface has 4 complication slots: Complication 1  and 2 (circle) for left and right area, Complication 3 rectangle in the middle, and Complication 4 (same size than Graph).
- In screenshot below you see the result with all Complication disabled in Watchface Preferences, all enabled (HeartRate, Step, Exercise, BG Graph Complication), and a third screenshot without Complication 2 and 4
CustomWatchfaceComplication

And the zip file if you want to see an example:
[Analog G-Watch_V3.zip](https://github.com/user-attachments/files/31117040/Analog.G-Watch_V3.zip)

Know limitation:
- Complications Slots are built during th Watchface load (so when you select the CustomWatchface within Watch), So selected complications are shared accross all CWF (if you change the zip file, you will probably have to selet again which complication you want to see where...)
- Samsung complications are sometimes limited (HeartRate complication don't show "live" the Hear Rate but only the Title, just because our watchface is considered as "Unsafe", not installed from PlayStore). Hopefully theses complication are often available with thirdparty provider

Now I need tester (and Watchface designer) to first test this proposal, and use this new feature in the other CWF...

Development Checklist:

- [x] 5 real `ComplicationSlot`s declared, positioned/sized/visible from CWF json
- [x] Rendering wired into the render loop, correct z-order for the tested CWF
- [x] Taps routed by the framework (no manual hit-testing)
- [x] Long-press "Customize" picker works end-to-end, providers persist correctly
- [x] Legacy-API false start replaced with real androidx implementation (Root Cause 1)
- [x] Dagger injection race in `createComplicationSlotsManager()` fixed (Root Cause 2)
- [x] `EditorSession` headless-instance crash-on-close fixed (Root Cause 3)
- [x] Black-screen-after-reinstall bug fixed (`loadData()` sync override) (Root Cause 5)
- [x] Temporary debug instrumentation stripped, permanent fix documented
- [x] Show currently-assigned provider name under "Complication N" in the preferences menu
- [x] AAPS Settings-menu picker doesn't open (root cause should be fixed with #5042)
- [x] **Complications aren't integrated into the CWF's live per-tick refresh flow** — no
  bounds/position/background update on CWF reload, and no path at all for live `DynProvider`-
  style changes (e.g. BG-level-driven resize/reposition) within the same CWF (see "Open bugs"
  below) — **next up, with Opus** (moved ahead of the menu picker fix, which is on hold —
  see below)
- [x] AAPS Settings-menu picker doesn't open
- [x] Restrict Settings-menu complication entries to CustomWatchface + visible slots only —
  **on hold**, same reason; also depends on resolving how non-`CustomWatchface` code may
  obtain CWF-derived state (see the hard rule in `CWF_ComplicationSlotsRules.md`)
- [x] **Complication visible content (icon/text) renders far smaller than the declared CWF
  bounds box, by a large margin** — confirmed via device testing (see "Open bugs" below)
- [x] **Comprehensive complication styling/display parameter audit** — the visual-polish items
  below are currently a loose list of individually-discovered problems, not a designed
  parameter set. Before or alongside the live-refresh work, produce a full inventory: every
  display parameter a complication slot could reasonably need (text/title font, size,
  color, weight; icon color/size; background; which field is shown when a provider sends
  both a label and a value; content padding/inset; anything else `ComplicationStyle`
  exposes), map each to an existing or new `JsonKeys`/`JsonKeyValues` entry, and organize
  the runtime side coherently (a `ComplicationStyleMap`-equivalent, alongside
  `ComplicationProviderMap` already planned below) rather than continuing to bolt on
  one-off keys per bug. Worth doing now since the live-refresh mechanism will need to
  re-apply this full parameter set on every tick anyway — better to know the complete set
  once than to revisit the refresh code each time a new parameter is added.

Visual/UX polish (from screenshot comparison against an OEM watch face):

- [x] Heart rate complication shows static label instead of live value (Samsung limitation for side loaded Watchface not coming from Play Store)
- [x] Complication bounds sizing guidance (square should contain the cutout circle, not the
  other way around) — CWF-authoring guidance, not code
- [x] Icon position (above/below text) — investigate whether framework- or provider-driven

Design work, not yet started:

- [ ] `DynProvider` integration for numeric/rangeable complication values


Philoul:
Branch updated with latest dev and kmp migration...

Issue #4965, opened 2026-07-10
Replace average and exponential smoothing with UKF In V4


tim2000s:
Having three separate smoothing algorithms in V4 isn't helpful to users and creates a complex, difficult to understand challenge in relation to how the different options work and what they do, as the explanations don't fit within the space in the config builder. 

The data here (https://github.com/tim2000s/Smoothing-Investigations) highlights that UKF is much more effective than exponential for many reasons, so instead of leaving exponential in place, don't release it as part of the v4 release and replace it with UKF. 

Additionally take the Average smoothing option away as it is technically not a smoothing algorithm, it is a retrospective adjustment to historic data and simply changes the delta value between now and the previous values rather than smoothing the current value to a safer one.

Instead offer:

No Smoothing
Advanced Smoothing ( The UKF smoothing options, and make it default for G7/One+/Stelo)

tim2000s:
The content of the GitHub repo has been updated to include cohort data from the data commons. 

The headlines are shown in the attached table. Across 183 users and approximately 8.3mn data points, the UKF is shown to be better. Suffice to say, if UKF is available, there is no point in retaining the exponential smoother. 

This arguably suggests that we shouldn't have more than one smoother and it should be the UKF version, as the "Average" is entirely retrospective and doesn't stop anomalous spikes from the likes of the G7. The AID system should be used with the safest, as selected by those who design it, not left to the users to pick. 

<img width="2050" height="900" alt="Image" src="https://github.com/user-attachments/assets/7ef5a7bb-ec1e-4a56-b377-f093176fc8fb" />

olorinmaia:
I haven't tested the new UKF with real BG data yet. But my main issue with exponential is the lag. This is also my concern for UKF, but I need to test it more with live data. (I'll get some bg from NS live on to my test rig and compare.) 

I agree we should remove exponential but I'm not sure about average smoothing which is 100% lag free.

Automations can also detect compression lows. Edit: we use it spesifically to block smb at night if certain conditions are met. We still get alarms from xDrip when low.

jbr7rr:
I support removing options to make it simpler for the user, that said some comments:

> Average smoothing option away as it is technically not a smoothing algorithm

This is not true, average smoothing is the simplest possible smoothing possible, and while it doesn't change (or predict as you will) the current value, it will change the delta which is the most important thing for openAPS in terms of treatment decisions

[...]

Issue #4677, opened 2026-04-03
Feature request: Offer Smoothing choice as an automation action


Navid200:
It could be very useful to be able to switch the filter (smoothing) option based on sensor age.
A Dexcom G7 has a lot more fluctuations during the first 24 hours.  It would be really nice to be able to automatically switch to second order exponential (more aggressive filtering at the cost of more delay) for a new sensor.  And switch back to average smoothing (less aggressive filtering with less delay) after sensor age becomes greater than 24 hours for example.

Sensor age is already an existing condition.
But, I don't believe it is currently possible to control the smoothing option as an automation action.

olorinmaia:
For your information I have submitted https://github.com/nightscout/AndroidAPS/pull/5056 related to this feature request.

olorinmaia:
@Navid200 @MilosKozak This issue can be closed as #5056 is merged to dev (AAPSV4)

olorinmaia:
Closing issue as https://github.com/nightscout/AndroidAPS/pull/5056 is merged to dev (AAPSV4)

Issue #3823, opened 2025-02-12
[Feature Request] Smoothing variant automation action


WesSec:
It would be super awesome if we could set the smoothing setting in automations.

Use Case: often a sensor is placed not ideal, and I get compression lows in the night. Exponential smoothing prevents the loop from acting upon these rapid changes. This only happens during the night; at daytime the loop can benefit from detecting the rapid changes (for instance in combination with AutoISF)

If we can schedule the smoothing by automations, we can benefit rapid change detection during daytime, and use smoothed values during nighttime.

olorinmaia:
Some other use case for this. Could use automations to change to exp smoothing when sensor age is less than 1 day or so to help even out the bumpy start of sensor life and then change to average smoothing after 1 day.

gerison77:
I just had this usecase the day before yesterday and posted a likewise proposal on discord. Just opened github to write a feature request and found this thread. I am already running an automation to sett high TT when a instant drop in BG takes place during the night, but it just doesn't work that nice, since it leads to neg IOB which is set off later on leading to the "real bump"

olorinmaia:
For your information I have submitted https://github.com/nightscout/AndroidAPS/pull/5056 related to this feature request.

olorinmaia:
@WesSec @MilosKozak This issue can be closed as https://github.com/nightscout/AndroidAPS/pull/5056 is merged to dev (AAPSV4)

olorinmaia:
Closing issue as https://github.com/nightscout/AndroidAPS/pull/5056 is merged to dev (AAPSV4)

PR #5056, opened 2026-08-03
Automation: new action to change smoothing type

olorinmaia:
New automation action "Change smoothing" that switches the active smoothing plugin (Configuration → Smoothing). The dropdown lists the installed smoothing plugins, so it always matches the configuration screen.

Typical use with the existing "Sensor age" condition:
- Sensor age < 24 h → Change smoothing to Unscented Kalman Filter
- Sensor age ≥ 24 h → Change smoothing to Average smoothing

New sensors are often noisy the first 24 hours. This lets users run stronger filtering while the sensor settles, then return to lag-free smoothing.

- The action calls `ConfigBuilder.performPluginSwitch()`, the same path as the Configuration screen: selection is persisted, logged to UserEntry, synced to paired clients, and BG data is re-smoothed and recalculated automatically.
- Added a precondition (same pattern as scene/TT actions) so the automation is skipped silently once the selected smoothing is active
- The stored plugin is validated in `isValid()` — if a smoothing plugin is removed in a future version, the action is marked invalid instead of failing.

**Bugfix included:**
Fix for an existing UI bug in the automation action editors: dropdowns and the SMB toggle did not show the newly selected value until the screen was reopened. Fixed for profile switch, SMB change, the three scene editors and the new smoothing editor.

**Closes:**
#3823 and #4677

**Related:** 
#4965 — this feature depends on keeping more than one smoothing option; it automates the exact workflow described by a user in this comment: https://github.com/nightscout/AndroidAPS/issues/4965#issuecomment-5056491643 — I agree on removing Exponential smoothing in favor of the new UKF, but disagree with removing Average smoothing which is preferred for our setup when sensor is stable. See my comment https://github.com/nightscout/AndroidAPS/issues/4965#issuecomment-5172112227

Tested ok on AAPS/AAPSClient. :plugins:automation:test passed

**Screenshots:**
| image | image |
|--------|--------|
| image | image |

### DIFF
--- a/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/AutomationEventTest.kt
+++ b/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/AutomationEventTest.kt
@@ -44,7 +44,7 @@ class AutomationEventTest : TestBase() {
         ActionFactory(
             triggerDeps, aapsLogger, rh, pumpEnactResultProvider, mock(), dateUtil, mock(), mock(), mock(),
             profileFunction, mock(), mock(), mock(), mock(), mock(), preferences, mock(), mock(), mock(), mock(),
-            mock(), mock()
+            mock(), mock(), mock()
         )
     }
 
--- a/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/AutomationJsonRoundTripTest.kt
+++ b/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/AutomationJsonRoundTripTest.kt
@@ -52,7 +52,7 @@ class AutomationJsonRoundTripTest : TestBase() {
         ActionFactory(
             triggerDeps, aapsLogger, rh, pumpEnactResultProvider, mock(), dateUtil, mock(), mock(), mock(),
             profileFunction, mock(), mock(), mock(), mock(), mock(), preferences, mock(), mock(), mock(), mock(),
-            mock(), mock()
+            mock(), mock(), mock()
         )
     }
 
--- a/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/actions/ActionSmoothingChangeTest.kt
+++ b/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/actions/ActionSmoothingChangeTest.kt
@@ -1,0 +1,108 @@
+package app.aaps.plugins.automation.actions
+
+import app.aaps.core.data.plugin.PluginType
+import app.aaps.core.interfaces.plugin.PluginBase
+import app.aaps.plugins.automation.AutomationStrings
+import com.google.common.truth.Truth.assertThat
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.mockito.Mock
+import org.mockito.kotlin.any
+import org.mockito.kotlin.anyOrNull
+import org.mockito.kotlin.never
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.skyscreamer.jsonassert.JSONAssert
+
+private const val STRING_JSON = """{"data":{"smoothingPlugin":"AvgSmoothingPlugin"},"type":"ActionSmoothingChange"}"""
+
+class ActionSmoothingChangeTest : ActionsTestBase() {
+
+    @Mock lateinit var avgSmoothingPlugin: PluginBase
+    @Mock lateinit var noSmoothingPlugin: PluginBase
+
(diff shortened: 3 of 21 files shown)


MilosKozak:
well when sensor is stable, like you say, differences between both algs should be minor

olorinmaia:
> well when sensor is stable, like you say, differences between both algs should be minor

I'm guessing you are referring to my comment regarding #4965 - when the BG is steady and sensor is stable the difference is minimal. But when BG is falling rapidly or rising rapidly - it is not. This is where Average smoothing in my opinion is a better smoothing algorithm than UKF - for kids especially. Stuff happen, lots of IOB, sudden activity. I also prefer to see what the most recent actual BG is, not what a smoothing alg think it is.

MilosKozak:
in this case: high IOB, negative deviations .... there will be zero temping independent to smoothing

olorinmaia:
True, there is probably no difference in APS output in this scenario. But that's isn't my main concern. My main point is that AAPS / Client is our most common view for looking at BG or graphs on phone/watch. Seeing a big BG drop smoothed by UKF with additional lag can delay our response to act (e.g. treat a potential incoming low with carbs) compared to that Average always gives us the actual recent BG. It will also cause a need to more often check xDrip for what the BG value from sensor is. 

I think the best way forward at least now is to merge this PR and remove Exponential. Perhaps in future when UKF or similar smoothing alg is even better and trusted we can remove Average also. Average is unique in terms it shows actual recent BG value, and smooths retrospective. A decision should probably be made in regards to #4965 before PR is either merged (remove Exp) or closed (remove Average and Exp). A third possibility is to add an option to UKF, not having it to alter the most recent BG.

UKF is new and not yet proven by masses plus we know Average works well with most sensors. I think for kids especially with more erratic behavior Average is the best option right now. For adults who got full control of their T1D UKF will most probably always be the best option I would imagine.

nl-ruud:
Apart from the first 24 hours, I also switched to exponential smoothing mainly because of nighttime issues. My child can be quite restless while sleeping, which sometimes makes the CGM readings jumpy or causes compression lows. Without sufficient smoothing, the algorithm can react too aggressively to those fluctuations and ultimately cause a real hypo.

I know I could disable SMBs at night, but that isn’t ideal either, as they can still be very useful when hormones kick in or after meals with delayed carbs.

So I currently keep exponential smoothing enabled most of the time, with the downside being the added lag during the day. This automation would mean I no longer have to make that trade-off.

olorinmaia:
@MilosKozak 
I've done some further comparing of UKF and Average Smoothing. Another issue is that UKF continues the rise, allthough the actual BG reading was a drop or a flat/similar BG of previous. This can inject an additional unnecessary SMB in some scenarios. The delay/lag here is also very noticable, and like I've pointed out earlier, one of the reasons I don't like Exponential smoothing, and UKF doesn't seem to be much better at this either unfortuntately.

Some screenshots for compare. The BG at left is from AAPS 4 using UKF smoothing feeded from xDrip Dexcom Share G7. Right is AAPS V3.4 using Average smoothing feeded from xDrip G7. The screenshots start at 13:40

| UKF | Average |
|--------|--------|
| image | image |
| image | image |
| image | image |
| image | image |
| image | image

It's not before 14:19 (first screenshot is 13:40 so almost 40 minutes) before the BG becomes even cloes. UKF doesn't simply believe that the drop is real before a long time has passed.. This makes it clear that UKF is to laggy for my taste.

| UKF | Average |
|--------|--------|
| image| image| 

This underlines my view on #4965 - we should NOT remove Average smoothing. We should remove Exponential smoothing and keep UKF and Average. This automation will enable users to get the best out of both smoothing algorithms without having to manually switch.

vanelsberg:
> I think the best way forward at least now is to merge this PR and remove Exponential.

👍Fully agree on that!
This has been also my main argument against UKF (and also Exponential Smoothing) as you documented in your previous post (one of the reasons I never use(d) Exponential Smoothing)

Agree with @MilosKozak that eventually algo (BG graphs are of no importance) will compensate IOB. That is also why imho smoothing shoot only be about filtering noise, mostly to have better delta data. Most recent BG imho is best (so not filtered) for several reasons as you mentioned.

vanelsberg:
Would love to see smoothing algo as an automation option👍
Find myself manually switching smoothing, especially at the G6/G7 start and end of life.

_Sidenote_: Actually, I would give more value to a feature that can "detect" sensor noise and instability. It could warn the user (just like the warning on carbs absorbing slowly?) and maybe even be input for a smoothing algo (just like sensitivity does for profile?) or even as a trigger for automations.

vanelsberg:
> **Related:** #4965

- As is now AAPS gives user a choice. We should not want to remove that choice!?
- UKF is new and not yet proven by masses plus we know Average works well with most sensors.
- Lagg: See [comment above](https://github.com/nightscout/AndroidAPS/pull/5056#issuecomment-5204725424)

PR #5062, opened 2026-08-07
Wear: running mode complication and picker + updated AAPS V4 WFF watchface

olorinmaia:
### What

- New complication showing the current running mode as a color icon (SMALL_IMAGE / ICON, image-only). 
- Tap opens a new running mode picker: all allowed mode changes (including Disable loop, which the 4-slot tile hides), current mode + remaining time in the header, same confirm flow as the tile.
- Picker also added to the wear main menu, for watchfaces without tiles or complications. BG graph moved next to it.
- New "Running mode" option for the complication tap action setting.
- Phone now sends the current mode (+ end time of temporary modes) with Status, and re-sends when the mode changes -complication and tile update right away instead of one loop cycle later.
- Disconnect/suspend request page shows the requested mode name, same layout as the temp target page.
- AAPS V4 watchface updated: two new top slots (running mode + uploader battery) with default providers bound.

### Compatibility

Old phone with new watch (and reverse) is safe: new Status fields have defaults and unknown fields are ignored.

### Tests

New: RunningModeComplicationTest, Status round-trip + legacy decode in EventDataTest, tile filter + picker overload in RunningModeSourceTest. 

:wear:test :core:interfaces:test :plugins:sync:test :core:ui:test passed

Tested OK on AAPS/AAPSClient wear

### Screenshots
image

image


vanelsberg:
Tested latest commit#d68f4e0a. Works great for my Samsung GW6. Did need to change settings for the WF to keep showing data when the watch goes into locking state (Samsungs Wrist detection):

`Settings->Security & Privacy->Hide information=off`

Latest commit has 2 additional slots for complications at the upper left and right: I use them for showing the AAPS snooze "button" and Phones battery level.

olorinmaia:
Further testing with other WFF watchfaces, like DMM13 from Glucodatahandler the same issue is present there. A quick google search also confirms that this is WFF related, and setting needs to be toggled by the user. Old code-based watchfaces aren't using complications and aren't affected. Samsung watchfaces may also be "whitelisted" from this, but there doesn't seem to be any way to bypass it by code. We can consider leaving a note in wiki regarding this and the WFF watchface.

image

PR #5064, opened 2026-08-09
CI: pass signing secrets via env: instead of interpolating them into shell

GrantAdkins:
Signing fails for any password containing a shell metacharacter, with an error that blames the wrong input:

```
jarsigner error: keystore load: Keystore was tampered with, or password was incorrect
```

**Cause.** Secrets are interpolated into the `run` block's script text, so bash parses the password as source before it ever becomes a value:

```yaml
echo "KEYSTORE_PASSWORD=${{ secrets.KEYSTORE_PASSWORD }}" >> $GITHUB_ENV
```

```console
$ P='mU3PftciJU&Xy$9zAb!c'                # 20 chars
$ eval "echo \"KEYSTORE_PASSWORD=$P\""
KEYSTORE_PASSWORD=mU3PftciJU&XyzAb!c      # 18 chars — $9 expanded to nothing
```

jarsigner receives the truncated value, and the reported failure points at the one input that was never wrong.

**It also defeats log masking.** Once mangled, the value no longer matches the registered secret, so the runner cannot redact it:

```
KEYSTORE_PASSWORD: mU3PftciJU&***
```

On a public fork that places part of the user's signing password in the run log.

**Fix.** Pass secrets in via `env:` and write `GITHUB_ENV` in heredoc form, which is safe for any value including newlines. Applied to the four build workflows and to `keystore-export.yml` — the latter also writes the decoded password into `keyStoreInfo.txt` inside the exported zip, so a mangled value would be handed back to the user as their recorded signing password. Values derived from `KEYSTORE_SET` keep explicit `::add-mask::` calls, since they are not registered secrets.

Verified on a fork with the same secrets and keystore: failing before, green after, password fully redacted in the log.

### DIFF
--- a/.github/workflows/aaps-ci.yml
+++ b/.github/workflows/aaps-ci.yml
@@ -64,35 +64,40 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Decode Secrets Keystore Set and Oauth2 to Env
+        env:
+          # Secrets arrive as environment variables so bash never parses them as
+          # source text. Interpolating them into the script body lets any $, ",
+          # backtick or newline in a password corrupt the value silently.
+          S_KEYSTORE_SET: ${{ secrets.KEYSTORE_SET }}
+          S_KEYSTORE_BASE64: ${{ secrets.KEYSTORE_BASE64 }}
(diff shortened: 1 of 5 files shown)


MilosKozak:
replaced by https://github.com/nightscout/AndroidAPS/pull/5116

PR #5072, opened 2026-08-11
Wear: fix install error on Wear OS 3 watches (missing wear-sdk library)

olorinmaia:
The Watch Face Push library (added in #5022) marks the `wear-sdk` platform library as required. That library only exists on Wear OS 4+, so installing the wear APK on Wear OS 3 watches (e.g. TicWatch E3) failed with `INSTALL_FAILED_MISSING_SHARED_LIBRARY`.
image

This overrides the entry to `required="false"` in our manifest. Safe because all Watch Face Push code is already behind a Wear OS 6+ runtime check — on older watches the watchface install menu simply does not appear.

Tested OK on Wear OS 6 (watchface push still works). 
Tested OK on Wear OS 3 Virtual device:
image

Tested OK on TicWatch E3 from @gerison77 who reported the issue. 

https://discord.com/channels/629952586895851530/1488837966188253206/1536831262533099632

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -37,6 +37,14 @@
         android:label="@string/app_name"
         android:theme="@android:style/Theme.DeviceDefault">
 
+        <!-- androidx.wear.watchfacepush declares the wear-sdk platform library as required, which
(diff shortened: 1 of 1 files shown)


gerison77:
Tested and works

PR #5076, opened 2026-08-14
Automation: add actions to change SMB and UAM max minutes

### LINKED ISSUE #1821: Suggestion for smb improvement
I accompany dozens of loops in Israel and I noticed something that would greatly improve the accuracy of smb handling,
There is a problem when setting a maximum of minutes to smb at certain hours it should be much more aggressive and although it is by basal it is still inaccurate and the result (especially in children) is that if you set a relatively high number of minutes it brings some hours to hippo and if you lower the number of minutes it does not Enough facility at other times which is very important,
As a pirated solution I set in these hours a higher basal than needed to allow the smb to work harder also at certain hours I wanted there to be no smb at all I automated a temporary target value 101 for those hours and set the smb not to work with high target values ​​and so it works great
But there is a possibility to greatly improve everything in a much simpler way, just put in the profile settings an additional value of smb limit that can be changed by the hours of the day and then it is possible even at certain hours to set 0 and it will disable smb in those hours
In my experience this will be a huge improvement to the ability to use smb especially in children
Thanks for all the amazing work

### LINKED ISSUE #2529: Having separate SMB "profiles"
Is it possible to have different SMB "profiles", or even just enable/disable SMB at different times of the day?
E.g. for a little child, we'd prefer to disable the SMB option at night only, or change its parameters so that it is less aggressive than we need it to be during the day...
Also, it would be amazing to have the same with sound alerts - e.g. turn them off during the day or specific occasions (when she's embarrassed to be "beeping" in publc) but on at night (so we'd wake up if her BG starts dropping). Right now this has to be done manually - turned on and off - and runs the risk of forgetting to turn it on at night. 
A safety feature can leave just the super-urgent notifications on.
Thanks!!!!

olorinmaia:
Adds two automation actions:

- **Change SMB max minutes** — changes "Max minutes of basal to limit SMB"
- **Change UAM SMB max minutes** — changes "UAM max minutes of basal to limit SMB"

With a recurring time trigger and the existing enable/disable SMB action, users can lower or stop SMB during selected hours, for example at night.
The value is limited to the same range as the preference (15–120 min).

Both actions share a small base class and one editor. Unit tests included.

Tested on AAPS / AAPSClient: value changes apply on the next loop run and show in the OpenAPS SMB / AutoISF preferences.

Closes #1821
Closes #2529 (together with the existing enable/disable SMB action)

Screenshots:
| image | image |
|--------|--------|
| image | image |

### DIFF
--- a/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/actions/ActionSMBMaxMinutesChangeTest.kt
+++ b/plugins/automation/src/androidHostTest/kotlin/app/aaps/plugins/automation/actions/ActionSMBMaxMinutesChangeTest.kt
@@ -1,0 +1,60 @@
+package app.aaps.plugins.automation.actions
+
+import app.aaps.core.keys.IntKey
+import app.aaps.plugins.automation.AutomationStrings
+import com.google.common.truth.Truth.assertThat
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.skyscreamer.jsonassert.JSONAssert
+
+private const val STRING_JSON = """{"data":{"minutes":30},"type":"ActionSMBMaxMinutesChange"}"""
+
+class ActionSMBMaxMinutesChangeTest : ActionsTestBase() {
+
+    private lateinit var sut: ActionSMBMaxMinutesChange
+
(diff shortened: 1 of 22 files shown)


olorinmaia:
Some further input from me why this is needed.

This is a safe and useful addition to automation, especially when using dynamic ISF. 

Our kid's insulin need is very variable at night, in addition to erratic BG behaviour due to compression low or other influences, it would help alot to be able to lower SMB/UAM max minutes from e.g. 00.00 to 06.00. 

Lowering these settings will make the system take less risks based on the uncertainty I mentioned earlier happening at night. This PR will enable such options.

PR #5097, opened 2026-08-25
Add preference to show/render BG in the notification icon

premnirmal:
Adding a new preference to allow the user to view their BG in the notification icon.

The main motivation behind this change is so you can see your BG in the status bar, without having to pull down the notification shade:
image

This is done by adding a new `BooleanKey#NotificationShowBgOnIcon` enum value, with shared preferences key being `"notification_show_bg_on_icon"`.

The setting falls under the "Local Alerts" group.

When the preference is enabled, it replaces the persistent notification's small icon (`NotificationCompat.Builder#setSmallIcon`) with an icon created inside `BgIconRenderer`. The icon itself is a notification `IconCompat` created from a `Bitmap` of size 96x96. The `IconCompat` is also cached, folllowing what the existing `IconBitmapRenderer` does.

This bitmap is safe on memory because:
1. Size is tiny. Size: 96×96 ARGB_8888 = ~36 KB per bitmap
2. Lifetime is short, each overview update replaces the previous notification, so at most one or two are alive at once; GC handles them.
3. The icon itself is cached and only updates when the value changes.

**Screenshots:**
|Light|Dark|
|---|---|
|image| image|

|image|Screenshot_20260825_111754|Screenshot_20260825_111459|Screenshot_20260825_111446|

PR #5122, opened 2026-09-11
Use Android live updates for the persistent notification

premnirmal:
For devices running android 17 (sdk 37) and above, we can use the Android live udpates feature to show the persistent notification. More info in the android documentation about live updates: https://developer.android.com/develop/ui/compose/notifications/live-update.

This PR takes the existing persistent notification in the `PersistentNotificationPlugin` and applies live update using the `NotificationCompat.MetricStyle` to show the BG as a `NotificationCompat.Metric` with a `FixedFloat` or `FixedInt` value and the associated unit `"BG (mg/dL)"` or `"BG (mmol/L)"`.

The existing notificataion data is still shown in the notification.

The live update shows a chip in the status bar:
|mmol/L|mg/dL|
|---|---|
|image|image|
image

And the notification shows as a regular notification in the notification list, but as a 'live update' notification:
|mmol/L|mg/dL|
|---|---|
image|image|

And on the lockscreen it appears as a persistent live update, which updates with the persistent notification:
|mmol/L|mg/dL|
|---|---|
image|image|

Works on android auto:
image

This PR supersedes #5097

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -231,6 +231,7 @@ dependencies {
             "pumpcontrolImplementation"(project(it.path))
         }
 
+    implementation(libs.androidx.core)
     implementation(libs.androidx.lifecycle.process)
 
     testImplementation(project(":shared:tests"))
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -21,6 +21,8 @@
 
     <uses-permission-sdk-23 android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
 
+    <uses-permission android:name="android.permission.POST_PROMOTED_NOTIFICATIONS" />
+
     <!-- Remove legacy storage permissions pulled in by library manifest merge.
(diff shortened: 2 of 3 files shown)


olorinmaia:
Quick feedback. 🙂 The IOB and COB info in the notification should not be replaced by BG unit in my opinion. Some of this is also shared with android auto so make sure it doesn't break. IOB and COB is more valuable.

See also https://github.com/nightscout/AndroidAPS/pull/4787

premnirmal:
@olorinmaia Good point. I can restore those in the notification instead of displaying the hardcoded `"BG (unit)"` string

premnirmal:
@olorinmaia tested on Android auto and it works as expected :)
image

Also updated the line2 to include the COB and IOB.
image

Updated the PR descripton with this info as well.

Thanks for the early feedback! :)

premnirmal:
The build should be passing now, can we merge?

belevine:
Adding comments here, although have been chatting @premnirmal directly.  Confirmed this works on a Pixel (Fold) on A17 but it does not work on a Samsung (Galaxy Fold 8) on A17.

Issue #3385, opened 2024-07-17
Watchface not available


MTR93600:
Hi everyone, 

I recieves the galaxy watch ultra, which is on wear os 5, one UI 6.0.
Complication ok 
TILES with the shortcut to bolus or TT, ok
Watchface aaps not visible in the list in wearable app on the phone or on the watch when i look in the wtachface list.

Any idea how to fix it ?

koelewij:
Looking forward to possible solutions to test.

Philoul:
It seems that with Wear OS 5 usage of WFF is now mandatory.
So we will have to recreate new Watchfaces with this format and how reorganize code to reuse (and share with current existing watchfaces) the maximum of existing code 🤔.
I don'tknow the capability of customization by code (nor how to implement them) with WFF.
So we will probably have to include new libraries (hope we will not have conflicts with previous one), check and test how to build/update/show graph (for bg curves, prediction lines...) show/update imageviews and textviews...

MTR93600:
Apparently the file watch_face.xml need a specific code to make the custom watchface available. The manifest.xml required probably modification too. Then i'm not sure everything need to be rewrite. The change is related to security.

Philoul:
We are not alone with this problem...
Facer that have a catalog of more than 500000 watchfaces available within playstore, and millioms of customers has the same issue... See https://community.facer.io/t/important-samsung-galaxy-watch7-ultra-compatibility-update/93887

I understood that if you have a watch with Wear OS 4 with our watchfaces installed, if you manage a full backup of your watch, update your watch to wear OS 5.0 and restore your previous backup after the update, then it can still work (but not sure).

If you buy one of the latest galaxy watch like ultra with Wear OS 5, no way it's locked.

reijoahola-nightscout:
I had a backip for my Galaxy Watch 4 (Wear OS 4) in Google cloud and when installed the new Galaxy Watch Ultra, it installed backup from Google cloud. Such apps like G-Watch Wear, xDrip+ was successfully installed from backup and offered correctly complications that also worked in other watchfaces. But AAPS behaved differently. It was clearly copied from the backup and offered complications but they did not work. I could select also AAPS Tiules but I am not sure could they work correctly. I then uninstalled AAPS from the watch and installed it again by using Wear Installer 2. The situation was the same, complications could be selected but did not work. In all these these test, AAPS real watchfaces could not be selected and displayed.

Philoul:
[...]

PR #5123, opened 2026-09-12
Custom Watchface on Wear OS 6 watches, through Watch Face Format

Philoul:
## What this does

The Custom watch face is a code-based watch face. Wear OS 6 firmware no longer offers those in its
picker, so on a Galaxy Watch 7, 8 or 9 it cannot be selected at all and the wearer's own design
becomes unreachable. This brings it back on those watches.

Only Wear OS 6 and later is affected. On every other watch nothing changes: the code-based face
still runs, and this work reuses it rather than replacing it.

## How

Watch Face Format is declarative XML with no code, and the Custom watch face is drawing code driven
by a zip, so the design cannot be translated into it. The drawing is kept and only the way it
reaches the screen changes:

1. `CustomWatchface` draws itself into a **bitmap** instead of onto a watch face canvas.
2. That bitmap is published as a **full-screen image complication**.
3. A small Watch Face Format document, shipped with AAPS and installed by Watch Face Push, does
   nothing but show that complication full screen.

The wearer keeps their design, sends the same zip from the phone, and picks "AAPS" in the watch
face list like any other face.

Two slots carry the picture, one with seconds and one without, and the runtime swaps them when the
screen dims - that is the only thing that removes the second hand at the moment of dimming. Two
small complications carry the glucose and status readouts in always-on, drawn by the runtime
because our process is frozen while the watch dozes.

## Why this branch also carries #5047

#5047 is merged into this branch, so the two arrive together. That is deliberate. This port reuses
`CustomWatchface` as a renderer, under a strict rule that the code-based face must not change
behaviour, so the same build has to be checked both ways on a wrist: the code-based face on a watch
that still runs it, and the Watch Face Format face on a Wear OS 6 watch. Carrying both is what
makes that cross-check possible, and it is how the non-regression was actually verified rather than
assumed. Read #5047 first; this PR sits on top of it.

## Limits

- **The five complication slots of #5047 are not shown.** Those slots are drawn by the watch face
  framework straight onto the canvas, not by the views; this port draws the views into a bitmap, so
  their content never appears. A design that uses them still works, but those areas stay empty on a
  Wear OS 6 watch.
- **Probably more battery use than the code-based face.** Every frame is drawn, compressed and sent
  across the system instead of being drawn straight to the screen. This has not been measured on a
  watch, so the size of the difference is not known.
- **In always-on, the wearer's own design refreshes only when a glucose reading arrives**, about
  every five minutes: the system freezes our process while the watch dozes, and new data is what
  thaws it. The simplified always-on display does not have this limit - its clock and readouts are
  drawn by the runtime, so they stay correct.
- That choice is a setting, reached by long-pressing the face: always-on shows either the plain
  simplified display (default) or the wearer's design, dimmed.

## Details

`_docs/CWF_WFF_Prompt.md` holds the design, the measurements behind each choice, and what was tried
and rejected. Its Appendix A is written for wearers rather than developers.


olorinmaia:
First great job making CWF work again on newer watches, needs to be confirmed by GW7+ beyond users, but I think it's a formality. :)

Some feedback based on initial testing on my GW5.

1. Per discussion on Discord; I think the AAPS V4 face designed in WatchFaceStudios using pure complications should still be available to push to the watch. This PR fixes CWF but removes something new and unique (A well functioning and customizable face with pre-set AAPS complications enabling 7 different tap-to-actions which reaches most AAPS functions you would want to use, from the watchface directly). It makes sense to ship an official .wfs face since it's very easy to design faces now in WFS and we should add instructions in wiki on how this face can be updated. As discussed perhaps a setting on phone side in wear plugin. If users want to use the AAPS V4 WFF they can do so, if they want to use CWF they can do so by toggling the setting.

2. Always on display "IOB" is missing the unit "U". Displays "0.95". Expected "0.95U"

3. The second hand is sporadically sluggish / laggy on my GW5. Some times it stops, then jumps ahead. If watch screen is off and wakes up it takes a second or two before the second hand is visible. CWF I tested with is "Custom Watchface: Default Watchface".

4. Concern is battery life, but needs to be tested. The CWF frame producer renders whenever the watch is awake, whichever face is active. It should only render when CWF face is active

Philoul:
1. For the first version of CFW with WFF, I wanted to keep it simple for end user and as simple as I could for the code... That's why I re-used, as discussed together, this single available WFF slot.
As explained in my PR, I had to work on my PR #5047 working branch as starting point (too many risks of regression or inconsistency between both working branch, and none #5047, #5022 and #5062 was merged during most of my work...

2. in AOD, the informations included within the simple UI, including IOB,  is provided by a AAPS standard complications, not by my code.
I selected BG above hour, and TBR/COB/IOB for bottom complication. This issue should be fixed outside this PR...

3. Second hand management was the most tricky topic I had to optimized. Most CWF don't have seconds so no problem for them. But for those with second hand (not tested with "digital second" but same constraint) I had to propose compromise:
- the first 2 min after the watchface selection are not representative (a real nightmare visible with huge jump frozen steps because of too many things to load and build at the same time) so I accepted the issues concerning this period to not complicate too mush the code. The regularity of second hand starts to be good after (strategy in place to build cwf image by layers, with a cache of 5 forecast seconds calculated in advanced, and delivered immediately on each complication update...).
- after the 2 min, regularity of second hand is not perfect, but quite good most of the time. The tricky period is when new data is received from the phone (all the layers has to be rebuilt at the same time 🥵), so from time to time, you can get an second update a bit late then the following second a bit too quick, but most of the time no more.
- when watch CPU is overloaded, then a strategy of update of second hand once every 2s or once every 5 s is in place (rare but can also occurs from time to time). It's not perfect but I found it was better than erratic jumps of second hands during 5s or more...).

I could not optimize more the management of seconds, so my choice was to let discuss with beta testers before decide if we keep seconds like it is currently (really not prefect), or if we force the second visibility parameter to "gone" whatever the selected watchface...

olorinmaia:
I've uploaded two commits to fix issue 1. and 4. So far my testing turns out OK. See branch: https://github.com/olorinmaia/AndroidAPS/tree/wear/cwf_wff_add_setting_aapsv4_wfs_wff

1. c79130d373525d6e8fe92cb49237e0165289ff5b Wear: setting to choose which Watch Face Format face AAPS installs

4. fecde525cfeb6801f2b678831cd40a56154365f4 Wear: render the Custom watchface picture only while a face shows it

olorinmaia:
@MilosKozak @Philoul 
This PR is now in my opinion ready for review / merge to dev after extensive testing with my GW5 (Wear os 6) and emulator simulating older/newer wear os versions. A few other users have also reported that is it working well with the latest commits.

Merging this will finally put an end to CWF not being supported on newer watches, and we also keep the new "AAPSV4" watchface designed in Watchface Studios which is based on 7 complications as users can switch between CWF and "AAPSV4" in the phone app wear plugin. We also keep the legacy code-based CWF watchfaces for those who have watches which support this, as some wear os / watch setups may not see the new WFF watchfaces.

The following issue can be closed when merged to dev: https://github.com/nightscout/AndroidAPS/issues/3385

Philoul:
Note that this PR is huge, but it also has the modifications included within PR #5047...

I kept both to be able to merge in one or 2 steps 😉

